PR #13, opened 2018-04-24
Initial IPAM support

asarfaty:
Adding support for ip_pool and ip_block resources

### DIFF
--- a/nsxt/resource_nsxt_ip_block.go
+++ b/nsxt/resource_nsxt_ip_block.go
@@ -12,18 +12,6 @@ import (
 	"net/http"
 )
 
-// formatLogicalRouterRollbackError defines the verbose error when
-// rollback fails on a logical router create operation.
-const formatIpBlockRollbackError = `
-WARNING:
-There was an error during the creation of IP block router %s:
-%s
-Additionally, there was an error deleting the IP block during rollback:
-%s
-The IP block may still exist in the NSX. If it does, please manually delete it
-and try again.
-`
-
 func resourceNsxtIpBlock() *schema.Resource {
 	return &schema.Resource{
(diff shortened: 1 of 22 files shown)


nsxt/resource_nsxt_ip_block.go
```diff
@@ -0,0 +1,337 @@
...
+				Required:    true,
+				ForceNew:    true,
+			},
+			"subnet": &schema.Schema{
+				Type:        schema.TypeList,
+				Description: "List of subnets",
+				Optional:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": &schema.Schema{
```
annakhm: I think we need to understand where these ids are used.. if other resources could be dependent on these, maybe it would be preferable to expose these as separate objects?

nsxt/resource_nsxt_ip_block_subnet.go
```diff
@@ -0,0 +1,172 @@
...
+
+	if err != nil {
+		return fmt.Errorf("Error during IpBlockSubnet create: %v", err)
+	}
+
+	if resp.StatusCode != http.StatusCreated {
+		return fmt.Errorf("Unexpected status returned during IpBlockSubnet create: %v", resp.StatusCode)
+	}
+	d.SetId(ipBlockSubnet.Id)
+
```
annakhm: I wonder if d.Set("block_id", blockId) here might work around the nsx bug (won't work for read though, not sure how bad this is)

asarfaty: Since the read is always called after create/update I don't think it will help. But anyway we will not fully support IPAM in this NSX version, and it will be fixed for the next one.

PR #28, opened 2018-05-15
Add LB pool resource

asarfaty:


asarfaty:
1) I'll add the group membership later in a separate patch (I think it is a more advanced option than normally used)
2) I saw that not all snat types can be configured with the go vendor, so I used only the supported ones. After the go vendor merged, I'll add it too.
Do you what us to open issues for those? Or handle it internally?

annakhm:
I think we can remember these :) sdk changes are merged

PR #30, opened 2018-05-15
Remove doc & tests for relay service on tier0/1 ports

asarfaty:
The NSX does not support attaching a DHCP relay service on a logical
router downlink port or tier0 or tier1


asarfaty:
I've updated the docs here, but maybe we need to talk to the PM about this too.

annakhm:
I meant - specify this limitation in dhcp relay doc file

PR #75, opened 2018-06-21
Add centralized service router port resource

asarfaty:


website/docs/r/logical_router_downlink_port.html.markdown
```diff
@@ -38,7 +38,6 @@ The following arguments are supported:
 * `logical_router_id` - (Required) Identifier for logical Tier-1 router on which this port is created
 * `linked_logical_switch_port_id` - (Required) Identifier for port on logical switch to connect to
 * `ip_address` - (Required) Logical router port subnet (ip_address / prefix length)
-* `mac_address` - (Optional) Mac Address
```
annakhm: we can have it as attribute (instead of argument)

asarfaty: It is an attribute already. It just appeared twice in this file

PR #81, opened 2018-07-03
Fix http version validation in LB resources

annakhm:
HTTP_VERSION_2_0 is not supported in current platform, although
declared in spec. This option will be added here when backed by
platform.


asarfaty:
Maybe we should add the list of allowed values to the relevant doc files as well

asarfaty:
Maybe we should add the list of allowed values to the relevant doc files as well

PR #87, opened 2018-07-18
Adding NS Service Group resource

asarfaty:
The Service group contains NS services of any type as its members.
Please Note - nested NS services groups is currently not supported
by the NSX backend, although the api exists.

### DIFF
--- a/nsxt/resource_nsxt_ns_service_group.go
+++ b/nsxt/resource_nsxt_ns_service_group.go
@@ -1,0 +1,178 @@
+/* Copyright © 2018 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"github.com/hashicorp/terraform/helper/schema"
+	api "github.com/vmware/go-vmware-nsxt"
+	"github.com/vmware/go-vmware-nsxt/common"
+	"github.com/vmware/go-vmware-nsxt/manager"
+	"log"
+	"net/http"
+)
+
+func resourceNsxtNsServiceGroup() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNsxtNsServiceGroupCreate,
+		Read:   resourceNsxtNsServiceGroupRead,
+		Update: resourceNsxtNsServiceGroupUpdate,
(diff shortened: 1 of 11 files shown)


nsxt/resource_nsxt_ns_service_group.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func getResourceReferencesFromStringsSet(d *schema.ResourceData, schemaAttrName string) []common.ResourceReference {
+	targetIds := interface2StringList(d.Get(schemaAttrName).(*schema.Set).List())
+	var referenceList []common.ResourceReference
+	for _, targetID := range targetIds {
+		elem := common.ResourceReference{
+			IsValid:    true,
+			TargetId:   targetID,
+			TargetType: "NSService",
```
annakhm: when service group membership is supported, do we plan to run a get call for each member, in order to understand if its a service or a group? if not, we should solve it on user level.

website/docs/r/ns_service_group.html.markdown
```diff
@@ -0,0 +1,66 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `description` - (Optional) Description of this resource.
+* `display_name` - (Optional) The display name of this resource. Defaults to ID if not set.
+* `tag` - (Optional) A list of scope + tag pairs to associate with this NS service group.
+* `members` - (Optional) List of NSServices IDs that can be added as members to an NSServiceGroup. All members should be of the same type : Ethernet, or Non Ethernet.
```
annakhm: Required

annakhm: lets say "same L2 type"

nsxt/resource_nsxt_ns_service_group.go
asarfaty: I think it should be solved in the user level, but I do not want to over complicate it now, since we do not know if it will ever be supported.
If it will be - we can add another parameter here for groups

website/docs/r/ns_service_group.html.markdown
asarfaty: Done

asarfaty: Done

PR #106, opened 2018-09-12
Sync vendor sdk to fix icmp_code

annakhm:
Icmp code needs to be omitted when configuring type-only services.
This fix may cause rejected request in early platform versions.


vendor/github.com/vmware/go-vmware-nsxt/manager/icmp_type_ns_service.go
```diff
@@ -9,7 +9,7 @@ type IcmpTypeNsServiceEntry struct {
 	ResourceType string `json:"resource_type"`
 
 	// ICMP message code
-	IcmpCode int64 `json:"icmp_code"`
+	IcmpCode int64 `json:"icmp_codei,omitempty"`
```
asarfaty: please remove the last 'i' in icmp_codei

PR #114, opened 2018-10-02
Allow unspecified value for max size in LB pool

annakhm:
Introduce another attribute to specify whether to limit number of
addresses in pool.

### DIFF
--- a/nsxt/resource_nsxt_lb_pool.go
+++ b/nsxt/resource_nsxt_lb_pool.go
@@ -189,9 +189,15 @@ func getPoolMemberGroupSchema() *schema.Schema {
 					ValidateFunc: validation.StringInSlice(ipRevisionFilterTypeValues, false),
 					Default:      "IPV4",
 				},
+				"limit_ip_list_size": &schema.Schema{
+					Type:        schema.TypeBool,
+					Description: "Specifies whether to limit pool members. If false, dynamic pool can grow up to the load balancer max pool member capacity.",
+					Optional:    true,
+					Default:     false,
+				},
 				"max_ip_list_size": &schema.Schema{
 					Type:        schema.TypeInt,
-					Description: "Define the maximum number of grouping object IP address list. These IP addresses would be used as pool members",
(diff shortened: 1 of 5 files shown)


nsxt/resource_nsxt_lb_pool.go
```diff
@@ -189,9 +189,15 @@ func getPoolMemberGroupSchema() *schema.Schema {
 					ValidateFunc: validation.StringInSlice(ipRevisionFilterTypeValues, false),
 					Default:      "IPV4",
 				},
+				"limit_ip_list_size": &schema.Schema{
+					Type:        schema.TypeBool,
+					Description: "Specifies whether to limit pool members. If false, dynamic pool can grow up to the load balancer max pool member capacity.",
+					Optional:    true,
+					Default:     false,
```
asarfaty: should  the default be false? This means that if the user forgets to set it (or runs older config), the size will be ingnored.

nsxt/resource_nsxt_lb_pool_test.go
```diff
@@ -252,6 +253,7 @@ func TestAccResourceNsxtLbPool_withMemberGroup(t *testing.T) {
 					resource.TestCheckResourceAttr(testResourceName, "algorithm", algorithm),
 					resource.TestCheckResourceAttr(testResourceName, "member.#", "0"),
 					resource.TestCheckResourceAttr(testResourceName, "member_group.#", "1"),
+					resource.TestCheckResourceAttr(testResourceName, "member_group.0.limit_ip_list_size", "true"),
```
asarfaty: Maybe also add a test for False value?

nsxt/resource_nsxt_lb_pool.go
asarfaty: Thinking about it again - I guess it should be false.

nsxt/resource_nsxt_lb_pool_test.go
annakhm: Done

PR #120, opened 2018-11-24
Allow proxy connection in go-vmware-nsxt

annakhm:
In addition, fix connection leak and increase concurrency.

This solves "too many open files" issue during testacc


annakhm:
Replacing with https://github.com/terraform-providers/terraform-provider-nsxt/pull/122

PR #157, opened 2019-03-18
Support for LogicalRouter FirewallSection and LogicalRouterPort Rule

fdhex:
EdgeFirewall sections cannot currently be edited due to lack of support for LogicalRouter applied_to sections. In a similar way, it is not possible to create a rule within a edge firewall section that applies to only a LogicalRouterPort.

Both are within /firewall/sections and can be updated using same API as DFW rules so there should be no reason the provider can't be used for them.

### DIFF
--- a/nsxt/resource_nsxt_firewall_section.go
+++ b/nsxt/resource_nsxt_firewall_section.go
@@ -59,7 +59,7 @@ func resourceNsxtFirewallSection() *schema.Resource {
 				Required:    true,
 				ForceNew:    true,
 			},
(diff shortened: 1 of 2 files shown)


fdhex:
@asarfaty I have updated the list of options in the firewall_section documentation

fdhex:
@asarfaty is there anything else you would need?

annakhm:
Hi @fdhex, thanks! We'll need to run a few tests for this change. We can acceptance tests as follow up.

fdhex:
@annakhm Have you had the chance to run the required tests?

One problem is the impossibility of deleting the default rule of an EFW but the API can be used to CRUD anything else on the EFW.

annakhm:
Hi @fdhex, unfortunately we didn't cover this test yet. We might need to add additional validations for EFW, so I'd rather not rush this one. Appreciate your patience on this.

annakhm:
Acceptance test for this will follow shortly.

PR #161, opened 2019-05-07
Change router adv default according to nsx version

annakhm:
The default was changed to True in NSXT 2.5
In addition, improve nsx version comparison.


annakhm:
Waiting for confirmation from platform

PR #175, opened 2019-07-04
Fix resource type for dhcp relay attachment

annakhm:
In NSX 2.5, resource type for dhcp relay attachment on router
downlink port has changed.
This patch extends validation to allow new type, and fixes the test
to assign and validate resource type based on backend version.


annakhm:
refiled with https://github.com/terraform-providers/terraform-provider-nsxt/pull/176

PR #190, opened 2019-09-16
Add resource nsxt_ip_pool_allocation_ip_address

MartinWeindel:
For allocating and releasing IP addresses from an IP pool, the resource `nsxt_ip_pool_allocation_ip_address` and the data source `nsxt_ip_pool` have been added.

Acceptance tests and documentation have also been added. For the data source and the allocation tests, an existing IP pool is needed. A new environmental variable `NSXT_TEST_IP_POOL` has been defined to specify it.

PR #196, opened 2019-10-18
fix for "Error during IPPoolAllocationIPAddress delete: EOF"

MartinWeindel:
A bug was introduced with the reordering of the error handling during the review of PR #190


annakhm:
Thanks!

PR #214, opened 2020-01-27
Fix edge cluster assignment in gateways

annakhm:
Unset edge cluster is not returned in platform response.


annakhm:
ok, thanks!

PR #220, opened 2020-02-03
Improve policy data source error handling

annakhm:


nsxt/resource_nsxt_policy_lb_pool.go
```diff
@@ -96,7 +96,7 @@ func getPolicyPoolMemberGroupSchema() *schema.Schema {
 					Type:        schema.TypeBool,
 					Description: "Use IPv4 addresses as server IPs",
 					Optional:    true,
-					Default:     true,
+					Default:     false,
```
asarfaty: Did you mean to do this change as well?

annakhm: Yes, I should have mentioned it in the commit (to align with platform default). Thanks for pointing this out, I noticed a copy-paste error in Description :)

PR #229, opened 2020-02-10
added proxy support

loperaja:
Included a line to support proxy from environment so it wont time out when beihnd a proxy


annakhm:
Thanks @loperaja

PR #250, opened 2020-02-25
Update vm's port's tags only if they are different

kytomaki:
Port tags should not be updated if they are identical.

### DIFF
--- a/nsxt/resource_nsxt_vm_tags.go
+++ b/nsxt/resource_nsxt_vm_tags.go
@@ -191,6 +191,10 @@ func updatePortTags(nsxClient *api.APIClient, id string, tags []common.Tag) erro
 	portsUpdated := 0
 
 	for _, port := range ports {
+		if equalTags(tags, port.Tags) {
+			log.Printf("[DEBUG] skipping update since tags are identical for port %s", port.Id)
+			continue
+		}
 		port.Tags = tags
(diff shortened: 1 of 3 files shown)


annakhm:
Hi @kytomaki, was your motivation to reduce unneeded roundtrips when the other attribute (vm tag) changes? In this case, please consider using d.HasChange terraform indicator instead.

kytomaki:
The problem manifested when running `terrafom destroy` on the resources with limited user. Problematic function is `resourceNsxtVMTagsDelete` which tries to delete vm's port's tags. The user in question didn't have privileges to alter the port.

https://github.com/kytomaki/terraform-provider-nsxt/blob/cd802f26993c5ad4d23acf8474a4bd29bddb7bb9/nsxt/resource_nsxt_vm_tags.go#L295-L297

So this is a bit of a workaround to get past that.

annakhm:
Thanks for the explanation @kytomaki, I think same workaround behavior would be achieved during deletion if we don't execute updateTags in case provider doesn't have any tags defined in schema. For update case, d.HasChange would do the job of avoiding unnecessary update. This approach seems simpler to me - would you agree? I can file a PR with this change if you prefer.

annakhm:
I'm suggesting to fix this with https://github.com/terraform-providers/terraform-provider-nsxt/pull/261

kytomaki:
Thank you so much the #261 seems to work perfectly! I'll close this :)

PR #262, opened 2020-03-11
Fix validation in policy vrf router

annakhm:
Following validation changes on backend, fix provider validaton and
corresponding tests for vrf attributes


annakhm:
VRF support is since 3.0.0 only

PR #274, opened 2020-03-25
Disable VMC temporarily

annakhm:
VMC environments are not yet officially supported with the provider.
After provider release this commit will be reverted.


annakhm:
Will refile

PR #282, opened 2020-03-29
Add nsxt_logical_switch data source

phanama:
This PR adds the nsxt_logical_switch data source. Our use case is to fetch logical switch ID by its name in terraform.

Requesting reviews and comments! @annakhm

### DIFF
--- a/nsxt/data_source_nsxt_logical_switch_test.go
+++ b/nsxt/data_source_nsxt_logical_switch_test.go
@@ -1,0 +1,113 @@
+/* Copyright © 2018 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"github.com/vmware/go-vmware-nsxt/manager"
+)
+
+func TestAccDataSourceNsxtLogicalSwitch_basic(t *testing.T) {
+	switchName := "terraform_test_switch"
+	testResourceName := "data.nsxt_logical_switch.test"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		CheckDestroy: func(state *terraform.State) error {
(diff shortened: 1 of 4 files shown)


annakhm:
Hi @phanama, thanks for this effort! Please take a look at test errors CI is showing.
`go test -timeout=30s -parallel=4 github.com/terraform-providers/terraform-provider-nsxt github.com/terraform-providers/terraform-provider-nsxt/nsxt 
# github.com/terraform-providers/terraform-provider-nsxt/nsxt [github.com/terraform-providers/terraform-provider-nsxt/nsxt.test]
nsxt/data_source_nsxt_logical_switch_test.go:60:20: undefined: getTransportZoneID
?   	github.com/terraform-providers/terraform-provider-nsxt	[no test files]
FAIL	github.com/terraform-providers/terraform-provider-nsxt/nsxt [build failed]
make: *** [test] Error 123
The command "make test" exited with 2.`

phanama:
@annakhm Looks like I skipped adding the utils_test.go. I've pushed it in the new commit, plus several missing changes on the data source's schema and testAcc case.

nsxt/data_source_nsxt_logical_switch_test.go
```diff
@@ -0,0 +1,117 @@
...
+				return fmt.Errorf("Unexpected status returned during logical switch deletion: %v", responseCode.StatusCode)
+			}
+			return nil
+		}
+	}
+	return fmt.Errorf("Error while deleting logical switch '%s': service not found", switchName)
+}
+
+func testAccNSXLogicalSwitchReadTemplate(name string) string {
+	return fmt.Sprintf(`
```
annakhm, line 109: May I suggest using logical switch resource to create the resource for this test? This way we won't need to introduce the new environment variable, and require less code.

phanama: That way, changes on the resource_nsxt_logical_switch might possibly affect the data source's test and in some cases possibly require some changes on the data source's code itself.
If that's alright or even preferable, we can do that. Wdyt?

annakhm: I think its OK. We prefer to have as little pre-configuring as possible before we run acc tests, so not having to pre-create the LS would help. Thanks!

phanama: Sorry for the really late reply!
When trying to do this, I can't find any example reference for using the resource in the data source test. Can you point me to any? All the other data sources seem to use the api directly to pre-create the related resource in the test.

PR #283, opened 2020-03-29
Ability to set tags on specific port/logical_switch

phanama:
This PR adds the ability to set vm port tags on specific ports by specifying the specific logical_switch_id/network the vm is attached to. Our use case is when integrating NSX Container Plug-In (NCP) with Kubernetes VMs.

NCP requires us to attach an additional network interface to the VMs to be used as the "Dataplane" network switch, and tag the ports attached there with special `ncp/node_name` and `ncp/cluster` tags, as explained [here](https://docs.vmware.com/en/VMware-NSX-T-Data-Center/2.5/ncp-kubernetes/GUID-4987811C-1107-4D54-AE8A-827D45F10506.html).

To work it around, we previously wrote a python script to tag those dataplane ports, and eventually created this feature to the nsxt_vm_tags.

Requesting reviews and comments @annakhm

### DIFF
--- a/nsxt/resource_nsxt_vm_tags.go
+++ b/nsxt/resource_nsxt_vm_tags.go
@@ -2,13 +2,14 @@ package nsxt
 
 import (
 	"fmt"
+	"log"
+	"net/http"
+	"strings"
+
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	api "github.com/vmware/go-vmware-nsxt"
 	"github.com/vmware/go-vmware-nsxt/common"
 	"github.com/vmware/go-vmware-nsxt/manager"
-	"log"
-	"net/http"
-	"strings"
 )
 
 // Note - this resource is time consuming to configure since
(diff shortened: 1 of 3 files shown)


nsxt/resource_nsxt_vm_tags_test.go
```diff
@@ -70,6 +73,68 @@ func TestAccResourceNsxtVMTags_import_basic(t *testing.T) {
...
 
+func TestAccResourceNsxtVMTags_withSpecificLogicalPort(t *testing.T) {
+	vmID := getTestVMID()
+	logicalSwitchID := getTestLogicalSwitchID()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccEnvDefined(t, "NSXT_TEST_VM_ID")
+			testAccEnvDefined(t, "NSXT_TEST_LOGICAL_SWITCH_ID")
```
annakhm, line 84: Can this please be renamed to NSXT_TEST_VM_LOGICAL_SWITCH_ID for clarity? Same below

nsxt/resource_nsxt_vm_tags.go
```diff
@@ -30,6 +31,12 @@ func resourceNsxtVMTags() *schema.Resource {
 				Description: "Instance id",
 				Required:    true,
 			},
+			"logical_switch_id": {
+				Type:        schema.TypeString,
+				Description: "(Optional) Logical switch id",
+				Optional:    true,
+				Computed:    true,
```
annakhm, line 38: Not sure why this is marked as computed?

phanama: Probably pasted. I've removed it in the new commit.

PR #298, opened 2020-04-27
Add the correct type of security policy rule tag

asarfaty:
The optional argument "Tag" under rule as part of the nsxt_policy_security_policy resource does not configure the Tag on the rules.
In Policy the Tag of a rule is a string not a scope:tag like on other objects as ultimatelty this is a field propagated to the Syslog.

### DIFF
--- a/nsxt/policy_common.go
+++ b/nsxt/policy_common.go
@@ -236,6 +236,11 @@ func getSecurityPolicyAndGatewayRulesSchema(scopeRequired bool) *schema.Schema {
 					Default:     false,
 				},
 				"tag": getTagsSchema(),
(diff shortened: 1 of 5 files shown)


nsxt/policy_common.go
```diff
@@ -235,7 +235,11 @@ func getSecurityPolicyAndGatewayRulesSchema(scopeRequired bool) *schema.Schema {
 					Optional:    true,
 					Default:     false,
 				},
-				"tag": getTagsSchema(),
```
annakhm, line 238: I think we need to preserve this tag, and add the other one with name "log_label" as it appears in the UI

PR #320, opened 2020-05-21
GM switching profile & GW profile & service data sources

asarfaty:


nsxt/data_source_nsxt_policy_qos_profile_test.go
```diff
@@ -72,20 +92,38 @@ func testAccDataSourceNsxtPolicyQosProfileDeleteByName(name string) error {
...
+		if err != nil {
+			return handleListError("QosProfile", err)
+		}
+		for _, objInList := range objList.Results {
+			if *objInList.DisplayName == name {
+				err := client.Delete(*objInList.Id)
+				if err != nil {
+					return handleDeleteError("QosProfile", *objInList.Id, err)
+				}
+				return fmt.Errorf("DEBUG ADIT this is global manager")
```
annakhm, line 109: letfover

nsxt/data_source_nsxt_policy_qos_profile_test.go
```diff
@@ -72,20 +92,38 @@ func testAccDataSourceNsxtPolicyQosProfileDeleteByName(name string) error {
...
-	if err != nil {
-		return handleListError("QosProfile", err)
-	}
-	for _, objInList := range objList.Results {
-		if *objInList.DisplayName == name {
-			err := client.Delete(*objInList.Id)
-			if err != nil {
-				return handleDeleteError("QosProfile", *objInList.Id, err)
+	if testAccIsGlobalManager() {
+		client := gm_infra.NewDefaultQosProfilesClient(connector)
```
annakhm, line 98: do you think we should replace this with search function as well?

nsxt/data_source_nsxt_policy_qos_profile_test.go
asarfaty: Done

nsxt/data_source_nsxt_policy_qos_profile_test.go
asarfaty: Not sure what you mean. There is not default QoS profile that we can use here, so forced to create it.

asarfaty: Done

PR #327, opened 2020-05-27
Federation transportzone support

enhaocui:


nsxt/data_source_nsxt_policy_transport_zone.go
```diff
@@ -41,11 +43,49 @@ func dataSourceNsxtPolicyTransportZone() *schema.Resource {
...
 
 func dataSourceNsxtPolicyTransportZoneRead(d *schema.ResourceData, m interface{}) error {
+	objSitePath := d.Get("site_path").(string)
+	transportType := d.Get("transport_type").(string)
+	if !isPolicyGlobalManager(m) && objSitePath != "" {
+		return globalManagerOnlyError()
+	}
+	if isPolicyGlobalManager(m) {
+		if objSitePath == "" {
+			return attributeRequiredGlobalManagerError()
```
annakhm, line 64: We should specify which attribute is required

nsxt/utils_test.go
```diff
@@ -102,6 +102,10 @@ func getTestCertificateName(isClient bool) string {
 	return os.Getenv("NSXT_TEST_CERTIFICATE_NAME")
 }
 
+func getTestSitePath() string {
+	return os.Getenv("NSXT_TEST_SITE_PATH")
```
annakhm, line 106: We already have NSXT_TEST_SITE_NAME variable, so I suggest to define nsxt_policy_site data source based on the var in the test, and use path from it as an input to TZ data source. This will also validate the suggested usage.

nsxt/data_source_nsxt_policy_transport_zone.go
enhaocui: Sure, fixing. Thanks!

nsxt/utils_test.go
enhaocui: Sure. Will fix. Thanks.

website/docs/d/policy_transport_zone.html.markdown
```diff
@@ -24,12 +24,20 @@ data "nsxt_policy_transport_zone" "vlan_transport_zone" {
 }
 ```
 
+```
+data "nsxt_policy_transport_zone" "overlay_transport_zone" {
```
annakhm, line 28: Lets indicate here that this is a global manager example

PR #332, opened 2020-05-31
Global manager realization data source

asarfaty:
More tests will be added once more GM resources are supported


nsxt/utils_test.go
```diff
@@ -116,9 +116,15 @@ func testAccIsGlobalManager() bool {
...
 }
 
-func testAccSkipIfIsGlobalManager(t *testing.T) {
+func testAccOnlyGlobalManager(t *testing.T) {
+	if !testAccIsGlobalManager() {
+		t.Skipf("This test requires a global manager environment")
+	}
+}
+
+func testAccOnlyLocalManager(t *testing.T) {
```
annakhm, line 125: I've duplicated this function, lets refactor to use a single one later

PR #334, opened 2020-06-01
GM support for Group

enhaocui:


nsxt/resource_nsxt_policy_group_test.go
```diff
@@ -60,6 +61,50 @@ func TestAccResourceNsxtPolicyGroup_singleIPAddressCriteria(t *testing.T) {
 	})
 }
 
+func TestAccResourceNsxtGlobalPolicyGroup_singleIPAddressCriteria(t *testing.T) {
+	name := fmt.Sprintf("test-nsx-global-policy-group-ipaddrs")
+	updatedName := fmt.Sprintf("%s-update", name)
+	testResourceName := "nsxt_policy_group.test"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccNsxtGlobalPolicyGroupPrecheck(t) },
```
asarfaty, line 70: you can add a list of precheck functions here, instead of defining a special one (there is already one for global manager and one for site)

nsxt/resource_nsxt_policy_group_test.go
```diff
@@ -352,15 +403,22 @@ func testAccNsxtPolicyGroupExists(resourceName string, domainName string) resour
...
 	for _, rs := range state.RootModule().Resources {
 
 		if rs.Type != "nsxt_policy_group" {
 			continue
 		}
 
+		var err error
 		resourceID := rs.Primary.Attributes["id"]
-		_, err := nsxClient.Get(domainName, resourceID)
+		if isPolicyGlobalManager(testAccProvider.Meta()) {
```
annakhm, line 415: I think resourceNsxtPolicyGroupExistsInDomain can be used here?

nsxt/resource_nsxt_policy_group_test.go
```diff
@@ -60,6 +61,50 @@ func TestAccResourceNsxtPolicyGroup_singleIPAddressCriteria(t *testing.T) {
 	})
 }
 
+func TestAccResourceNsxtGlobalPolicyGroup_singleIPAddressCriteria(t *testing.T) {
```
annakhm, line 64: Can we reuse all LM tests to run for GM also? It is important to test all scenarios since group is a complex resource.

enhaocui: Yes, current LM tests are already run for GM. This one is just to test Group resource works for specific domain.

nsxt/resource_nsxt_policy_group_test.go
enhaocui: Good catch. Let me fix this.

nsxt/resource_nsxt_policy_group_test.go
enhaocui: Fixed. Thanks!

PR #339, opened 2020-06-03
Add Global Manager support for Tier1

annakhm:


annakhm:
Tests are waiting for edge cluster data source

PR #343, opened 2020-06-04
GM Security Policy Support

enhaocui:


nsxt/resource_nsxt_policy_security_policy_test.go
```diff
@@ -240,7 +350,16 @@ func testAccNsxtPolicySecurityPolicyExists(resourceName string, domainName strin
 			return fmt.Errorf("Policy SecurityPolicy resource ID not set in resources")
 		}
 
-		_, err := nsxClient.Get(domainName, resourceID)
+		var err error
+		if isPolicyGlobalManager(testAccProvider.Meta()) {
```
annakhm, line 354: can we use resourceNsxtPolicySecurityPolicyExistsInDomain here?

enhaocui: Thanks for the catch. Yes we can, but that way we will lose track of the error message here. Instead the error will be logged in logAPIError in resourceNsxtPolicySecurityPolicyExistsInDomain. Do you think it's OK here?

annakhm: Yes, I think its OK since we don't expect errors other than 404 at this stage

enhaocui: Sure. Let me fix this here. Thanks!

nsxt/resource_nsxt_policy_security_policy_test.go
```diff
@@ -240,35 +350,64 @@ func testAccNsxtPolicySecurityPolicyExists(resourceName string, domainName strin
...
 	for _, rs := range state.RootModule().Resources {
 
 		if rs.Type != "nsxt_policy_security_policy" {
 			continue
 		}
 
 		resourceID := rs.Primary.Attributes["id"]
-		_, err := nsxClient.Get(domainName, resourceID)
+		if isPolicyGlobalManager {
+			_, err = nsxClient.(*gm_domains.DefaultSecurityPoliciesClient).Get(domainName, resourceID)
```
annakhm, line 379: I think we can use resourceNsxtPolicySecurityPolicyExistsInDomain here as well?

enhaocui: Thanks for the catch! fixed.

PR #355, opened 2020-06-11
Add GM-specific tests for gateways

annakhm:
Fix some issues with gateway resources.
In addition, allow fetching edge cluster by site only.

### DIFF
--- a/nsxt/resource_nsxt_policy_tier0_gateway_gm_test.go
+++ b/nsxt/resource_nsxt_policy_tier0_gateway_gm_test.go
@@ -116,7 +116,6 @@ func TestAccResourceNsxtPolicyTier0Gateway_globalManagerNoSubnet(t *testing.T) {
 	})
 }
 
-
 // TODO: Add test for ACTIVE_ACTIVE when HA VIP config is supported
 
 func testAccNsxtPolicyGMGatewayDeps() string {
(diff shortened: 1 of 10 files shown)


nsxt/resource_nsxt_policy_tier0_gateway_gm_test.go
```diff
@@ -0,0 +1,197 @@
...
+
+var testAccGmGatewayIntersiteSubnet = "10.10.2.0/24"
+
+// NOTE: This test assumes single edge cluster on both sites
+func TestAccResourceNsxtPolicyTier0Gateway_globalManager(t *testing.T) {
+	testResourceName := "nsxt_policy_tier0_gateway.test"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
```
enhaocui, line 21: Shall we also add "testAccOnlyGlobalManager(t)" here?

nsxt/resource_nsxt_policy_tier0_gateway_gm_test.go
```diff
@@ -0,0 +1,197 @@
...
+	})
+}
+
+// NOTE: This test assumes single edge cluster on both sites
+func TestAccResourceNsxtPolicyTier0Gateway_globalManagerNoSubnet(t *testing.T) {
+	testResourceName := "nsxt_policy_tier0_gateway.test"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
```
enhaocui, line 80: ditto

nsxt/resource_nsxt_policy_tier0_gateway_gm_test.go
annakhm: You're right, just counting on SITE definition is not enough. Thanks

PR #373, opened 2020-06-24
Support for AD in Group

enhaocui:


nsxt/resource_nsxt_policy_group_test.go
```diff
@@ -962,3 +1023,40 @@ resource "nsxt_policy_group" "test" {
...
+
+func testAccNsxtPolicyGroupIdentityGroup(name string) string {
+	return fmt.Sprintf(`
+resource "nsxt_policy_group" "test" {
+  display_name = "%s"
+  description  = "Acceptance Test"
+
+  identity_group {
+        distinguished_name = "test-dn"
+        domain_base_distinguished_name = "test-dbdn"
```
asarfaty, line 1035: What about the sid field?

nsxt/resource_nsxt_policy_group.go
```diff
@@ -72,6 +72,13 @@ func resourceNsxtPolicyGroup() *schema.Resource {
 				Elem:        getConjunctionSchema(),
 				Optional:    true,
 			},
+			"identity_group": {
```
annakhm, line 75: Should this be part of criteria? Can it be part of AND/OR clauses with other expressions?

nsxt/resource_nsxt_policy_group_test.go
enhaocui: Sure. Will add Sid field in the test.

nsxt/resource_nsxt_policy_group.go
enhaocui: Nope this is different from Criteria. Criteria is added in expression field in the actual api call. But identity_group is only supported in extended_expression field. In fact, identity_group is the only resource supported in extended_expression now, and is limit to one at most. From api doc:

"Extended Expression allows additional higher level context to be
specified for grouping criteria. (e.g. user AD group)
...
Current version only supports a single IdentityGroupExpression. In the
future, this might expand to support other conjunction and non-conjunction
expression."

On second thought I think it would be best if we introduce an extended_criteria field, and the schema of it should be similar to criteria, only that currently we support one single identity_group in it.

enhaocui: > Please also add the new structure to the group doc

Sure. Will update the doc when code is finalized. Thanks.

annakhm: Thanks for the explanation :) Is there an AND relationship with "old" expression?

enhaocui: > Thanks for the explanation :) Is there an AND relationship with "old" expression?

Yes Anna, according to api doc, "Extended expression are implicitly AND with expression". I will add this in Group doc.

website/docs/r/policy_group.html.markdown
```diff
@@ -101,6 +112,8 @@ The following arguments are supported:
     * `value` (Required for a `condition`) User specified string value to use in the query. For `Tag` criteria, use 'scope|value' notation if you wish to specify scope in criteria.
 * `conjunction` (Required for multiple `criteria`) When specifying multiple `criteria`, a conjunction is used to specify if the criteria should selected using `AND` or `OR`.
   * `operator` (Required for `conjunction`) The operator to use. Must be one of `AND` or `OR`. If `AND` is used, then the `criteria` block before/after must be of the same type and if using `condition` then also must use the same `member_type`.
+* `extended_criteria` (Optional) A condition block to specify higher level context to include in this Group's members. (e.g. user AD group). Currently only one block is supported by NSX. Note that `extended_criteria` is implicitly `AND` with `criteria`.
+  * `identity_group` (Optional) A repeatable condition block select user AD groups to be included in this Group. Note that `identity_groups` are `OR` with each other.
```
annakhm, line 116: Please add identity_group attributes below

nsxt/resource_nsxt_policy_group_test.go
```diff
@@ -381,6 +381,85 @@ func TestAccResourceNsxtPolicyGroup_multipleNestedCriteria(t *testing.T) {
 	})
 }
 
+func TestAccResourceNsxtPolicyGroup_identityGroup(t *testing.T) {
+	name := fmt.Sprintf("test-nsx-policy-group-identity-group")
+	testResourceName := "nsxt_policy_group.test"
+	updatedName := fmt.Sprintf("%s-update", name)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t); testAccOnlyLocalManager(t) },
```
annakhm, line 390: Is there a reason for LM only?

enhaocui: Yes, identity group is not supported on GM.

website/docs/r/policy_group.html.markdown
enhaocui: Sure, added. Thanks for catching this.

nsxt/resource_nsxt_policy_group.go
```diff
@@ -596,15 +686,21 @@ func resourceNsxtPolicyGroupCreate(d *schema.ResourceData, m interface{}) error
 		return err
 	}
 
+	extendedCriteriaSets := d.Get("extended_criteria").([]interface{})
```
annakhm, line 689: do you think we should add a condition for local manager only? Or maybe there should be a validation func that errors out if these are defined for GM.

website/docs/r/policy_group.html.markdown
```diff
@@ -101,6 +112,11 @@ The following arguments are supported:
     * `value` (Required for a `condition`) User specified string value to use in the query. For `Tag` criteria, use 'scope|value' notation if you wish to specify scope in criteria.
 * `conjunction` (Required for multiple `criteria`) When specifying multiple `criteria`, a conjunction is used to specify if the criteria should selected using `AND` or `OR`.
   * `operator` (Required for `conjunction`) The operator to use. Must be one of `AND` or `OR`. If `AND` is used, then the `criteria` block before/after must be of the same type and if using `condition` then also must use the same `member_type`.
+* `extended_criteria` (Optional) A condition block to specify higher level context to include in this Group's members. (e.g. user AD group). Currently only one block is supported by NSX. Note that `extended_criteria` is implicitly `AND` with `criteria`.
```
annakhm, line 115: Let's mention here that this is for LM only?

nsxt/resource_nsxt_policy_group.go
enhaocui: Sure. I think we can do both. Let me add a validation func here and add some explanation in doc.

website/docs/r/policy_group.html.markdown
enhaocui: > Let's mention here that this is for LM only?

Yes I'm on it.

enhaocui:
> Thanks!

Sure!

PR #382, opened 2020-07-02
Support GM Tier1 Interface

enhaocui:


nsxt/resource_nsxt_policy_tier1_gateway_interface.go
```diff
@@ -67,24 +76,52 @@ func getGatewayInterfaceSubnetList(d *schema.ResourceData) []model.InterfaceSubn
...
-	localeService, err := getPolicyTier1GatewayLocaleServiceEntry(tier1ID, connector)
-	if err != nil {
-		return err
-	}
-	if localeService == nil {
-		return fmt.Errorf("Edge cluster is mandatory on gateway %s in order to create interfaces", tier1ID)
+	localeServiceID := ""
+	if isPolicyGlobalManager(m) {
+		localeServices, err := listPolicyTier1GatewayLocaleServices(connector, tier1ID, true)
+		if err != nil {
```
annakhm, line 87: Since model.LocaleServices is same for T0 and T1, could we perhaps reuse the code below for both T0 and T1?

nsxt/resource_nsxt_policy_tier1_gateway_interface.go
```diff
@@ -42,6 +45,12 @@ func resourceNsxtPolicyTier1GatewayInterface() *schema.Resource {
 				Description: "Locale Service ID for this interface",
 				Computed:    true,
 			},
+			"site_path": {
```
annakhm, line 48: should this be ForceNew? Both in T1 and T0

website/docs/r/policy_tier1_gateway_interface.html.markdown
```diff
@@ -50,6 +50,7 @@ The following arguments are supported:
 * `mtu` - (Optional) Maximum Transmission Unit for this interface.
 * `ipv6_ndra_profile_path` - (Optional) IPv6 NDRA profile to be associated with this interface.
 * `urpf_mode` - (Optional) Unicast Reverse Path Forwarding mode, one of `NONE`, `STRICT`. Default is `STRICT`. This attribute is supported with NSX 3.0.0 onwards.
+* `site_path` - (Optional for global manager only) Path of the site the Tier0 edge cluster belongs to
```
annakhm, line 53: I think this can sound misleading (can be read as required for local manager, optional for global manager)

nsxt/resource_nsxt_policy_tier1_gateway_interface.go
enhaocui: > Since model.LocaleServices is same for T0 and T1, could we perhaps reuse the code below for both T0 and T1?
Sure, I will wrap the common parts. Thanks.

nsxt/resource_nsxt_policy_tier1_gateway_interface.go
enhaocui: Thanks for the catch. Will fix for both t1 and t0.

> should this be ForceNew? Both in T1 and T0

website/docs/r/policy_tier1_gateway_interface.html.markdown
enhaocui: Makes sense. I think it is better to mark it as "Optional" here, and explain in the description that this config only applies to Global Manager. BTW do you think we should make site_path a required param for global manager? Otherwise the Site of the interface would be non-deterministic.

> I think this can sound misleading (can be read as required for local manager, optional for global manager)

annakhm: I think if its required in UI, we should make it same.

PR #400, opened 2020-07-22
Context profile

enhaocui:
Policy ContextProfile Resource


website/docs/r/policy_context_profile.html.markdown
```diff
@@ -0,0 +1,80 @@
...
+This resource provides a method for the management of a Context Profile.
+ 
+## Example Usage
+
+```hcl
+resource "nsxt_policy_context_profile" "test" {
+    display_name      = "test"
+    description       = "Terraform provisioned ContextProfile"
+    attributes {
+        data_type = "STRING"
```
asarfaty, line 19: please correct the spacing in the example

website/docs/r/policy_context_profile.html.markdown
```diff
@@ -0,0 +1,80 @@
...
+
+This resource provides a method for the management of a Context Profile.
+ 
+## Example Usage
+
+```hcl
+resource "nsxt_policy_context_profile" "test" {
+    display_name      = "test"
+    description       = "Terraform provisioned ContextProfile"
+    attributes {
```
asarfaty, line 18: should be attribute and not attributes

website/docs/r/policy_context_profile.html.markdown
```diff
@@ -0,0 +1,80 @@
...
+        data_type = "STRING"
+        key = "URL_CATEGORY"
+        value = ["Abortion"]
+    }
+    attributes {
+        data_type = "STRING"
+        key = "APP_ID"
+        value = ["SSL"]
+        sub_attributes {
```
asarfaty, line 32: should be sub_attribute

nsxt/resource_nsxt_policy_context_profile.go
```diff
@@ -0,0 +1,486 @@
...
+		err = client.Delete(id, &force)
+	}
+	if err != nil {
+		return handleDeleteError("ContextProfile", id, err)
+	}
+
+	return nil
+}
+
+func getContextProfilePolicyAttributesSchema() *schema.Schema {
```
asarfaty, line 259: I think this should be at the top of the file. just before or after resourceNsxtPolicyContextProfile. This is the way it is done with most of the resources and it makes it easier to understand the structure

website/docs/r/policy_context_profile.html.markdown
```diff
@@ -0,0 +1,80 @@
...
+ 
+## Attributes Reference
+
+In addition to arguments listed above, the following attributes are exported:
+
+* `id` - ID of the resource.
+* `revision` - Indicates current revision number of the object as seen by NSX-T API server. This attribute can be useful for debugging.
+* `path` - The NSX path of the policy resource.
+* `attribute`:
+  * `description` - Description of the attribute value.
```
asarfaty, line 67: the attribute description is a regular field. not a computed one. it should be moved to the above section

nsxt/resource_nsxt_policy_context_profile.go
```diff
@@ -0,0 +1,486 @@
...
+	// Validate that attribute keys are unique
+	res := make(map[string][]string)
+	for _, attribute := range attributes {
+		attributeMap := attribute.(map[string]interface{})
+		key := attributeMap["key"].(string)
+		values := interface2StringList(attributeMap["value"].(*schema.Set).List())
+		subAttributes := attributeMap["sub_attribute"].(*schema.Set).List()
+		// There should be only one value if sub attributes are specified
+		if len(subAttributes) > 0 && len(values) > 1 {
+			err := fmt.Errorf("Multiple values found for attribute key %s. Sub-attribtes are only applicable to an attribute with a single value", key)
```
asarfaty, line 340: please fix the typo in Sub-attribtes

nsxt/resource_nsxt_policy_context_profile_test.go
```diff
@@ -0,0 +1,346 @@
...
+    tag   = "orange"
+  }
+%s
+}`, name, attributes)
+}
+
+func testAccNsxtPolicyContextProfileAttributeDomainNameTemplate() string {
+	return fmt.Sprintf(`
+attribute {
+  data_type = "STRING"
```
asarfaty, line 296: Please fix the spacing and indentations in the tests configurations here

website/docs/r/policy_context_profile.html.markdown
enhaocui: The description of attribute is in API doc, but I have tested it and it does not work for attribute - NSX ignores it in request body. And description is never set in NSX. That's why I put it as "computed". Do you think we should just not support this field?

website/docs/r/policy_context_profile.html.markdown
enhaocui: Sure, thanks for the catch. Will fix.

website/docs/r/policy_context_profile.html.markdown
enhaocui: Thanks for the catch. Will fix.

website/docs/r/policy_context_profile.html.markdown
enhaocui: Sure

nsxt/resource_nsxt_policy_context_profile.go
enhaocui: Sure. Thanks.

nsxt/resource_nsxt_policy_context_profile.go
enhaocui: Sure.

website/docs/r/policy_context_profile.html.markdown
asarfaty: I think you should open a bug to the NSX about this field, and exclude it from the testing as of now, but it shouldn't be a computed field.

nsxt/resource_nsxt_policy_context_profile_test.go
enhaocui: Sure. fixed

website/docs/r/policy_context_profile.html.markdown
enhaocui: OK, that sounds good to me. I will add it back as optional to the schema. Thanks!

nsxt/resource_nsxt_policy_context_profile.go
```diff
@@ -0,0 +1,479 @@
...
+					Required:     true,
+					ValidateFunc: validation.StringInSlice(attributeDataTypes, false),
+				},
+				"description": getDescriptionSchema(),
+				"is_alg_type": {
+					Type:        schema.TypeBool,
+					Description: "Whether the APP_ID value is ALG type or not",
+					Computed:    true,
+				},
+				"key": {
```
annakhm, line 80: Looks like only 1 attribute for each key is supported - I wonder if it would be more user friendly to specify three distinct attributes for each attr key.

nsxt/resource_nsxt_policy_context_profile.go
```diff
@@ -0,0 +1,479 @@
...
+			"revision":     getRevisionSchema(),
+			"tag":          getTagsSchema(),
+			"attribute":    getContextProfilePolicyAttributesSchema(),
+		},
+	}
+}
+
+func getContextProfilePolicyAttributesSchema() *schema.Schema {
+	return &schema.Schema{
+		Type:     schema.TypeSet,
```
annakhm, line 64: Are we sure order is not important here?

nsxt/resource_nsxt_policy_context_profile_test.go
```diff
@@ -0,0 +1,342 @@
...
+  key       = "DOMAIN_NAME"
+  value     = ["*-myfiles.sharepoint.com"]
+}`)
+}
+
+func testAccNsxtPolicyContextProfileAttributeAppIDTemplate() string {
+	return fmt.Sprintf(`
```
annakhm, line 299: I think this can be removed (same below)

website/docs/r/policy_context_profile.html.markdown
```diff
@@ -0,0 +1,77 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `display_name` - (Required) Display name of the resource.
+* `description` - (Optional) Description of the resource.
+* `tag` - (Optional) A list of scope + tag pairs to associate with this resource.
+* `nsx_id` - (Optional) The NSX ID of this resource. If set, this ID will be used to create the resource.
+* `attribute` - (Required) A repeatable block to specify attributes for the context profile. At least one block is required.
+  * `datatype` - (Required) Datatype for `attribute`, must be `STRING`.
```
annakhm, line 48: is there a reason to expose it then?

nsxt/resource_nsxt_policy_context_profile.go
enhaocui: I was thinking the same thing. But I think current implementation allows for more flexibility if this restriction changes on NSX in the future?

nsxt/resource_nsxt_policy_context_profile.go
enhaocui: Attribute order was not mentioned in the API doc. I did some experiment on UI and did not find the order of attributes  matter when creating context profiles.

nsxt/resource_nsxt_policy_context_profile_test.go
enhaocui: Thanks for the catch. Let me remove the redundant "fmt.Sprintf".

website/docs/r/policy_context_profile.html.markdown
enhaocui: I was thinking other datatypes may be allowed in the future, so I exposed this. On second thought this is indeed somewhat redundant as of now. Let me remove this config then. Thanks!

nsxt/resource_nsxt_policy_context_profile.go
annakhm: I think we should choose ease of use here. But if order of attributes is important, then we'll loose flexibility when choosing this approach.

nsxt/resource_nsxt_policy_context_profile.go
```diff
@@ -0,0 +1,506 @@
...
+}
+
+func getPolicyAttributeSubAttributeSchema() *schema.Schema {
+	return &schema.Schema{
+		Type:     schema.TypeSet,
+		Optional: true,
+		MaxItems: 1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"tls_cipher_suite": getPolicyAttributeSubAttributeValueSchema("tls_cipher_suite"),
```
annakhm, line 124: are these the only possible sub-attrs today?

nsxt/resource_nsxt_policy_context_profile.go
```diff
@@ -0,0 +1,506 @@
...
+	return nil
+}
+
+func validateSubAttributes(attributes []interface{}) error {
+	// Validates that sub-attribute keys only present in an attribute with one value
+	for _, attribute := range attributes {
+		attributeMap := attribute.(map[string]interface{})
+		values := attributeMap["value"].(*schema.Set).List()
+		if len(attributeMap["sub_attribute"].(*schema.Set).List()) > 0 && len(values) > 1 {
+			err := fmt.Errorf("Multiple values found for attribute. Sub-attribtes are only applicable to an attribute with a single value")
```
annakhm, line 381: nit: typo

nsxt/resource_nsxt_policy_context_profile.go
enhaocui: Yes as documented in API doc

nsxt/resource_nsxt_policy_context_profile.go
enhaocui: Thanks for catching this. will fix.

nsxt/resource_nsxt_policy_context_profile.go
```diff
@@ -0,0 +1,508 @@
...
+		Optional: true,
+		MaxItems: 1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"description": getDescriptionSchema(),
+				"value": {
+					Type:        schema.TypeSet,
+					Description: "Values for attribute key",
+					Required:    true,
+					MinItems:    1,
```
annakhm, line 106: nit: MinItems is not needed if Required is true

nsxt/resource_nsxt_policy_context_profile.go
```diff
@@ -0,0 +1,508 @@
...
+	}
+	var attrClient interface{}
+	connector := getPolicyConnector(m)
+	isPolicyGlobalManager := isPolicyGlobalManager(m)
+	if isPolicyGlobalManager {
+		attrClient = gm_cont_prof.NewDefaultAttributesClient(connector)
+	} else {
+		attrClient = cont_prof.NewDefaultAttributesClient(connector)
+	}
+	attributeValues, err := listAttributesWithKey(attributeKeyMap[key], attrClient, isPolicyGlobalManager)
```
annakhm, line 360: I wonder if it would make sense to cache this. But I don't expect large amounts of context profiles to be created, so probably an overkill

website/docs/r/policy_context_profile.html.markdown
```diff
@@ -0,0 +1,76 @@
...
+Note: At least one of `app_id`, `domain_name`, or `url_category` must present.
+
+* `display_name` - (Required) Display name of the resource.
+* `description` - (Optional) Description of the resource.
+* `tag` - (Optional) A list of scope + tag pairs to associate with this resource.
+* `nsx_id` - (Optional) The NSX ID of this resource. If set, this ID will be used to create the resource.
+* `app_id` - (Optional) A block to specify app id attributes for the context profile. Only one block is allowed.
+  * `description` - (Optional) Description of the attribute.
+  * `value` - (Required) A list of string indicating values for the `app_id`. Must be a subset of valid values for `app_id` on NSX.
+  * `sub_attribute` - (Optional) A block to specify sub attribute for the `app_id`. Only one block is allowed.
```
annakhm, line 45: Can we add a note here that certain types of sub attributes are applicable to certain types of apps?

nsxt/resource_nsxt_policy_context_profile_test.go
```diff
@@ -0,0 +1,317 @@
...
+					resource.TestCheckResourceAttr(testResourceName, "app_id.832058275.value.#", "1"),
+					resource.TestCheckResourceAttr(testResourceName, "app_id.832058275.value.1810355442", "CIFS"),
+					resource.TestCheckResourceAttr(testResourceName, "app_id.832058275.sub_attribute.#", "1"),
+					resource.TestCheckResourceAttr(testResourceName, "app_id.832058275.sub_attribute.2393957168.cifs_smb_version.#", "2"),
+					resource.TestCheckResourceAttr(testResourceName, "app_id.832058275.sub_attribute.2393957168.tls_cipher_suite.#", "0"),
+					resource.TestCheckResourceAttr(testResourceName, "app_id.832058275.sub_attribute.2393957168.tls_version.#", "0"),
+					resource.TestCheckResourceAttr(testResourceName, "app_id.832058275.sub_attribute.2393957168.cifs_smb_version.3398512106", "CIFS_SMB_V1"),
+					resource.TestCheckResourceAttr(testResourceName, "app_id.832058275.sub_attribute.2393957168.cifs_smb_version.3787226665", "CIFS_SMB_V2"),
+				),
+			},
```
annakhm, line 210: can you please add another step here where sub attribute is removed, in order to make sure empty list is not ignores by patch?

enhaocui: Sure I will add these two actual steps.

website/docs/r/policy_context_profile.html.markdown
enhaocui: Sure. Thanks.

nsxt/resource_nsxt_policy_context_profile.go
enhaocui: Thanks for the note. Will remove the redundant MinItems here.

enhaocui:
> Thanks!

Sure NP. I will add test cases to include context profile in security policy in following PR.

PR #407, opened 2020-07-29
Smallfix docs/r/policy_segment.html.markdown

Thomas-Webber:
Fixed typo


annakhm:
Thank you @Thomas-Webber

PR #410, opened 2020-07-30
Fix NAT test for GM

annakhm:


asarfaty:
For some reason this test is still failing with GM on our CI

PR #418, opened 2020-08-09
Enable_pim is not supported with GM

asarfaty:
Enable_pim is not supported with GM

### DIFF
--- a/nsxt/resource_nsxt_policy_tier0_gateway_interface_test.go
+++ b/nsxt/resource_nsxt_policy_tier0_gateway_interface_test.go
@@ -193,16 +193,15 @@ func TestAccResourceNsxtPolicyTier0GatewayInterface_external(t *testing.T) {
 	ipAddress := "1.1.12.2"
 	updatedIPAddress := "1.2.12.2"
 	testResourceName := "nsxt_policy_tier0_gateway_interface.test"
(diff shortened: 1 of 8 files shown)


nsxt/resource_nsxt_policy_tier0_gateway_interface_test.go
```diff
@@ -193,6 +193,13 @@ func TestAccResourceNsxtPolicyTier0GatewayInterface_external(t *testing.T) {
 	ipAddress := "1.1.12.2"
 	updatedIPAddress := "1.2.12.2"
 	testResourceName := "nsxt_policy_tier0_gateway_interface.test"
+	var enablePim string
+	// enablePim is supported only with local manager
+	if testAccIsGlobalManager() {
+		enablePim = "false"
```
annakhm, line 199: did this work for you? GM seems to reject pim even if set to false, it needs to be absent from request

asarfaty: You are right. It worked on my local setup but not on the CI.

asarfaty: But still we need to update the code & doc as well as the test.

PR #420, opened 2020-08-12
Support Tier0 Gateway HA Vip config

asarfaty:
currently tested only with GM


asarfaty:
Verified that the tests works on local manager as well, but I'm not sure how to make sure it runs only if EC has 2 nodes.

nsxt/resource_nsxt_policy_tier0_gateway_ha_vip_config.go
```diff
@@ -0,0 +1,364 @@
...
+							Description: "Flag to enable this HA VIP config",
+							Optional:    true,
+							Default:     true,
+						},
+						"external_interface_paths": {
+							Type:        schema.TypeList,
+							Description: "paths to Tier0 external interfaces which are to be paired to provide redundancy",
+							Required:    true,
+							MinItems:    2,
+							MaxItems:    2,
```
annakhm, line 49: Is there a limit of 2?

nsxt/resource_nsxt_policy_tier0_gateway_ha_vip_config.go
```diff
@@ -0,0 +1,364 @@
...
+							Elem: &schema.Schema{
+								Type:         schema.TypeString,
+								ValidateFunc: validatePolicyPath(),
```
annakhm, line 52: can we validate same GW and  locale service here?

nsxt/resource_nsxt_policy_tier0_gateway_ha_vip_config.go
```diff
@@ -0,0 +1,364 @@
...
+							MaxItems:    2,
+							Elem: &schema.Schema{
+								Type:         schema.TypeString,
+								ValidateFunc: validateIPCidr(),
+							},
+						},
+					},
+				},
+			},
+			"locale_service_id": {
```
annakhm, line 69: I think this can be inferred from the interfaces?

annakhm, line 69: we expose locale service path, so I think path should be used here? This attribute is for sake of dependencies I think (otherwise it can be deferred from interface path), so it would make sense to use path.

nsxt/resource_nsxt_policy_tier0_gateway_ha_vip_config.go
```diff
@@ -0,0 +1,364 @@
...
+				Type:        schema.TypeString,
+				Description: "Id of associated Gateway Locale Service on NSX",
+				Computed:    true,
+			},
+			"tier0_id": {
```
annakhm, line 74: I don't think we need both locale service and gw path, since dependency need should be satisfied by locale service path.

nsxt/resource_nsxt_policy_tier0_gateway_ha_vip_config.go
annakhm: Actually I don't think we need this at all, since interfaces are already dependent on GW

nsxt/resource_nsxt_policy_tier0_gateway_ha_vip_config.go
asarfaty: Not in the api doc, but in the NSX gui it is limited to exactly 2.

nsxt/resource_nsxt_policy_tier0_gateway_ha_vip_config.go
asarfaty: I'm validating it in create & update. I on't think we can have a validation function with dependencies on few fields. right?

nsxt/resource_nsxt_policy_tier0_gateway_ha_vip_config.go
asarfaty: Its here just because it is useful in the code to have it for update/read/delete.
We can hide it from the documentation if you think it is not needed there.

nsxt/resource_nsxt_policy_tier0_gateway_ha_vip_config.go
annakhm: Yes sorry, we can't validate across fields.

nsxt/resource_nsxt_policy_tier0_gateway_ha_vip_config.go
annakhm: Sorry my bad, didn't notice its computed.

annakhm: However, I think in this case locale_service_id should be inside the config clause, as it should be possible to have separate configs for separate locale services on GM.

website/docs/r/policy_tier0_gateway_ha_vip_config.html.markdown
```diff
@@ -0,0 +1,73 @@
...
+  description            = "connection to segment2"
+  type                   = "SERVICE"
+  gateway_path           = data.nsxt_policy_tier0_gateway.gw1.path
+  segment_path           = nsxt_policy_vlan_segment.segment2.path
+  subnets                = ["12.12.2.2/24"]
+}
+
+resource "nsxt_policy_tier0_gateway_ha_vip_config" "ha-vip" {
+  config {
+    enabled                  = "true"
```
annakhm, line 39: nit: no quotes for boolean

nsxt/gateway_common.go
```diff
@@ -274,7 +274,7 @@ func initGatewayLocaleServices(d *schema.ResourceData, connector *client.RestCon
 
 	services := d.Get("locale_service").(*schema.Set).List()
```
annakhm, line 275: I think we should go into the update only if locale service changed (which is very rare occasion). This way we'll avoid the retry issue for most cases. I'm not sure HasChange("locale_service") would work for map (it didn't, but perhaps this was fixed with latest plugin sdk upgrade)
But GetChange definately works, and we can derive whether locale service is changed from there.
Perhaps this would be better as follow up patch.

nsxt/resource_nsxt_policy_tier0_gateway_ha_vip_config.go
asarfaty: The way it currently work is that all the entries in the same resource config have to match the same locale-service. If the use wants another vip-config for a different locale-service, it will be in a different resource.
I think it is simpler this way. I'll update the doc to reflect it.

nsxt/gateway_common.go
asarfaty: OK. I'll add a TODO, and do it later.

website/docs/r/policy_tier0_gateway_ha_vip_config.html.markdown
asarfaty: Done

nsxt/gateway_common.go
asarfaty: ok. I'll add a TODO in the code

nsxt/resource_nsxt_policy_tier0_gateway_ha_vip_config.go
annakhm: But is it possible to have more than one pair of HA VIP within same locale service?

annakhm: Yep I see its possible to create 2 pairs. I'm trying to think how to name the 'config' section so its less confusong.. maybe vip_pair? Anyway we can rename it later.

PR #422, opened 2020-08-12
Set Tier0 locale services as mandatory for GM

asarfaty:


website/docs/r/policy_tier0_gateway.html.markdown
```diff
@@ -101,7 +101,7 @@ The following arguments are supported:
 * `tag` - (Optional) A list of scope + tag pairs to associate with this Tier-0 gateway.
 * `nsx_id` - (Optional) The NSX ID of this resource. If set, this ID will be used to create the policy resource.
 * `edge_cluster_path` - (Optional) The path of the edge cluster where the Tier-0 is placed. Must be specified when `bgp_config` is enabled. This argument is not applicable for NSX Global Manager - use locale-services clause instead.
-* `locale_service` - (Optional) This argument is applicable for NSX Global Manager only. Multiple locale services can be specified for multiple locations.
+* `locale_service` - (Required) This argument is applicable for NSX Global Manager only. Multiple locale services can be specified for multiple locations.
```
enhaocui, line 104: Shall we use the wording similar to site_path in policy_edge_cluster.html?  Like: 
`locale_service ` - (Optional) This argument is applicable for NSX Global Manager only. Multiple locale services can be specified for multiple locations.

asarfaty: Ok. Thanks.

PR #423, opened 2020-08-13
Added nsxt_license Resource

ColemanJohnston:
Added a resource to allow terraform users to apply an NSX-T license.


website/docs/r/license.html.markdown
```diff
@@ -0,0 +1,25 @@
...
+resource "nsxt_license" "license" {
+  license_key = "0000-00000-00000-00000-00000"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `license_key` - (Required) NSX-T license key.
```
annakhm, line 25: Hi Coleman, looks good, just one nit - if import is supported, it should be mentioned here in the docs.
I also think that resources that are dependent on this license will fail to create/delete together with the licence, unless an explicit depends_on is specified in resource config. I don't think we can fix this easily, but worth mentioning this here in the doc.

nsxt/resource_nsxt_license.go
```diff
@@ -0,0 +1,101 @@
...
+	"log"
+	"net/http"
+	"regexp"
+)
+
+func resourceNsxtLicense() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNsxtLicenseCreate,
+		Read:   resourceNsxtLicenseRead,
+		Update: resourceNsxtLicenseCreate,
```
annakhm, line 20: I think we can remove the Update and thus forbid the update

nsxt/utils_test.go
```diff
@@ -112,6 +112,10 @@ func getTestCertificateName(isClient bool) string {
 	return os.Getenv("NSXT_TEST_CERTIFICATE_NAME")
 }
 
+func getTestLicenseKey() string {
+	return os.Getenv("NSXT_LICENSE_KEY")
```
annakhm, line 116: lets call it NSXT_TEST_LICENSE_KEY to align with other test env settings. Thanks!

CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 ## 2.2.0 (Unreleased)
+
+FEATURES:
+* **New Resource**: `nsxt_license`
```
annakhm, line 4: this is not accurate any more

nsxt/provider.go
```diff
@@ -541,6 +556,33 @@ func (processor remoteBasicAuthHeaderProcessor) Process(req *http.Request) error
...
+	}
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("Unexpected status returned during license create: %v", resp.StatusCode)
+	}
+
+	return nil
+}
+
+func configureLicenses(d *schema.ResourceData, clients *nsxtClients) error {
+	for _, lic_key := range d.Get("license_keys").([]interface{}) {
```
annakhm, line 577: I suggest we only do this is d.HasChange("license_keys").
Also shouldn't we handle license removal as well?
d.GetChange can help determine which ones were added and which ones removed

ColemanJohnston: It seems as if provider attributes are not persisted, and d.HasChange returns true even when there has not been a change in configuration. d.GetChange always produces an empty list and the current list instead of the previous configuration in this case.  It may make sense for License to be a resource if removal is necessary. Should we keep license in the provider and not handle removal, or make it a resource and document the need for explicit depends_on?

annakhm: I see.. Removing a license is probably a rare use case.. I'm ok with leaving this as is (it would be nice to leave a comment about this in the code and in docs).

annakhm (approved):

annakhm:
Hi Coleman, sorry to ask for yet another change on this.. 
Could you please add `ConflictsWith:` []string{"vmc_token"}` to the new provider attribute. This way we can protect the vmc use case from trying to configure the license, which will not work.

PR #433, opened 2020-08-23
Update GM locale services only if changed

asarfaty:
This will reduce the number of conflicts during update

### DIFF
--- a/nsxt/gateway_common.go
+++ b/nsxt/gateway_common.go
@@ -271,6 +271,51 @@ func initChildLocaleService(serviceStruct *model.LocaleServices, markForDelete b
 	return dataValue.(*data.StructValue), nil
 }
 
+func policyGatewayLocalServiceChanged(d *schema.ResourceData) bool {
+	// Return true if the locale service configuration changed on scheme
+	//First compare the number of locale services
+	oldS, newS := d.GetChange("locale_service")
+	if oldS.(*schema.Set).Len() != newS.(*schema.Set).Len() {
+		return true
+	}
+
+	// Compare each locale service
+	oldLocaleServices := oldS.(*schema.Set).List()
+	newLocaleServices := newS.(*schema.Set).List()
+	for index := range oldLocaleServices {
+		new := newLocaleServices[index].(map[string]interface{})
+		old := oldLocaleServices[index].(map[string]interface{})
+
+		// Compare the edge cluster path
+		if old["edge_cluster_path"] != new["edge_cluster_path"] {
+			return true
+		}
+
+		// compare the redistribution config
+		// If it has redistribution_config assume it changed because it is hard to compare
+		// TODO(asarfaty): improve this logic
+		oldRDC := old["redistribution_config"].([]interface{})
+		newRDC := new["redistribution_config"].([]interface{})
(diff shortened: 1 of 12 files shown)


nsxt/gateway_common.go
```diff
@@ -271,6 +271,50 @@ func initChildLocaleService(serviceStruct *model.LocaleServices, markForDelete b
...
+	}
+
+	// Compare each locale service
+	oldLocaleServices := oldS.(*schema.Set).List()
+	newLocaleServices := newS.(*schema.Set).List()
+	for index := range oldLocaleServices {
+		new := newLocaleServices[index].(map[string]interface{})
+		old := oldLocaleServices[index].(map[string]interface{})
+
+		// Compare the edge cluster path
```
annakhm, line 289: would reflect.DeepEqual work here for comparing maps?

nsxt/gateway_common.go
```diff
@@ -271,6 +271,50 @@ func initChildLocaleService(serviceStruct *model.LocaleServices, markForDelete b
 	return dataValue.(*data.StructValue), nil
 }
 
+func policyGatewayLocalServiceChanged(d *schema.ResourceData) bool {
```
annakhm, line 274: There seem to be a fix upstream for this https://github.com/hashicorp/terraform-plugin-sdk/pull/362. Not sure how to test this as it seems to be maintenance branch. But perhaps worth checking this out as this code is complex..

annakhm: According to this https://github.com/hashicorp/terraform-plugin-sdk/commit/83ea1db2e126b6b3bf2bd1f6a367367034359b2f#diff-4ac32a78649ca5bdd8e0ba38b7006a1eL35 this fix should be in plugin sdk 0.12.

annakhm: I've tested HasChange with maps in the provider and it works!

asarfaty: It didn't work for me on my setup, but maybe something is not up to date here. I'll update the code. Thanks.

nsxt/resource_nsxt_policy_tier0_gateway.go
```diff
@@ -787,12 +787,19 @@ func policyTier0GatewayResourceToInfraStruct(d *schema.ResourceData, connector *
 		}
 	} else {
 		// Global Manager
-		localeServices, err := initGatewayLocaleServices(d, connector, listPolicyTier0GatewayLocaleServices)
-		if err != nil {
-			return infraStruct, err
+		addLocaleServices := true
+		// If its an update scenario - update locale services only if they have changed
+		if !isCreate {
+			addLocaleServices = d.HasChange("locale_service")
```
annakhm, line 793: I think for create flow HasChange will also be true

website/docs/d/policy_context_profile.html.markdown
```diff
@@ -9,6 +9,8 @@ description: Policy Context Profile Profile data source.
 
 This data source provides information about Policy Context Profile configured in NSX.
 
+**NOTE:** This data source is experimental.
```
annakhm, line 12: This is already changed. I think it would be best to do a new PR to avoid rebasing and confusion..

asarfaty:
I'll create a new pull request for this

PR #434, opened 2020-08-24
Landing page for NSX-T Federation support

madhukark:


website/docs/guides_federation.html.markdown
```diff
@@ -0,0 +1,71 @@
+---
+layout: "nsxt"
+page_title: "Provider: NSXT"
+sidebar_current: "docs-nsxt-guides-vmc-index"
```
annakhm, line 4: This should be federation

website/docs/guides_federation.html.markdown
```diff
@@ -0,0 +1,71 @@
...
+description: |-
+  The VMware NSX-T Terraform Provider support for Federation 
+---
+
+# Terraform Provider for NSX-T extension for Federation
+
+The NSX Terraform Resources and Data Sources are now extended to support Federation. More information on Fedeartion that was introduced in NSX-T 3.0 can be found on the [NSX-T Product Page for Fedeartion](https://docs.vmware.com/en/VMware-NSX-T-Data-Center/3.0/administration/GUID-D5B6DC79-6733-44A7-8072-50221CF2122A.html)
```
annakhm, line 11: typo

website/docs/guides_federation.html.markdown
```diff
@@ -0,0 +1,71 @@
...
+  username       = "admin"
+  password       = "default"
+  global_manager = true
+}
+```
+
+Note that authentication with the Global Manager uses the same NSX-T Terraform provider but uses a `global_manager = true` flag for identifiation.
+
+## Using Resources and Data Sources
+Just like authentication, the resources and data soruces are common between the Local Manager (LM) and Global Manager (GM). However, there could be small differences in the attributes for each of the objects. These differences, when applicable, is called out in the documentation.
```
annakhm, line 30: are

PR #449, opened 2020-09-03
Expose predefined security policy resource

annakhm:
This is actually default security policy, named for consistency with
predefined gateway policy (which is not necessarily default).
This resource allows to modify default layer2 or default layer3
security policy.

### DIFF
--- a/nsxt/resource_nsxt_policy_predefined_security_policy.go
+++ b/nsxt/resource_nsxt_policy_predefined_security_policy.go
@@ -1,0 +1,439 @@
+/* Copyright © 2019 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
+	gm_domains "github.com/vmware/vsphere-automation-sdk-go/services/nsxt-gm/global_infra/domains"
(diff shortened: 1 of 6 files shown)


website/docs/r/policy_predefined_security_policy.html.markdown
```diff
@@ -0,0 +1,96 @@
...
+subcategory: "Policy - Firewall"
+layout: "nsxt"
+page_title: "NSXT: nsxt_policy_predefined_security_policy"
+description: A resource to update Predefined (Default) Security Security Policies.
+---
+
+# nsxt_policy_predefined_security_policy
+
+This resource provides a method to modify default Security Policy and its rules.
+This can be default layer2 policy or default layer2 policy. Maximum one resource
```
enhaocui, line 11: default layer2 policy or default layer3 policy

nsxt/resource_nsxt_policy_predefined_security_policy.go
```diff
@@ -0,0 +1,439 @@
...
+					return err
+				}
+				log.Printf("[DEBUG]: Deleting child rule with id %s", oldRuleID)
+				childRules = append(childRules, childRule)
+
+			}
+		}
+	}
+
+	if d.HasChange("default_rule") {
```
enhaocui, line 279: Maybe also using "else if" here since the two "if"s are mutually exclusive?

enhaocui (commented):
Some nits similar to fixed gw policy

website/docs/r/policy_predefined_security_policy.html.markdown
annakhm: Thanks for the catch! Done

nsxt/resource_nsxt_policy_predefined_security_policy.go
annakhm: In this resource you can define both rules and default rules.

enhaocui: Ah I see. Sorry for missing this.

enhaocui (approved):

PR #514, opened 2020-11-22
Support GM domain

asarfaty:


nsxt/resource_nsxt_policy_domain.go
```diff
@@ -0,0 +1,336 @@
...
+		if child != nil {
+			children = append(children, child)
+		}
+	}
+
+	if isUpdate {
+		// Delete old existing locations
+		connector := getPolicyConnector(m)
+		converter := bindings.NewTypeConverter()
+		dmClient := gm_domain.NewDefaultDomainDeploymentMapsClient(connector)
```
annakhm, line 117: Please consider, if not yet, getting old locations from the state without backend call, using d.GetChange("location")

annakhm (commented):
We should enable use of this resource in other resources (groups, sec policies), both in docs and in tests. This can be done in follow up.

nsxt/resource_nsxt_policy_domain.go
asarfaty: Thanks. I will try that

asarfaty: Actually I don't think it will help much, because I also need the ID of the one to be deleted, and I don't have it in the scheme. It could be an imported resources so the ID can be anyhting

nsxt/resource_nsxt_policy_domain.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+func resourceNsxtPolicyDomainCreate(d *schema.ResourceData, m interface{}) error {
+	// Initialize resource Id and verify this ID is not yet used
+	id, err := getOrGenerateID(d, m, resourceNsxtPolicyDomainExists)
+	if err != nil {
+		return err
+	}
+
+	if !isPolicyGlobalManager(m) {
```
annakhm, line 165: nit: this is already covered in getOrGenerateID

nsxt/resource_nsxt_policy_domain.go
annakhm: yes, good point

nsxt/resource_nsxt_policy_domain.go
```diff
@@ -0,0 +1,336 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"nsx_id":       getNsxIDSchema(),
+			"path":         getPathSchema(),
+			"display_name": getDisplayNameSchema(),
+			"description":  getDescriptionSchema(),
+			"revision":     getRevisionSchema(),
+			"tag":          getTagsSchema(),
+			"location": {
```
annakhm, line 38: I think historically we use plural when attribute is specified as a list, not as a clause, so suggest to use "locations" here. Maybe these should be "sites", since these correspond to nsxt_policy_site, is that correct?

website/docs/r/policy_domain.html.markdown
```diff
@@ -0,0 +1,53 @@
+---
+subcategory: "Policy - Grouping and Tagging"
+layout: "nsxt"
+page_title: "NSXT: nsxt_policy_domain"
+description: A resource to configure a Global manager domain.
```
annakhm, line 5: Lets specify here that's its also called region

nsxt/resource_nsxt_policy_domain.go
asarfaty: Removed

nsxt/resource_nsxt_policy_domain.go
asarfaty: Right. In the UI is locations, but everywhere else its sites. I'll change it.

annakhm (approved):

PR #516, opened 2020-11-23
Skip marked_for_delete data sources in search by name/type

asarfaty:


annakhm (commented):
Should we add the same check to listPolicyResourcesByID?

annakhm (approved):
Pls ignore the website linter error - its already fixed

asarfaty:
> Should we add the same check to listPolicyResourcesByID?

I'm not sure. I think if you ask for a specific ID you probably know what you want so its unlikely to be deleted. But we can if you think so.

annakhm:
We should be consistent with LM behavior. Do you know if GET for an object would return it in case its marked for deletion?

asarfaty:
No. Regular GET (or LIST) does not return those.

annakhm (approved):
Thx!

PR #529, opened 2020-12-09
Add dns forwarder resources

annakhm:
Add the following resources:
* nsxt_policy_dns_forwarder_zone
* nsxt_policy_gateway_dns_forwarder

In addition, enhance generator to avoid double policy keyword for
resources that contain Policy prefix in sdk model.


asarfaty (approved):

asarfaty:
nit: We should change the template header to set the year as 2020 (or 2021 soon)

annakhm:
Thanks Adit, will do

PR #543, opened 2020-12-21
Update changelog towards 3.1.1

annakhm:
In addition, add FAQ to documentation

### DIFF
--- a/website/docs/guides/faq.html.markdown
+++ b/website/docs/guides/faq.html.markdown
@@ -1,0 +1,64 @@
+---
+layout: "nsxt"
+page_title: "Frequently Asked Questions"
+description: |-
+  Frequently Asked Questions and Workarounds
+---
+
+<!-- TOC -->
+
+- [Dependency Errors on Update or Destroy](#dependency-errors-lifecycle)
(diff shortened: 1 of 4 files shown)


website/docs/guides/faq.html.markdown
```diff
@@ -0,0 +1,64 @@
...
+## Authorization Error on VMC
+
+Consider the following error:
+```
+User is not authorized to perform this operation on the application. Please contact the system administrator to get access. (code 401)
+```
+
+Assuming user permissions are sufficient, this error is usually due to missing `domain` configuration on the resource. For example, configuring group resource on VMC needs to contain domain configuration:
+
+```hcl
```
madhukark, line 57: Is the 'hcl' intentional?

madhukark (approved):
Just the comment about 'hcl' in the FAQ. Rest looks good.

website/docs/guides/faq.html.markdown
annakhm: Yes, but indeed I need to add it in above sample as well :)

asarfaty (approved):

PR #558, opened 2021-01-13
Support policy qos profile as a resource

asarfaty:


nsxt/resource_nsxt_policy_qos_profile.go
```diff
@@ -0,0 +1,352 @@
...
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
+	gm_infra "github.com/vmware/vsphere-automation-sdk-go/services/nsxt-gm/global_infra"
+	gm_model "github.com/vmware/vsphere-automation-sdk-go/services/nsxt-gm/model"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+)
+
+var rateLimiterResourceTypes = []string{"IngressRateLimiter", "IngressBroadcastRateLimiter", "EgressRateLimiter"}
```
annakhm, line 21: Maybe use model.QosBaseRateLimiter_RESOURCE_TYPE_INGRESSRATELIMITER etc. constants here?

nsxt/resource_nsxt_policy_qos_profile.go
```diff
@@ -0,0 +1,352 @@
...
+	schemaName := rateShaperSchemaNames[index]
+	resourceType := rateLimiterResourceTypes[index]
+	var shapers []map[string]interface{}
+	converter := bindings.NewTypeConverter()
+	converter.SetMode(bindings.REST)
+	for _, dataShaper := range shaperConf {
+		dataValue, _ := converter.ConvertToGolang(dataShaper, model.IngressRateLimiterBindingType())
+		shaper := dataValue.(model.IngressRateLimiter)
+		if shaper.ResourceType == resourceType {
+			average := *shaper.AverageBandwidth
```
annakhm, line 126: are these guaranteed not to be nil?

nsxt/resource_nsxt_policy_qos_profile.go
```diff
@@ -0,0 +1,352 @@
...
+	if isPolicyGlobalManager(m) {
+		client := gm_infra.NewDefaultQosProfilesClient(connector)
+		err = client.Delete(id, &boolFalse)
+	} else {
+		client := infra.NewDefaultQosProfilesClient(connector)
+		err = client.Delete(id, &boolFalse)
+	}
+
+	if err != nil {
+		return handleDeleteError("DhcpServer", id, err)
```
annakhm, line 348: QosProfile

nsxt/resource_nsxt_policy_qos_profile.go
asarfaty: Done

nsxt/resource_nsxt_policy_qos_profile.go
asarfaty: Done

nsxt/resource_nsxt_policy_qos_profile.go
asarfaty: Done

annakhm: f.e. in gateway_prefix_list all prefix fields are optional

asarfaty: In gateway_prefix_list there is no validation that optional fields are nil.

annakhm: It would be safer not to assume field being set in the struct, as we've seen NSX not specifying fields in response, which leads to these being set to nil in the SDK struct. 
Pointer assignment to schema map(like below) is safe, since terraform core handles nil use case:
elem["burst_size"] = shaper.BurstSize

But explicitly dereferencing pointer (like *shaper.BurstSize) should be nil-checked, unless we're 100% sure NSX will always sets these.

asarfaty: ok. Thanks.

annakhm (approved):

PR #563, opened 2021-01-20
Expose EVPN tunnel endpoint resource

annakhm:
In addition, improve test resource name randomization

### DIFF
--- a/nsxt/resource_nsxt_policy_evpn_tunnel_endpoint.go
+++ b/nsxt/resource_nsxt_policy_evpn_tunnel_endpoint.go
@@ -1,0 +1,220 @@
+/* Copyright © 2021 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"strings"
+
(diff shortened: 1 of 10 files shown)


nsxt/resource_nsxt_policy_evpn_tunnel_endpoint.go
```diff
@@ -0,0 +1,220 @@
...
+		DisplayName:    &displayName,
+		Description:    &description,
+		Tags:           tags,
+		EdgePath:       &edgePath,
+		LocalAddresses: localAddressList,
+	}
+
+	if mtu > 0 {
+		obj.Mtu = &mtu
+	}
```
enhaocui, line 78: Shall we raise an error instead of ignoring it?

annakhm: Hi Enhao, not sure what you mean? MTU is optional here

enhaocui: I see, sorry I did not see the validation.IntAtLeast(64) for mtu, so I thought it desirable to return error upon negative input.

enhaocui (approved):

Issue #584, opened 2021-03-05
Allow IP Address, CIDR, Range, or "ANY" as source_groups or destination_groups in nsxt_policy_security_policy resource


akgiesler:
The NSX-T API allows setting the source_groups and/or destination_groups of a Distributed firewall policy to an NSGroup, IP, IP Range, CIDR, or the "ANY" keyword.  This can also be accomplished in the policy UI and was tested in NSX-T v3.1.

The schema for the rules object in NSX-T 3.1 can be found [here](https://vdc-download.vmware.com/vmwb-repository/dcr-public/787988e9-6348-4b2a-8617-e6d672c690ee/a187360c-77d5-4c0c-92a8-8e07aa161a27/api_includes/types_Rule.html).

It appears that [the nsxt_policy_security_policy resource](https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/resource_nsxt_policy_security_policy.go) is enforcing the source_groups and destination_groups fields to be paths to NSgroups.  Link to the group validation is [here](https://github.com/vmware/terraform-provider-nsxt/blob/7860cf208555092647544be0ba7217da4ec00da6/nsxt/policy_utils.go#L160). Could this be adjusted to match the API?

annakhm:
This is now addressed, thanks to @akgiesler

AlexDias95:
@annakhm 
Version 3.1.1 of the provider allow the use of IP CIDR in the destination groups? Thx for the help!

PR #589, opened 2021-03-09
Allow IP CIDR Range or Any as Source and/or Destination Policy Groups

akgiesler:
Adds a new validator specific to source and destination policy groups.
NSX-T can accept an IP, Range, CIDR, or a Group Path as a source
and/or destination group through the security policy interface.

Updates the getSecurityPolicyAndGatewayRulesSchema function to use the
new validator.

Resolves: Issue #584

### DIFF
--- a/nsxt/resource_nsxt_policy_gateway_policy_test.go
+++ b/nsxt/resource_nsxt_policy_gateway_policy_test.go
@@ -384,6 +384,147 @@ func TestAccResourceNsxtGlobalPolicyGatewayPolicy_withDomain(t *testing.T) {
 	})
 }
 
+func TestAccResourceNsxtPolicyGatewayPolicy_withIPCidrRange(t *testing.T) {
+	name := getAccTestResourceName()
+	testResourceName := "nsxt_policy_gateway_policy.test"
+	defaultDirection := "IN_OUT"
+	defaultProtocol := "IPV4_IPV6"
+	policyIP := "10.10.20.5"
+	policyCidr := "10.10.20.0/22"
+	policyRange := "10.10.20.6-10.10.20.7"
+	updatedPolicyIP := "10.10.40.5"
+	updatedPolicyCidr := "10.10.40.0/22"
+	updatedPolicyRange := "10.10.40.6-10.10.40.7"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		CheckDestroy: func(state *terraform.State) error {
+			return testAccNsxtPolicyGatewayPolicyCheckDestroy(state, name, defaultDomain)
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: testAccNsxtPolicyGatewayPolicyWithIPCidrRange(name, policyIP, policyCidr, policyRange, policyIP, policyCidr, policyRange),
+				Check: resource.ComposeTestCheckFunc(
+					testAccNsxtPolicyGatewayPolicyExists(testResourceName, defaultDomain),
+					resource.TestCheckResourceAttr(testResourceName, "display_name", name),
+					resource.TestCheckResourceAttr(testResourceName, "description", "Acceptance Test"),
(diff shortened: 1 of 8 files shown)


annakhm:
Thank you @akgiesler! This looks great, but a documentation update is also needed, as well as acc tests. If you can add those it will be great, otherwise we'll follow up with those on top of your change.

akgiesler:
Sure thing @annakhm , we are working on those now.

akgiesler:
Hi @annakhm, 
I have updated the PR to include tests for Security Policy and Gateway Policy as they share the same function to get & validate their schema.  I have also updated the docs for both to reference the availability of IP, CIDR, or IP ranges.  

During our build-out of the tests, it was discovered that the provider already supports using "ANY" as a source or destination in security policy.  The way to do this was to map the source_groups or destination_groups to an empty set.

for example:
```
  rule {
    display_name       = "rule1"
    source_groups      = [ nsxt_policy_group.group1.path ]
    destination_groups = [ ]
    action             = "ALLOW"
    services           = [data.nsxt_policy_service.icmp.path]
  }
```
This was tested in NSX-T 3.1 using NSX-T Provider v3.1.0.  I added this feature into the Security policy documentation.  I _did not_ add this feature to the gateway policy documentation as I was unable to test it.

nsxt/resource_nsxt_policy_gateway_policy_test.go
```diff
@@ -384,6 +384,204 @@ func TestAccResourceNsxtGlobalPolicyGatewayPolicy_withDomain(t *testing.T) {
...
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		CheckDestroy: func(state *terraform.State) error {
+			return testAccNsxtPolicyGatewayPolicyCheckDestroy(state, name, defaultDomain)
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: testAccNsxtPolicyGatewayPolicyWithIPCidrRangeDestCreate(name, policyIP, policyCidr, policyRange),
```
annakhm, line 407: I would suggest to use a single function to test both Create and Update use cases (just pass different values in params), and also I think one test should be enough to cover both source and dest use cases. While our goal is to have good test coverage, we also try to keep CI running time reasonable :)

website/docs/r/policy_gateway_policy.html.markdown
```diff
@@ -93,7 +93,7 @@ The following arguments are supported:
 * `rule` (Optional) A repeatable block to specify rules for the Gateway Policy. Each rule includes the following fields:
   * `display_name` - (Required) Display name of the resource.
   * `description` - (Optional) Description of the resource.
-  * `destination_groups` - (Optional) A list of destination group paths to use for the policy.
+  * `destination_groups` - (Optional) Set of group paths, IPs, IP ranges, or CIDRs that serve as the destination for this rule.
```
annakhm, line 96: It should be helpful specifying here that IP/ranges/cidrs are only supported starting NSX 3.0.

annakhm:
Thanks for those changes @akgiesler! You are correct, "ANY" is equivalent to empty set. I've ran your tests on gateway policy and they pass fine. I noticed that docs for gateway policy are updated in your change.
I've left couple of small comments above, another nit would be to also update documentation in `resource_nsxt_policy_predefined_security_policy` and `resource_nsxt_policy_predefined_gateway_policy`.

nsxt/resource_nsxt_policy_gateway_policy_test.go
akgiesler: I went ahead and collapsed the tests into a single test for both security policy and gateway policy.  I also reduced the terraform generating functions down to single functions as you requested.

website/docs/r/policy_gateway_policy.html.markdown
akgiesler: Hi @annakhm, I have updated all of the documentation as requested.

annakhm:
Thank you @akgiesler, merging this!

PR #610, opened 2021-04-12
Set "site_path" and "gateway_path" in gateway redistribution config

enhaocui:


nsxt/resource_nsxt_policy_gateway_redistribution_config.go
```diff
@@ -181,6 +181,12 @@ func resourceNsxtPolicyGatewayRedistributionConfigRead(d *schema.ResourceData, m
 	d.Set("bgp_enabled", config.BgpEnabled)
 	d.Set("ospf_enabled", config.OspfEnabled)
 	d.Set("rule", getLocaleServiceRedistributionRuleConfig(config))
+	if isPolicyGlobalManager(m) {
+		d.Set("site_path", getSitePathFromEdgePath(*obj.EdgeClusterPath))
```
annakhm, line 185: Are we sure EdgeClusterPath is set? Import can be requested for all sorts of weird objects.

annakhm (commented):
We might want to try and set ImportStateVerify to true in TestAccResourceNsxtPolicyGatewayRedistributionConfig_importBasic test after this change. This check verifies that state and intent are equal post-import. I think after this change these should be equal.

annakhm (approved):

PR #620, opened 2021-05-17
Update terraform-plugin-sdk

zmb3:
The newer version of the SDK is necessary in order for the acceptance tests to run with terraform v0.15, which has slightly different CLI option handling.

https://www.terraform.io/upgrade-guides/0-15.html#other-minor-command-line-behavior-changes


annakhm:
Hi @zmb3, thanks again for this PR. I've fixed a couple of acceptance tests that broke following the change, and opened a new PR based on this one plus test fixes - https://github.com/vmware/terraform-provider-nsxt/pull/642

Issue #618, opened 2021-05-12
Data source for nsxt_policy_lb_service


pezhore:
I'm trying to create a new virtual server on an existing LB, but it looks like there isn't a `data_source_nsxt_policy_lb_service` resource available.

I can just copy/paste from the UI, but is there a programmatic way of retrieving this path without the data source?

Expected behavior:

```hcl
data "nsxt_policy_lb_service" "existing" {
   display_name  = "Existing LB"
}

resource "nsxt_policy_lb_virtual_server" "test" {
  display_name = "Terraform Test"
  description = "Terraform provisioned virtual server"
  application_profile_path = data.nsxt_policy_lb_app_profile.app.path
  ip_address = "10.0.0.20"
  service_path = data.nsxt_policy_lb_service.existing.path
  ports = ["80"]
}
```

annakhm:
Hi @pezhore, there is no such feature today - please use policy path as string. Marking as enhancement.

annakhm:
This data source [exists ](https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/data_source_nsxt_policy_lb_service.go) today, hence closing this.

PR #670, opened 2021-09-10
Add nsxt_policy_lb_service data source as per issue #618

martinrohrbach:
This is my attempt to work on #618 which is marked as "good first issue". 

I've created the new data source which is pretty straight forward for this resource and added the appropriate basic tests as well. 

I ran the acceptance tests in the lab and they look fine (see below). I also used the development provider to run some actual terraform plan/apply's using the new resource and everything looks good so far. 

```
martin@ubuntu20:~/devel/terraform-provider-nsxt$ make testacc TESTARGS="-run=TestAccDataSourceNsxtPolicyLBService"
==> Checking that code complies with gofmt requirements...
GO111MODULE=on TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccDataSourceNsxtPolicyLBService -timeout 360m
?   	github.com/vmware/terraform-provider-nsxt	[no test files]
=== RUN   TestAccDataSourceNsxtPolicyLBService
=== PAUSE TestAccDataSourceNsxtPolicyLBService
=== CONT  TestAccDataSourceNsxtPolicyLBService
--- PASS: TestAccDataSourceNsxtPolicyLBService (11.09s)
PASS
ok  	github.com/vmware/terraform-provider-nsxt/nsxt	11.137s
martin@ubuntu20:~/devel/terraform-provider-nsxt$ 
```

Signed-off-by: Martin Rohrbach <martin.rohrbach@dell.com>

### DIFF
--- a/nsxt/data_source_nsxt_policy_lb_service.go
+++ b/nsxt/data_source_nsxt_policy_lb_service.go
@@ -1,0 +1,87 @@
+/* Copyright © 2019 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+)
+
+func dataSourceNsxtPolicyLbService() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNsxtPolicyLbServiceRead,
+
+		Schema: map[string]*schema.Schema{
+			"id":           getDataSourceIDSchema(),
+			"display_name": getDataSourceDisplayNameSchema(),
+			"description":  getDataSourceDescriptionSchema(),
+			"path":         getPathSchema(),
+		},
+	}
+}
+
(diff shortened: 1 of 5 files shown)


annakhm:
Hi @martinrohrbach, thank you for contributing!
As we move forward with NSX releases, we want to get rid of complex data source implementation that uses the heavy List API. With old data sources, we still keep in place for the sake of NSX 2.5 and below that does not support the alternative search API yet, but all new data sources should be implemented using search (and marked in docs as supported since 3.0 only). Its a much shorter 4-line implementation :) Please look at example here https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/data_source_nsxt_policy_bfd_profile.go.
Thank you!

martinrohrbach:
Hi @annakhm, thanks for looking at my pull request and thanks for the feedback. 

I was not aware of the search API so thanks for the pointer. And I have to admit that I was only able to spin up an NSX 2.5 lab on my laptop due to resource restrictions, I couldn't get any newer version to work (the manager would throw ENONMEMs using the ExtraSmall install). So I wouldn't be able to test what you're suggesting and wouldn't feel comfortable submitting something untested.

Then again, I am aware of the End of Support for 2.5 so I guess I need to find a better lab option before I submit anything else. Feel free to close the pull request for now if there is no way you can accept the implementation using the list API.

martinrohrbach:
OK, disregard what I wrote. I scrapped my lab and started off fresh with an older vcenter version in order to save some resources and low and behold I can run NSX-T 3.1.3 after all.

The search API is a breeze indeed, I changed it as you suggested in literally two minutes and my test still runs fine. Documentation is amended as well. Let me know if I need to change anything else.

annakhm:
Thank you, looks good! I'll trigger tests on our test center.

annakhm (approved):

PR #733, opened 2022-04-06
Add External IDs Support for Group

enhaocui:
This change adds external ids support for Group.

### DIFF
--- a/nsxt/resource_nsxt_policy_group.go
+++ b/nsxt/resource_nsxt_policy_group.go
@@ -44,6 +44,13 @@ var conjunctionOperatorValues = []string{
 	model.ConjunctionOperator_CONJUNCTION_OPERATOR_AND,
 }
 
+var externalMemberTypeValues = []string{
+	model.ExternalIDExpression_MEMBER_TYPE_VIRTUALMACHINE,
+	model.ExternalIDExpression_MEMBER_TYPE_VIRTUALNETWORKINTERFACE,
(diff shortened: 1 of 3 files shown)


nsxt/resource_nsxt_policy_group.go
```diff
@@ -117,6 +124,27 @@ func getMACAddressExpressionSchema() *schema.Resource {
...
+				Type:        schema.TypeSet,
+				Required:    true,
+				Description: "List of external ids",
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+			"member_type": {
+				Type:         schema.TypeString,
+				Required:     true,
```
annakhm, line 140: I think we can make this optional with default being `VirtualMachine`

enhaocui: Sure. Let me change it. Thanks!

annakhm:
```TestAccResourceNsxtGlobalPolicyGroup_externalIDCriteria
    resource_nsxt_policy_group_test.go:139: Step 1/2 error: Check failed: Check 7/9 error: nsxt_policy_group.test: list or set attribute 'conjunction' must be checked by element count key (conjunction.#) or element value keys (e.g. conjunction.0). Set element value checks should use TestCheckTypeSet functions instead.```

This is a result of recent SDK upgrade..

annakhm (approved):

Issue #758, opened 2022-05-24
nsxt_policy_nat_rule, reflexsive rule and destination_networks


benohara:
### Describe the bug

When using the nsxt_policy_nat_rule to create a reflexsive nat rule it appears to set destination_networks to an empty string.

This cause the UI to longer respond, not able to open, edit, delete nat rules on any gateway that has a reflexsive nat rule with destination_networks set to ""

vmware support confirmed this as per SR 22326897505

"I have an update from the engineering team on this issue. They were able to reproduce the issue in the lab. 

Issue is related to the destination_network field in REFLEXIVE rule. destination_network field is not applicable in case of REFLEXIVE rule. If we pass destination_network field as empty string ("") it saves it in DB and while loading in UI it performs IP validation and it fails. Now this buggy behaviour has been found and fixed in versions Jetfire & impactor (3.2.1)

Since issue is happening because of destination_network = "", as workaround we can set this field as "null" in case of REFLEXIVE rules, or simply don't add this in the payload. For eg:

PATCH  [https://<manager>/policy/api/v1/infra/tier-0s/<gateway_name>/nat/USER/nat-rules/<rule_name](https://%3Cmanager%3E/policy/api/v1/infra/tier-0s/%3cgateway_name%3e/nat/USER/nat-rules/%3crule_name)>

{
"sequence_number": 100,
"action": "REFLEXIVE",
"source_network": "172.16.10.31",
"destination_network": "", <======== set to "null" or remove this field
"service": "",
"translated_network": "192.178.50.2",
"scope": [],
"enabled": true,
"logging": true,
"firewall_match": "BYPASS",
"resource_type": "PolicyNatRule",
"id": "NatRule_1",
"display_name": "NatRule_1",
"path": "/global-infra/tier-0s/Tier0Gateway1/nat/USER/nat-rules/NatRule_1",
"relative_path": "NatRule_1",
"parent_path": "/global-infra/tier-0s/Tier0Gateway1/nat/USER",
"unique_id": "07b96ffe-6015-47de-a369-5440c623f127",
"marked_for_delete": false,
"overridden": false,
"_create_time": 1649313290768,
"_create_user": "admin",
"_last_modified_time": 1649313290768,
"_last_modified_user": "admin",
"_system_owned": false,
"_protection": "NOT_PROTECTED",
"_revision": 0
}"

Attempting to pass in desintation_networks as an empty list or not passing always causes the rule to create with detination_networks set to ""

[...]

Issue #753, opened 2022-05-13
NAT rule with action 'no_snat' requiring optional parameter translated_networks


erikverbruggen:
### Describe the bug

When creating a NAT rule with the action 'no_snat', the parameter 'translated_networks' should be optional but when running a 'terraform plan' an error is given that it is required.

resource "nsxt_policy_nat_rule" "no_nsat192" {
  display_name         = "no_snat192"
  action               = "NO_SNAT"
  source_networks      = ["10.0.0.0/8"]
  destination_networks = ["192.168.0.0/16"]
  gateway_path         = nsxt_policy_tier0_gateway.tier0_gateway.path
  logging              = false
  firewall_match       = "MATCH_INTERNAL_ADDRESS"
  enabled              = true
  rule_priority        = "100"
}

│ Error: Missing required argument
│
│   on main.tf line 659, in resource "nsxt_policy_nat_rule" "no_nsat192":
│  659: resource "nsxt_policy_nat_rule" "no_nsat192" {
│
│ The argument "translated_networks" is required, but no definition was found.

It is now not possible to create a 'no_snat' rule where the translated networks must by 'Any' instead of a specific CIDR.

### Reproduction steps

```bash
1. Use the NSX-T terraform provider
2. Create a NAT rule using the nsxt_policy_nat_rule resource
3. Configure the NAT rule to use action "NO_SNAT"
4. Do not configure the parameter "translated_networks"
5. Run "terraform plan"
```

### Expected behavior

When defining a NAT rule with the action "NO_SNAT" it should be able to create a rule where the translated networks parameter is "Any" instead of a specific CIDR.

### Additional context

Tested with: 
- NSX-T 3.1.3.7 and 3.1.3.5
- NSX-T provider 3.2.6

PR #759, opened 2022-06-02
Fix some issues with NAT

2ez4szliu:
Fix some issues with NAT

(1) When creating REFLEXIVE Nat rules, the destination_network should be set to nil instead of empty string (issue #758)

(2) Allow translated_network field to be optional for no_snat/no_dnat (issue #753)

### DIFF
--- a/nsxt/resource_nsxt_policy_nat_rule.go
+++ b/nsxt/resource_nsxt_policy_nat_rule.go
@@ -112,7 +112,7 @@ func resourceNsxtPolicyNATRule() *schema.Resource {
 			"translated_networks": {
 				Type:        schema.TypeList,
 				Description: "The translated network(s) for the NAT Rule",
-				Required:    true,
+				Optional:    true,
 				Elem: &schema.Schema{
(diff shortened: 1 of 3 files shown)


nsxt/resource_nsxt_policy_nat_rule.go
```diff
@@ -345,8 +362,8 @@ func resourceNsxtPolicyNATRuleCreate(d *schema.ResourceData, m interface{}) erro
 		ruleStruct.TranslatedPorts = &ports
 	}
 
-	if len(sNets) > 0 {
-		ruleStruct.SourceNetwork = &sNets
+	if sNets != nil && len(*sNets) > 0 {
```
annakhm, line 365: I think since sNets would be nil unless specified, we can move it up into struct initialization, no need for this if (same below)

2ez4szliu: Done. Also added a test case to test create NO_SNAT with no translated network

annakhm (approved):

PR #762, opened 2022-06-08
Do not set AS number in BGP unless specified

annakhm:
Due to platform validations, AS number needs to remain nil
unless specified. In order to avoid non-empty diff when AS
is set by platform, we need to define this attribute as
computed.


2ez4szliu (approved):
Changes lgtm

PR #764, opened 2022-06-14
Support locale_service Gateway clause for LM

annakhm:
Today, locale service clause is only supported on Global Manager.
This change allows to configure it for Local Manager as well. This
configuration conflicts with edge_cluster_path and allows more
flexibility.
When importing Gateway, edge_cluster_path will be initialized rather
than locale_service, and additional apply is required to switch to
locale_service configuration.

### DIFF
--- a/nsxt/resource_nsxt_policy_tier1_gateway.go
+++ b/nsxt/resource_nsxt_policy_tier1_gateway.go
@@ -427,17 +427,15 @@ func policyTier1GatewayResourceToInfraStruct(d *schema.ResourceData, connector *
 		gwChildren = append(gwChildren, dataValue)
 	}
 
-	if isGlobalManager {
-		if d.HasChange("locale_service") {
-			// Update localse services only if configuration changed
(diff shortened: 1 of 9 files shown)


nsxt/resource_nsxt_policy_tier1_gateway.go
```diff
@@ -559,10 +557,15 @@ func resourceNsxtPolicyTier1GatewayRead(d *schema.ResourceData, m interface{}) e
 		return handleReadError(d, "Locale Service for T1", id, err)
 	}
 	var services []map[string]interface{}
+	_, shouldSetLS := d.GetOk("locale_service")
 	if len(localeServices) > 0 {
 
+		if isGlobalManager {
```
2ez4szliu, line 563: nit: can we move this out of the `if len(localeServices) > 0` block, like L933-938 of nsxt/resource_nsxt_policy_tier0_gateway.go

annakhm: Good point, done

2ez4szliu (approved):

PR #769, opened 2022-06-29
Support VPN profiles

annakhm:

### DIFF
--- a/nsxt/resource_nsxt_policy_ipsec_vpn_dpd_profile.go
+++ b/nsxt/resource_nsxt_policy_ipsec_vpn_dpd_profile.go
@@ -1,0 +1,203 @@
+/* Copyright © 2022 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+)
+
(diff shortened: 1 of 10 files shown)


nsxt/resource_nsxt_policy_ipsec_vpn_dpd_profile.go
```diff
@@ -0,0 +1,198 @@
...
+	_, err := client.Get(id)
+	if err == nil {
+		return true, nil
+	}
+
+	if isNotFoundError(err) {
+		return false, nil
+	}
+
+	return false, logAPIError("Error retrieving resource", err)
```
2ez4szliu, line 75: shall we specify resource type?

nsxt/resource_nsxt_policy_ipsec_vpn_dpd_profile.go
```diff
@@ -0,0 +1,198 @@
...
+func resourceNsxtPolicyIPSecVpnDpdProfileUpdate(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+
+	id := d.Id()
+	if id == "" {
+		return fmt.Errorf("Error obtaining IPSecVpnDpdProfile ID")
+	}
+
+	// Read the rest of the configured parameters
+	description := d.Get("description").(string)
```
2ez4szliu, line 156: Are these type casting needed? Seems calling Get on schema will return the correct data type
https://github.com/hashicorp/terraform/blob/v0.14.11/helper/schema/schema.go#L65

nsxt/resource_nsxt_policy_ipsec_vpn_dpd_profile.go
annakhm: terraform would inform what resource is causing the error, so I think we can remove resource type from other places as well..

nsxt/resource_nsxt_policy_ipsec_vpn_dpd_profile.go
annakhm: `Get` returns `interface{}`, so unless we type cast here compiler would complain..

ksamoray (approved):

2ez4szliu (approved):

PR #773, opened 2022-07-27
Add support for VPN resources

2ez4szliu:
Add support for several VPN resources.
The PR includes support for:
Data sources:
- nsxt_policy_ipsec_vpn_local_endpoint
- nsxt_policy_ipsec_vpn_service 
- nsxt_policy_l2vpn_service

Resources:
- nsxt_policy_ipsec_vpn_session
- nsxt_policy_ipsec_vpn_service 
- nsxt_policy_ipsec_vpn_local_endpoint
- nsxt_policy_l2vpn_session
- nsxt_policy_l2vpn_service

Co-authered-by: Nico Vibert, Anna Khmelnitsky

### DIFF
--- a/nsxt/data_source_nsxt_policy_ipsec_vpn_local_endpoint.go
+++ b/nsxt/data_source_nsxt_policy_ipsec_vpn_local_endpoint.go
@@ -1,0 +1,33 @@
+/* Copyright © 2022 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceNsxtPolicyIPSecVpnLocalEndpoint() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNsxtPolicyIPSecVpnLocalEndpointRead,
+
+		Schema: map[string]*schema.Schema{
+			"id":           getDataSourceIDSchema(),
+			"nsx_id":       getNsxIDSchema(),
+			"display_name": getDataSourceDisplayNameSchema(),
+			"description":  getDataSourceDescriptionSchema(),
+			"path":         getPathSchema(),
+		},
+	}
+}
+
+func dataSourceNsxtPolicyIPSecVpnLocalEndpointRead(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+
+	_, err := policyDataSourceResourceRead(d, connector, isPolicyGlobalManager(m), "IPSecVpnLocalEndpoint", nil)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/nsxt/data_source_nsxt_policy_ipsec_vpn_service.go
+++ b/nsxt/data_source_nsxt_policy_ipsec_vpn_service.go
@@ -1,0 +1,35 @@
+/* Copyright © 2022 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceNsxtPolicyIPSecVpnService() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNsxtPolicyIPSecVpnServiceRead,
+
+		Schema: map[string]*schema.Schema{
+			"id":           getDataSourceIDSchema(),
+			"nsx_id":       getNsxIDSchema(),
+			"path":         getPathSchema(),
+			"display_name": getDisplayNameSchema(),
+			"description":  getDescriptionSchema(),
+			"revision":     getRevisionSchema(),
+			"tag":          getTagsSchema(),
+		},
+	}
+}
+
+func dataSourceNsxtPolicyIPSecVpnServiceRead(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+
+	_, err := policyDataSourceResourceRead(d, connector, isPolicyGlobalManager(m), "IPSecVpnService", nil)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/nsxt/resource_nsxt_policy_ipsec_vpn_service.go
+++ b/nsxt/resource_nsxt_policy_ipsec_vpn_service.go
@@ -1,0 +1,325 @@
+/* Copyright © 2022 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
+	t0_locale_service "github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/tier_0s/locale_services"
+	t1_locale_service "github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/tier_1s/locale_services"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+)
+
+var IPSecVpnServiceIkeLogLevelTypes = []string{
+	model.IPSecVpnService_IKE_LOG_LEVEL_DEBUG,
+	model.IPSecVpnService_IKE_LOG_LEVEL_INFO,
+	model.IPSecVpnService_IKE_LOG_LEVEL_WARN,
+	model.IPSecVpnService_IKE_LOG_LEVEL_ERROR,
+	model.IPSecVpnService_IKE_LOG_LEVEL_EMERGENCY,
+}
+
+func resourceNsxtPolicyIPSecVpnService() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNsxtPolicyIPSecVpnServiceCreate,
+		Read:   resourceNsxtPolicyIPSecVpnServiceRead,
+		Update: resourceNsxtPolicyIPSecVpnServiceUpdate,
+		Delete: resourceNsxtPolicyIPSecVpnServiceDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"nsx_id":              getNsxIDSchema(),
+			"path":                getPathSchema(),
+			"display_name":        getDisplayNameSchema(),
+			"description":         getDescriptionSchema(),
+			"revision":            getRevisionSchema(),
+			"tag":                 getTagsSchema(),
+			"locale_service_path": getPolicyPathSchema(true, false, "Polciy path for the locale service."),
+			"enabled": {
+				Type:        schema.TypeBool,
+				Description: "Enable/Disable IPSec VPN service.",
+				Optional:    true,
+				Default:     true,
+			},
+			"ha_sync": {
+				Type:        schema.TypeBool,
+				Description: "Enable/Disable IPSec VPN service HA state sync.",
+				Optional:    true,
+				Default:     true,
+			},
+			"ike_log_level": {
+				Type:         schema.TypeString,
+				Description:  "Log level for internet key exchange (IKE).",
+				ValidateFunc: validation.StringInSlice(IPSecVpnServiceIkeLogLevelTypes, false),
+				Optional:     true,
+				Default:      model.IPSecVpnService_IKE_LOG_LEVEL_INFO,
+			},
+			"bypass_rule": getIPSecVPNRulesSchema(),
+		},
+	}
+}
+
+func getNsxtPolicyIPSecVpnServiceByID(connector *client.RestConnector, gwID string, isT0 bool, localeServiceID string, serviceID string, isGlobalManager bool) (model.IPSecVpnService, error) {
+	if isT0 {
+		client := t0_locale_service.NewIpsecVpnServicesClient(connector)
+		return client.Get(gwID, localeServiceID, serviceID)
+	}
+	client := t1_locale_service.NewIpsecVpnServicesClient(connector)
+	return client.Get(gwID, localeServiceID, serviceID)
+}
+
+func patchNsxtPolicyIPSecVpnService(connector *client.RestConnector, gwID string, localeServiceID string, ipSecVpnService model.IPSecVpnService, isT0 bool) error {
+	id := *ipSecVpnService.Id
+	if isT0 {
+		client := t0_locale_service.NewIpsecVpnServicesClient(connector)
+		return client.Patch(gwID, localeServiceID, id, ipSecVpnService)
+	}
+	client := t1_locale_service.NewIpsecVpnServicesClient(connector)
+	return client.Patch(gwID, localeServiceID, id, ipSecVpnService)
+}
+
+func updateNsxtPolicyIPSecVpnService(connector *client.RestConnector, gwID string, localeServiceID string, ipSecVpnService model.IPSecVpnService, isT0 bool) error {
+	id := *ipSecVpnService.Id
+	if isT0 {
+		client := t0_locale_service.NewIpsecVpnServicesClient(connector)
+		_, err := client.Update(gwID, localeServiceID, id, ipSecVpnService)
+		return err
+	}
+	client := t1_locale_service.NewIpsecVpnServicesClient(connector)
+	_, err := client.Update(gwID, localeServiceID, id, ipSecVpnService)
+	return err
+}
+
+func deleteNsxtPolicyIPSecVpnService(connector *client.RestConnector, gwID string, localeServiceID string, isT0 bool, id string) error {
+	if isT0 {
+		client := t0_locale_service.NewIpsecVpnServicesClient(connector)
+		return client.Delete(gwID, localeServiceID, id)
+	}
+	client := t1_locale_service.NewIpsecVpnServicesClient(connector)
+	return client.Delete(gwID, localeServiceID, id)
+}
+
+func setBypassRuleInSchema(d *schema.ResourceData, bypassRules []model.IPSecVpnRule) {
+	var ruleList []map[string]interface{}
+	for _, rule := range bypassRules {
+		elem := make(map[string]interface{})
+		var srcList []string
+		for _, src := range rule.Sources {
+			srcList = append(srcList, *src.Subnet)
+		}
+		var destList []string
+		for _, dest := range rule.Destinations {
+			destList = append(destList, *dest.Subnet)
+		}
+		elem["sources"] = srcList
+		elem["destinations"] = destList
+		elem["action"] = rule.Action
+		ruleList = append(ruleList, elem)
+	}
+	err := d.Set("bypass_rule", ruleList)
+	if err != nil {
+		log.Printf("[WARNING] Failed to set bypass_rule in schema: %v", err)
+	}
+}
+
+func getIPSecVPNBypassRulesFromSchema(d *schema.ResourceData) []model.IPSecVpnRule {
+	rules := d.Get("bypass_rule")
+	if rules != nil {
+		rules := rules.([]interface{})
+		var ruleList []model.IPSecVpnRule
+		for _, rule := range rules {
+			data := rule.(map[string]interface{})
+			action := data["action"].(string)
+			sourceRanges := interface2StringList(data["sources"].(*schema.Set).List())
+			destinationRanges := interface2StringList(data["destinations"].(*schema.Set).List())
+			/// Source Subnets
+			sourceIPSecVpnSubnetList := make([]model.IPSecVpnSubnet, 0)
+			if len(sourceRanges) > 0 {
+				for _, element := range sourceRanges {
+					subnet := element
+					ipSecVpnSubnet := model.IPSecVpnSubnet{
+						Subnet: &subnet,
+					}
+					sourceIPSecVpnSubnetList = append(sourceIPSecVpnSubnetList, ipSecVpnSubnet)
+				}
+			}
+			/// Destination Subnets
+			destinationIPSecVpnSubnetList := make([]model.IPSecVpnSubnet, 0)
+			if len(destinationRanges) > 0 {
+				for _, element := range destinationRanges {
+					subnet := element
+					ipSecVpnSubnet := model.IPSecVpnSubnet{
+						Subnet: &subnet,
+					}
+					destinationIPSecVpnSubnetList = append(destinationIPSecVpnSubnetList, ipSecVpnSubnet)
+				}
+			}
+			ruleID := data["nsx_id"].(string)
+			if ruleID == "" {
+				ruleID = newUUID()
+			}
+			elem := model.IPSecVpnRule{
+				Action:       &action,
+				Sources:      sourceIPSecVpnSubnetList,
+				Destinations: destinationIPSecVpnSubnetList,
+				UniqueId:     &ruleID,
+				Id:           &ruleID,
+			}
+			ruleList = append(ruleList, elem)
+		}
+		return ruleList
+	}
+	return nil
+}
+
+func resourceNsxtPolicyIPSecVpnServiceRead(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+
+	id := d.Id()
+	if id == "" {
+		return fmt.Errorf("Error obtaining IPSecVpnService ID")
+	}
+	localeServicePath := d.Get("locale_service_path").(string)
+	isT0, gwID, localeServiceID, err := parseLocaleServicePolicyPath(localeServicePath)
+	if err != nil {
+		return err
+	}
+	obj, err := getNsxtPolicyIPSecVpnServiceByID(connector, gwID, isT0, localeServiceID, id, isPolicyGlobalManager(m))
+	if err != nil {
+		return handleReadError(d, "IPSecVpnService", id, err)
+	}
+	d.Set("display_name", obj.DisplayName)
+	d.Set("description", obj.Description)
+	setPolicyTagsInSchema(d, obj.Tags)
+	d.Set("nsx_id", id)
+	d.Set("path", obj.Path)
+	d.Set("revision", obj.Revision)
+	d.Set("enabled", obj.Enabled)
+	d.Set("ha_sync", obj.HaSync)
+	setBypassRuleInSchema(d, obj.BypassRules)
+	if obj.IkeLogLevel != nil {
+		d.Set("ike_log_level", obj.IkeLogLevel)
+	}
+	return nil
+}
+
+func resourceNsxtPolicyIPSecVpnServiceCreate(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+	localeServicePath := d.Get("locale_service_path").(string)
+	isT0, gwID, localeServiceID, err := parseLocaleServicePolicyPath(localeServicePath)
+	if err != nil {
+		return err
+	}
+	isGlobalManager := isPolicyGlobalManager(m)
+	id := d.Get("nsx_id").(string)
+	if id == "" {
+		id = newUUID()
+	} else {
+		_, err := getNsxtPolicyIPSecVpnServiceByID(connector, gwID, isT0, localeServiceID, id, isGlobalManager)
+		if err == nil {
+			return fmt.Errorf("IPSecVpnService with nsx_id '%s' already exists", id)
+		} else if !isNotFoundError(err) {
+			return err
+		}
+	}
+
+	displayName := d.Get("display_name").(string)
+	description := d.Get("description").(string)
+	enabled := d.Get("enabled").(bool)
+	haSync := d.Get("ha_sync").(bool)
+	rules := getIPSecVPNBypassRulesFromSchema(d)
+	tags := getPolicyTagsFromSchema(d)
+
+	ipSecVpnService := model.IPSecVpnService{
+		Id:          &id,
+		DisplayName: &displayName,
+		Description: &description,
+		Tags:        tags,
+		Enabled:     &enabled,
+		HaSync:      &haSync,
+		BypassRules: rules,
+	}
+
+	ikeLogLevel := d.Get("ike_log_level").(string)
+	if ikeLogLevel != "" {
+		ipSecVpnService.IkeLogLevel = &ikeLogLevel
+	}
+
+	err = patchNsxtPolicyIPSecVpnService(connector, gwID, localeServiceID, ipSecVpnService, isT0)
+	if err != nil {
+		return handleCreateError("IPSecVpnService", id, err)
+	}
+	d.SetId(id)
+	d.Set("nsx_id", id)
+	return resourceNsxtPolicyIPSecVpnServiceRead(d, m)
+}
+
+func resourceNsxtPolicyIPSecVpnServiceUpdate(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+
+	id := d.Id()
+	if id == "" {
+		return fmt.Errorf("Error obtaining IPSec VPN Service ID")
+	}
+	localeServicePath := d.Get("locale_service_path").(string)
+	isT0, gwID, localeServiceID, err := parseLocaleServicePolicyPath(localeServicePath)
+	if err != nil {
+		return err
+	}
+
+	displayName := d.Get("display_name").(string)
+	description := d.Get("description").(string)
+	enabled := d.Get("enabled").(bool)
+	haSync := d.Get("ha_sync").(bool)
+	rules := getIPSecVPNBypassRulesFromSchema(d)
+	tags := getPolicyTagsFromSchema(d)
+	revision := int64(d.Get("revision").(int))
+	ipSecVpnService := model.IPSecVpnService{
+		Id:          &id,
+		DisplayName: &displayName,
+		Description: &description,
+		Tags:        tags,
+		Enabled:     &enabled,
+		HaSync:      &haSync,
+		BypassRules: rules,
+		Revision:    &revision,
+	}
+
+	ikeLogLevel := d.Get("ike_log_level").(string)
+	if ikeLogLevel != "" {
+		ipSecVpnService.IkeLogLevel = &ikeLogLevel
+	}
+
+	log.Printf("[INFO] Updating IPSecVpnService with ID %s", id)
+	err = updateNsxtPolicyIPSecVpnService(connector, gwID, localeServiceID, ipSecVpnService, isT0)
+	if err != nil {
+		return handleUpdateError("IPSecVpnService", id, err)
+	}
+	d.Set("nsx_id", id)
+	return resourceNsxtPolicyIPSecVpnServiceRead(d, m)
+}
+
+func resourceNsxtPolicyIPSecVpnServiceDelete(d *schema.ResourceData, m interface{}) error {
+	id := d.Id()
+	if id == "" {
+		return fmt.Errorf("Error obtaining IPSec VPN Service ID")
+	}
+
+	localeServicePath := d.Get("locale_service_path").(string)
+	isT0, gwID, localeServiceID, err := parseLocaleServicePolicyPath(localeServicePath)
+	if err != nil {
+		return err
+	}
+
+	err = deleteNsxtPolicyIPSecVpnService(getPolicyConnector(m), gwID, localeServiceID, isT0, id)
+	if err != nil {
+		return handleDeleteError("IPSecVpnService", id, err)
+	}
+	return nil
+}
--- a/nsxt/resource_nsxt_policy_ipsec_vpn_service_test.go
+++ b/nsxt/resource_nsxt_policy_ipsec_vpn_service_test.go
@@ -1,0 +1,193 @@
+/* Copyright © 2022 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+var accTestPolicyIPSecVpnServiceCreateAttributes = map[string]string{
+	"display_name":  getAccTestResourceName(),
+	"description":   "terraform created",
+	"enabled":       "true",
+	"ha_sync":       "true",
+	"ike_log_level": "INFO",
+	"sources":       "192.168.10.0/24",
+	"destinations":  "192.169.10.0/24",
+	"action":        "BYPASS",
+}
+
+var accTestPolicyIPSecVpnServiceUpdateAttributes = map[string]string{
+	"display_name":  getAccTestResourceName(),
+	"description":   "terraform updated",
+	"enabled":       "false",
+	"ha_sync":       "true",
+	"ike_log_level": "DEBUG",
+	"sources":       "192.170.10.0/24",
+	"destinations":  "192.171.10.0/24",
+	"action":        "BYPASS",
+}
+
+func TestAccResourceNsxtPolicyIPSecVpnService_basic(t *testing.T) {
+	testResourceName := "nsxt_policy_ipsec_vpn_service.test"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t); testAccOnlyLocalManager(t) },
+		Providers: testAccProviders,
+		CheckDestroy: func(state *terraform.State) error {
+			return testAccNsxtPolicyIPSecVpnServiceCheckDestroy(state, accTestPolicyIPSecVpnServiceUpdateAttributes["display_name"])
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: testAccNsxtPolicyIPSecVpnServiceTemplate(true),
+				Check: resource.ComposeTestCheckFunc(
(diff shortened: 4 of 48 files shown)


nsxt/data_source_nsxt_policy_ipsec_vpn_local_endpoint.go
```diff
@@ -0,0 +1,117 @@
...
+				Optional:    true,
+				Default:     "vmc",
+			},
+			"locale_service": {
+				Type:        schema.TypeString,
+				Description: "Local_service",
+				Optional:    true,
+				Default:     "default",
+			},
+			"service_id": {
```
annakhm, line 37: Is this path of endpoint as indicated in description? In this case `endpoint_path` seems like a better fit

nsxt/data_source_nsxt_policy_ipsec_vpn_local_endpoint.go
```diff
@@ -0,0 +1,117 @@
...
+			"display_name": getDataSourceDisplayNameSchema(),
+			"description":  getDataSourceDescriptionSchema(),
+			"path":         getPathSchema(),
+			"tier0_id": {
+				Type:        schema.TypeString,
+				Description: "Policy path referencing Local endpoint.",
+				Optional:    true,
+				Default:     "vmc",
+			},
+			"locale_service": {
```
annakhm, line 31: In order to be consistent, this should be `locale_service_id`.
However, we might want to use `locale_service_path` as an input, in which case adding locale_service data source might be needed. We need to consider how this resource would be used along with T0 resource/T0 data source.

nsxt/data_source_nsxt_policy_ipsec_vpn_local_endpoint.go
```diff
@@ -0,0 +1,117 @@
...
+		},
+	}
+}
+
+func dataSourceNsxtPolicyIPSecVpnLocalEndpointRead(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+	Tier0ID := d.Get("tier0_id").(string)
+	LocaleService := d.Get("locale_service").(string)
+	ServiceID := d.Get("service_id").(string)
+
```
annakhm, line 58: For new data sources, we normally use search API. This saves lines of code and a need to support pagination (which is not supported below0

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,482 @@
...
+				Description:  "Connection initiation mode used by local endpoint to establish ike connection with peer site. INITIATOR - In this mode local endpoint initiates tunnel setup and will also respond to incoming tunnel setup requests from peer gateway. RESPOND_ONLY - In this mode, local endpoint shall only respond to incoming tunnel setup requests. It shall not initiate the tunnel setup. ON_DEMAND - In this mode local endpoint will initiate tunnel creation once first packet matching the policy rule is received and will also respond to incoming initiation request.",
+				ValidateFunc: validation.StringInSlice(IPSecVpnSessionComplianceSuite, false),
+				Optional:     true,
+			},
+			"connection_initiation_mode": {
+				Type:         schema.TypeString,
+				Description:  "Compliance suite.",
+				ValidateFunc: validation.StringInSlice(IPSecVpnSessionConnectionInitiationMode, false),
+				Optional:     true,
+				Default:      "INITIATOR",
```
annakhm, line 83: model.IPSecVpnSession_CONNECTION_INITIATION_MODE_INITIATOR

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,482 @@
...
+			},
+			"authentication_mode": {
+				Type:         schema.TypeString,
+				Description:  "Peer authentication mode. PSK - In this mode a secret key shared between local and peer sites is to be used for authentication. The secret key can be a string with a maximum length of 128 characters. CERTIFICATE - In this mode a certificate defined at the global level is to be used for authentication.",
+				ValidateFunc: validation.StringInSlice(IPSecVpnSessionAuthenticationMode, false),
+				Optional:     true,
+				Default:      "PSK",
```
annakhm, line 90: model.IPSecVpnSession_AUTHENTICATION_MODE_PSK

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,482 @@
...
+				Type:        schema.TypeBool,
+				Description: "Enable/Disable IPSec VPN session.",
+				Optional:    true,
+				Default:     true,
+			},
+			"dpd_profile_path": {
+				Type:        schema.TypeString,
+				Description: "Policy path referencing Dead Peer Detection (DPD) profile. Default is set to system default profile.",
+				Optional:    true,
+				Default:     "/infra/ipsec-vpn-dpd-profiles/nsx-default-l3vpn-dpd-profile",
```
annakhm, line 102: We don't use NSX absolute paths in the provider. Is profile path required or default is taken if not provided?

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,482 @@
...
+			},
+			"psk": {
+				Type:        schema.TypeString,
+				Description: "IPSec Pre-shared key. Maximum length of this field is 128 characters.",
+				Optional:    true,
+			},
+			"peer_id": {
```
annakhm, line 109: if this is IP address, validation needs to be added

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,482 @@
...
+				Type:        schema.TypeString,
+				Description: "Peer ID to uniquely identify the peer site. The peer ID is the public IP address of the remote device terminating the VPN tunnel. When NAT is configured for the peer, enter the private IP address of the peer.",
+				Optional:    true,
+			},
+			"peer_address": {
```
annakhm, line 114: IP address validation is missing

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,482 @@
...
+				Type:        schema.TypeString,
+				Description: "Public IPV4 address of the remote device terminating the VPN connection.",
+				Optional:    true,
+			},
+			"tunnel_profile_path": {
```
annakhm, line 119: We use `getPolicyPathSchema` for path attributes (same below)

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,482 @@
...
+			"ike_profile_path": {
+				Type:        schema.TypeString,
+				Description: "Policy path referencing IKE profile.",
+				Optional:    true,
+			},
+			"tier0_id": {
+				Type:        schema.TypeString,
+				Description: "Unique identifier of the T0 resource.",
+				Optional:    true,
+				Default:     "vmc",
```
annakhm, line 138: This is not VMC-specific resource, so VMC should not appear in default
As above, lets consider `locale_service_path` here

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,482 @@
...
+}
+
+func resourceNsxtPolicyIPSecVpnSessionCreate(d *schema.ResourceData, m interface{}) error {
+
+	Tier0ID := d.Get("tier0_id").(string)
+	LocaleService := d.Get("locale_service").(string)
+	ServiceID := d.Get("service_id").(string)
+
+	// Initialize resource Id and verify this ID is not yet used
+	id := newUUID()
```
annakhm, line 359: To be consistent with other resources, we should offer `nsx_id` as argument in schema, unless there is a good reason not to

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,482 @@
...
+		sourceRanges := interface2StringList(data["sources"].(*schema.Set).List())
+		destinationRanges := interface2StringList(data["destinations"].(*schema.Set).List())
+
+		/// Source Subnets
+
+		SourceIPSecVpnSubnetList := make([]model.IPSecVpnSubnet, 0)
+		if len(sourceRanges) > 0 {
+			for _, element := range sourceRanges {
+				subnet := element
+				log.Println(subnet)
```
annakhm, line 318: this needs to be removed

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,482 @@
...
+				SourceIPSecVpnSubnetList = append(SourceIPSecVpnSubnetList, IPSecVpnSubnet)
+			}
+		}
+
+		/// Destination Subnets
+		DestinationIPSecVpnSubnetList := make([]model.IPSecVpnSubnet, 0)
+		if len(destinationRanges) > 0 {
+			for _, element := range destinationRanges {
+				subnet := element
+				log.Println(subnet)
```
annakhm, line 331: same

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,482 @@
...
+			Id:           &ruleID,
+		}
+		ruleList = append(ruleList, elem)
+	}
+	return ruleList
+}
+
+func resourceNsxtPolicyIPSecVpnSessionCreate(d *schema.ResourceData, m interface{}) error {
+
+	Tier0ID := d.Get("tier0_id").(string)
```
annakhm, line 354: according to coding conventions, local vars should start with lowercase, same below

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,453 @@
...
+	localEndpointPath := d.Get("local_endpoint_path").(string)
+	dpdProfilePath := d.Get("dpd_profile_path").(string)
+	tunnelProfilePath := d.Get("tunnel_profile_path").(string)
+	connectionInitiationMode := d.Get("connection_initiation_mode").(string)
+	authenticationMode := d.Get("authentication_mode").(string)
+	complianceSuite := d.Get("compliance_suite").(string)
+	prefixLengh := int64(d.Get("prefix_length").(int))
+	enabled := d.Get("enabled").(bool)
+
+	if ResourceType == "RouteBasedIPSecVpnSession" {
```
annakhm, line 175: shouldn't this be `resourceType`?

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,453 @@
...
+		}
+
+		dataValue, err := converter.ConvertToVapi(routeObj, model.RouteBasedIPSecVpnSessionBindingType())
+		if err != nil {
+			return nil, err[0]
+		}
+		return dataValue.(*data.StructValue), nil
+	}
+
+	if ResourceType == "PolicyBasedIPSecVpnSession" {
```
annakhm, line 217: ditto

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,453 @@
...
+
+	if err != nil {
+		if isNotFoundError(err) {
+			d.SetId("")
+			return nil
+		}
+		return handleReadError(d, "VPN Session", id, err)
+	}
+
+	interfaceVpn, errs := converter.ConvertToGolang(obj, model.RouteBasedIPSecVpnSessionBindingType())
```
annakhm, line 386: Session based VPN should be supported here as well

nsxt/data_source_ipsec_vpn_service.go
```diff
@@ -0,0 +1,48 @@
+/* Copyright © 2022 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
```
annakhm, line 3: package in missing here?

nsxt/data_source_ipsec_vpn_service.go
```diff
@@ -0,0 +1,48 @@
...
+func dataSourceNsxtPolicyIPSecVpnService() **schema.Resource {
```
annakhm, line 4: do we need the extra `*`?

annakhm:
I think several documentation files are missing

nsxt/resource_nsxt_policy_ipsec_vpn_service.go
```diff
@@ -0,0 +1,242 @@
...
+			"revision":     getRevisionSchema(),
+			"tag":          getTagsSchema(),
+			"gateway_path": getPolicyGatewayPathSchema(),
+			"enabled": {
+				Type:        schema.TypeBool,
+				Description: "Enable/Disable IPSec VPN service.",
+				Optional:    true,
+				Default:     true,
+			},
+			"haSync": {
```
annakhm, line 50: please use snake case naming convention here and below - `ha_sync`, `bypass_rule`..

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,451 @@
...
+			for _, element := range destinationRanges {
+				subnet := element
+				IPSecVpnSubnet := model.IPSecVpnSubnet{
+					Subnet: &subnet,
+				}
+				DestinationIPSecVpnSubnetList = append(DestinationIPSecVpnSubnetList, IPSecVpnSubnet)
+			}
+		}
+
+		ruleID := newUUID()
```
annakhm, line 317: Updating rules every time with new uuid might not be a good idea. Those rules might have internal statistics that would be reset with every update. It would be better to keep id from previous update.

nsxt/data_source_ipsec_vpn_service.go
```diff
@@ -0,0 +1,55 @@
...
+func dataSourceNsxtPolicyIPSecVpnService() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNsxtPolicyIPSecVpnServiceRead,
+
+		Schema: map[string]*schema.Schema{
+			"nsx_id":       getNsxIDSchema(),
+			"path":         getPathSchema(),
+			"display_name": getDisplayNameSchema(),
+			"description":  getDescriptionSchema(),
+			"revision":     getRevisionSchema(),
```
annakhm, line 20: for data sources, unless there is a specific reason, we only expose `id`, `display_name` as input args and `path` and `description` as output args. The rest of attributes defined below are never set as far as I can see.

nsxt/data_source_nsxt_policy_ipsec_vpn_local_endpoint.go
```diff
@@ -0,0 +1,55 @@
...
+func dataSourceNsxtPolicyIPSecVpnLocalEndpoint() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNsxtPolicyIPSecVpnLocalEndpointRead,
+
+		Schema: map[string]*schema.Schema{
+			"id":           getDataSourceIDSchema(),
+			"display_name": getDataSourceDisplayNameSchema(),
+			"description":  getDataSourceDescriptionSchema(),
+			"path":         getPathSchema(),
+			"tier0_id": {
```
annakhm, line 19: ditto

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,494 @@
...
+				Type:        schema.TypeString,
+				Description: "Unique identifier of the T0 resource.",
+				Optional:    true,
+			},
+			"locale_service": {
+				Type:        schema.TypeString,
+				Description: "Unique identifier of the Locale Service resource.",
+				Optional:    true,
+			},
+			"service_id": {
```
annakhm, line 130: To be consistent with provider convention, we should use `service_path` rather than `tier0_id`, `locale_service`, `service_id`. This way user can define service resource and reference it in subsequent VPN config, like here.

nsxt/data_source_nsxt_policy_ipsec_vpn_service.go
```diff
@@ -0,0 +1,55 @@
...
+				Optional:    true,
+				Default:     true,
+			},
+			"ike_loglevel": {
+				Type:         schema.TypeString,
+				Description:  "Log level for internet key exchange (IKE).",
+				ValidateFunc: validation.StringInSlice(IPSecVpnServiceIkeLogLevelTypes, false),
+				Optional:     true,
+			},
+			"bypass_rules": getIPSecVPNRulesSchema(),
```
annakhm, line 41: as per convention, this should be singular -  `bypass_rule`, since each block would define a single rule

nsxt/resource_nsxt_policy_ipsec_vpn_service.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+func resourceNsxtPolicyIPSecVpnService() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNsxtPolicyIPSecVpnServiceCreate,
+		Read:   resourceNsxtPolicyIPSecVpnServiceRead,
+		Update: resourceNsxtPolicyIPSecVpnServiceUpdate,
+		Delete: resourceNsxtPolicyIPSecVpnServiceDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
```
annakhm, line 33: `ImportStatePassthrough` works well when there is a single object ID that can serve as basis for Read function (behind the scenes, import calls the Read function and initializes the object). For resources in this PR, however, single ID is not good enough, since gwID and sometimes more IDs are required for the Get API call. In such cases, we use custom import functions (example https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/resource_nsxt_policy_nat_rule.go#L45)

annakhm:
Lets be consistent in file naming - lets follow `ipsec_vpn` convention this one is as already present in the repo.

nsxt/resource_nsxt_policy_ipsec_vpn_service_test.go
```diff
@@ -0,0 +1,193 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+var accTestPolicyIPSecVpnServiceCreateAttributes = map[string]string{
+	"display_name":        getAccTestResourceName(),
+	"description":         "terraform created",
+	"locale_service_path": "/infra/tier-0s/test-t0/locale-services/default",
```
annakhm, line 17: The T0 gateway needs to be created within the test, since we can't rely on test-t0 to be pre-created on the testbed. We usually just configure the resources the test depends on as part of the template.

2ez4szliu: I've changed the code to use the EdgeCluster and Tier0 template in the test_utils files, and get the locale_service_path from the tier0.

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,630 @@
...
+		setPolicyTagsInSchema(d, blockVPN.Tags)
+		d.Set("nsx_id", blockVPN.Id)
+		d.Set("path", blockVPN.Path)
+		d.Set("revision", blockVPN.Revision)
+		d.Set("vpn_type", blockVPN.ResourceType)
+		d.Set("authentication_mode", blockVPN.AuthenticationMode)
+		d.Set("compliance_suite", blockVPN.ComplianceSuite)
+		d.Set("connection_initiation_mode", blockVPN.ConnectionInitiationMode)
+		d.Set("enabled", blockVPN.Enabled)
+		if blockVPN.IkeProfilePath != nil {
```
annakhm, line 497: `d.Set` can handle nil value, so its safe to remove the check

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,630 @@
...
+				Optional:     true,
+				Default:      model.IPSecVpnSession_AUTHENTICATION_MODE_PSK,
+			},
+			"enabled": {
+				Type:        schema.TypeBool,
+				Description: "Enable/Disable IPSec VPN session.",
+				Optional:    true,
+				Default:     true,
+			},
+			"psk": {
```
annakhm, line 105: is this sensitive? If so we should use `Sensitive:   true`

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,630 @@
...
+				Type:        schema.TypeString,
+				Description: "IPSec Pre-shared key. Maximum length of this field is 128 characters.",
+				Optional:    true,
+				Default:     "DefaultPSK",
```
annakhm, line 109: does this default make sense?

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
2ez4szliu: Done

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
2ez4szliu: Added.

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
2ez4szliu: Removed

nsxt/utils_test.go
```diff
@@ -445,6 +445,19 @@ resource "nsxt_policy_tier%s_gateway" "test" {
 }`, tier, tier, edgeClusterName, haMode)
 }
 
+func testAccNsxtPolicyTier0WithEdgeClusterForVPN(edgeClusterName string) string {
+	// VPN is supported only on Tier-0 with ACTIVE-STANDBY HA mode
+	return fmt.Sprintf(`
+resource "nsxt_policy_tier0_gateway" "test" {
+  display_name = "terraform-t0-gw-for-vpn"
```
annakhm, line 452: It is preferrable to use `getAccTestResourceName` to avoid collisions between different tests running simultaneously.

nsxt/resource_nsxt_policy_ipsec_vpn_service.go
```diff
@@ -0,0 +1,324 @@
...
+				Default:     true,
+			},
+			"ike_log_level": {
+				Type:         schema.TypeString,
+				Description:  "Log level for internet key exchange (IKE).",
+				ValidateFunc: validation.StringInSlice(IPSecVpnServiceIkeLogLevelTypes, false),
+				Optional:     true,
+				Default:      model.IPSecVpnService_IKE_LOG_LEVEL_INFO,
+			},
+			"bypass_rules": getIPSecVPNRulesSchema(),
```
annakhm, line 67: please change this to `bypass_rule` per convention

nsxt/resource_nsxt_policy_ipsec_vpn_service.go
```diff
@@ -0,0 +1,324 @@
...
+		var ruleList []model.IPSecVpnRule
+		for _, rule := range rules {
+			data := rule.(map[string]interface{})
+			action := data["action"].(string)
+			sourceRanges := interface2StringList(data["sources"].(*schema.Set).List())
+			destinationRanges := interface2StringList(data["destinations"].(*schema.Set).List())
+
+			/// Source Subnets
+
+			SourceIPSecVpnSubnetList := make([]model.IPSecVpnSubnet, 0)
```
annakhm, line 138: we follow golang convention to start variables with lowercase (same below)

nsxt/resource_nsxt_policy_ipsec_vpn_service.go
```diff
@@ -0,0 +1,324 @@
...
+	d.Set("display_name", obj.DisplayName)
+	d.Set("description", obj.Description)
+	setPolicyTagsInSchema(d, obj.Tags)
+	d.Set("nsx_id", id)
+	d.Set("path", obj.Path)
+	d.Set("revision", obj.Revision)
+	d.Set("enabled", obj.Enabled)
+	d.Set("ha_sync", obj.HaSync)
+
+	if obj.BypassRules != nil {
```
annakhm, line 203: if rules are deleted, we need to make sure they are deleted in schema, so I don't think this `if` is needed?

nsxt/resource_nsxt_policy_ipsec_vpn_service.go
```diff
@@ -0,0 +1,324 @@
...
+	if ikeLogLevel != "" {
+		ipSecVpnService.IkeLogLevel = &ikeLogLevel
+	}
+
+	log.Printf("[INFO] Updating IPSecVpnService with ID %s", id)
+	err = patchNsxtPolicyIPSecVpnService(connector, gwID, localeServiceID, ipSecVpnService, isT0)
+	if err != nil {
+		return handleUpdateError("IPSecVpnService", id, err)
+	}
+	d.SetId(id)
```
annakhm, line 302: `setId` is not needed in `Update` since it can't change

nsxt/resource_nsxt_policy_ipsec_vpn_service_test.go
```diff
@@ -0,0 +1,193 @@
...
+	"ha_sync":       "true",
+	"ike_log_level": "INFO",
+	"sources":       "192.168.10.0/24",
+	"destinations":  "192.169.10.0/24",
+	"action":        "BYPASS",
+}
+
+var accTestPolicyIPSecVpnServiceUpdateAttributes = map[string]string{
+	"display_name":  getAccTestResourceName(),
+	"description":   "terraform created",
```
annakhm, line 27: can we change the description here to "updated"?

nsxt/resource_nsxt_policy_ipsec_vpn_service_test.go
```diff
@@ -0,0 +1,193 @@
...
+	"enabled":       "true",
```
annakhm, line 28: If its possible to change values of `enabled` and `action`, it would be preferrable, since we're testing update. But if it creates validation errors on platform then lets leave as is.

nsxt/resource_nsxt_policy_ipsec_vpn_service_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t); testAccOnlyLocalManager(t) },
+		Providers: testAccProviders,
+		CheckDestroy: func(state *terraform.State) error {
+			return testAccNsxtPolicyIPSecVpnServiceCheckDestroy(state, accTestPolicyIPSecVpnServiceUpdateAttributes["display_name"])
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: testAccNsxtPolicyIPSecVpnServiceMinimalistic(),
```
annakhm, line 47: It would be best to end the test with minimalistic config, since it validates unneeded attributes are deleted or reset back to default, which reveals bugs quite often.

nsxt/resource_nsxt_policy_ipsec_vpn_service_test.go
```diff
@@ -0,0 +1,193 @@
...
+	var attrMap map[string]string
+	if createFlow {
+		attrMap = accTestPolicyIPSecVpnServiceCreateAttributes
+	} else {
+		attrMap = accTestPolicyIPSecVpnServiceUpdateAttributes
+	}
+	return testAccNsxtPolicyEdgeClusterReadTemplate(getEdgeClusterName()) +
+		testAccNsxtPolicyTier0WithEdgeClusterForVPN("test") + fmt.Sprintf(`
+   resource "nsxt_policy_ipsec_vpn_service" "test" {
+	 display_name                   = "%s"
```
annakhm, line 167: cosmetic comment - please align attributes

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,623 @@
...
+				ValidateFunc: validation.StringInSlice(IPSecVpnSessionResourceType, false),
+				Required:     true,
+			},
+			"compliance_suite": {
+				Type:         schema.TypeString,
+				Description:  "Connection initiation mode used by local endpoint to establish ike connection with peer site. INITIATOR - In this mode local endpoint initiates tunnel setup and will also respond to incoming tunnel setup requests from peer gateway. RESPOND_ONLY - In this mode, local endpoint shall only respond to incoming tunnel setup requests. It shall not initiate the tunnel setup. ON_DEMAND - In this mode local endpoint will initiate tunnel creation once first packet matching the policy rule is received and will also respond to incoming initiation request.",
```
annakhm, line 89: this description should go to next attr, and vice versa

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,623 @@
...
+		Schema: map[string]*schema.Schema{
+			"nsx_id":              getNsxIDSchema(),
+			"path":                getPathSchema(),
+			"display_name":        getDisplayNameSchema(),
+			"description":         getDescriptionSchema(),
+			"revision":            getRevisionSchema(),
+			"tag":                 getTagsSchema(),
+			"tunnel_profile_path": getPolicyPathSchema(true, false, "Policy path referencing Tunnel profile to be used."),
+			"local_endpoint_path": getPolicyPathSchema(true, false, "Policy path referencing Local endpoint."),
+			"ike_profile_path": {
```
annakhm, line 70: lets use `getPolicyPathSchema` here too?

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,623 @@
...
+				Type:        schema.TypeString,
+				Description: "Policy path referencing IKE profile.",
+				Optional:    true,
+			},
+			"dpd_profile_path": {
```
annakhm, line 75: We can use `getComputedPolicyPathSchema` here

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,623 @@
...
+				ValidateFunc: validation.IsIPv4Address,
+			},
+			"service_path": getPolicyPathSchema(true, true, "Policy path for IPSec VPN service"),
+			"subnets": {
+				Type:        schema.TypeList,
+				Description: "IP Tunnel interface (commonly referred as VTI) subnet.",
+				Optional:    true,
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validateSingleIP(),
```
annakhm, line 139: I'm a bit confused here - is this a subnet or IP?

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,623 @@
...
+		}
+		dataValue, err := converter.ConvertToVapi(policyObj, model.PolicyBasedIPSecVpnSessionBindingType())
+		if err != nil {
+			return nil, err[0]
+		}
+		return dataValue.(*data.StructValue), nil
+
+	}
+
+	var err error
```
annakhm, line 261: Although this should never to happen, Iet's give the error some substance.

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
```diff
@@ -0,0 +1,623 @@
...
+				Description: "Policy path referencing IKE profile.",
+				Optional:    true,
+			},
+			"dpd_profile_path": {
+				Type:        schema.TypeString,
+				Description: "Policy path referencing Dead Peer Detection (DPD) profile. Default is set to system default profile.",
+				Optional:    true,
+				Computed:    true,
+			},
+			"vpn_type": {
```
annakhm, line 81: We try to use human-friendly strings in the provider, and `RouteBasedIPSecVpnSession` is too long, people will curse us :) I'd suggest to go with `RouteBased` and `PolicyBased`, even if it means a bit more code.

nsxt/resource_nsxt_policy_ipsec_vpn_session_test.go
```diff
@@ -0,0 +1,404 @@
...
+	"prefix_length":              "24",
+	"peer_address":               "18.18.18.22",
+	"peer_id":                    "18.18.18.22",
+	"psk":                        "VMware123!",
+	"connection_initiation_mode": "INITIATOR",
+}
+
+var accTestPolicyIPSecVpnSessionRouteBasedUpdateAttributes = map[string]string{
+	"display_name":               getAccTestResourceName(),
+	"description":                "Terraform-provisioned IPsec Route-Based VPN",
```
annakhm, line 31: as above, it would be preferrable to add some more variation to update, so that the test has more coverage

nsxt/resource_nsxt_policy_ipsec_vpn_session_test.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+	return testAccNsxtPolicyEdgeClusterReadTemplate(getEdgeClusterName()) +
+		testAccNsxtPolicyTier0WithEdgeClusterForVPN("test") +
+		testAccNsxtPolicyIPSecVpnSessionPreConditionTemplate() +
+		fmt.Sprintf(`
+resource "nsxt_policy_ipsec_vpn_session" "test" {
+	display_name               = "%s"
+	description                = "%s"
+	ike_profile_path           = nsxt_policy_ipsec_vpn_ike_profile.test.path
+	tunnel_profile_path        = nsxt_policy_ipsec_vpn_tunnel_profile.test.path
```
annakhm, line 339: please fix alignment here and below

nsxt/resource_nsxt_policy_l2_vpn_service.go
```diff
@@ -0,0 +1,247 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"nsx_id":       getNsxIDSchema(),
+			"path":         getPathSchema(),
+			"display_name": getDisplayNameSchema(),
+			"description":  getDescriptionSchema(),
+			"revision":     getRevisionSchema(),
+			"tag":          getTagsSchema(),
+			"locale_service_path": {
```
annakhm, line 40: `getPolicyPathSchema` to save lines of code and have validation

nsxt/resource_nsxt_policy_l2_vpn_service.go
```diff
@@ -0,0 +1,247 @@
...
+	d.Set("display_name", obj.DisplayName)
+	d.Set("description", obj.Description)
+	setPolicyTagsInSchema(d, obj.Tags)
+	d.Set("nsx_id", id)
+	d.Set("path", obj.Path)
+	d.Set("revision", obj.Revision)
+	d.Set("enable_hub", obj.EnableHub)
+	d.Set("mode", obj.Mode)
+
+	if obj.EncapIpPool != nil {
```
annakhm, line 124: would this property be reset in state if EncapIpPool is removed?

nsxt/resource_nsxt_policy_l2_vpn_service.go
```diff
@@ -0,0 +1,247 @@
...
+	if len(encapIPPool) != 0 {
+		l2VpnService.EncapIpPool = encapIPPool
+	}
+
+	log.Printf("[INFO] Updating L2VpnService with ID %s", id)
+	err = patchNsxtPolicyL2VpnService(connector, gwID, localeServiceID, l2VpnService, isT0)
+	if err != nil {
+		return handleUpdateError("L2VpnService", id, err)
+	}
+	d.SetId(id)
```
annakhm, line 225: not needed in Update

nsxt/resource_nsxt_policy_l2_vpn_service.go
```diff
@@ -0,0 +1,247 @@
...
+
+	l2VpnService := model.L2VPNService{
+		Id:          &id,
+		DisplayName: &displayName,
+		Description: &description,
+		Tags:        tags,
+		EnableHub:   &enableHub,
+		Mode:        &mode,
+	}
+
```
annakhm, line 210: revision seems to be missing here

nsxt/resource_nsxt_policy_l2_vpn_session.go
```diff
@@ -0,0 +1,227 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"nsx_id":       getNsxIDSchema(),
+			"path":         getPathSchema(),
+			"display_name": getDisplayNameSchema(),
+			"description":  getDescriptionSchema(),
+			"revision":     getRevisionSchema(),
+			"tag":          getTagsSchema(),
+			"service_path": getPolicyPathSchema(true, true, "Policy path for L2 VPN service"),
+			"transport_tunnels": {
```
annakhm, line 34: Sorry if I'm missing something, but shouldn't there be more attributes here? I see some in both UI and the SDK under `L2VPNSession`

nsxt/resource_nsxt_policy_l2_vpn_session.go
```diff
@@ -0,0 +1,227 @@
...
+				Type:        schema.TypeList,
+				Description: "List of transport tunnels(vpn sessions path) for redundancy",
+				Required:    true,
+				Elem: &schema.Schema{
```
annakhm, line 38: If this is policy path, lets use PolicyPathSchema here

nsxt/resource_nsxt_policy_l2_vpn_session.go
2ez4szliu: yes you are right, I just added the mss and direction attribute for TcpMssClamping and also local/peer address and protocol for its TunnelEncapsulation

nsxt/resource_nsxt_policy_ipsec_vpn_service_test.go
2ez4szliu: `enabled` can be configured, however seems action for IpsecVpnService can only be `BYPASS`, `PROTECT` only works for IpsecVpnSession

nsxt/resource_nsxt_policy_ipsec_vpn_session.go
2ez4szliu: this is the ipAddresses of the TunnelInterfaceIPSubnet, together with prefix_length they form the IPSubnets.
I've changed it into ip_addresses to avoid confusion

website/docs/r/policy_ipsec_vpn_service.html.markdown
```diff
@@ -0,0 +1,64 @@
...
+
+* `display_name` - (Required) Display name of the resource.
+* `description` - (Optional) Description of the resource.
+* `tag` - (Optional) A list of scope + tag pairs to associate with this resource.
+* `nsx_id` - (Optional) The NSX ID of this resource. If set, this ID will be used to create the resource.
+* `locale_service_path` - (Required) Path of the locale_service associated with the IPSec VPN Service.
+* `enabled` - (Optional) Whether this IPSec VPN Service is enabled. Default is `true`.
+* `ha_sync` - (Optional) Enable/Disable IPSec VPN service HA state sync. Default is `true`.
+* `ike_log_level` - (Optional) Set of algorithms to be used for message digest during IKE negotiation. Value is one of `DEBUG`, `INFO`, `WARN`, `ERROR` and `EMERGENCY`. Default is `INFO`. 
+* `bypass_rules` - (Optional) Set the bypass rules for this IPSec VPN Service. `sources` and `destinations` are list of `ipv4-cidr-block`, `Action` value must be set to `BYPASS` or `PROTECT`.
```
annakhm, line 44: bypass_rule here

annakhm, line 44: bypass_rule

website/docs/r/policy_l2_vpn_service.html.markdown
```diff
@@ -0,0 +1,59 @@
...
+
+This resource is applicable to NSX Policy Manager and VMC.
+
+## Example Usage
+
+```hcl
+resource "nsxt_policy_l2_vpn_service" "test" {
+  display_name          = "l2-vpn-service1"
+  description           = "Terraform provisioned L2 VPN service"
+  locale_service_path   = /infra/tier-0s/aaa/locale-services/default
```
annakhm, line 20: We normally don't suggest to use absolute path strings in documentation, but rather things like `data.nsxt_policy_gateway_locale_service.path`. Regardless, string needs to be quoted.

website/docs/r/policy_ipsec_vpn_service.html.markdown
```diff
@@ -0,0 +1,64 @@
...
+
+* `display_name` - (Required) Display name of the resource.
+* `description` - (Optional) Description of the resource.
+* `tag` - (Optional) A list of scope + tag pairs to associate with this resource.
+* `nsx_id` - (Optional) The NSX ID of this resource. If set, this ID will be used to create the resource.
+* `locale_service_path` - (Required) Path of the locale_service associated with the IPSec VPN Service.
+* `enabled` - (Optional) Whether this IPSec VPN Service is enabled. Default is `true`.
+* `ha_sync` - (Optional) Enable/Disable IPSec VPN service HA state sync. Default is `true`.
+* `ike_log_level` - (Optional) Set of algorithms to be used for message digest during IKE negotiation. Value is one of `DEBUG`, `INFO`, `WARN`, `ERROR` and `EMERGENCY`. Default is `INFO`. 
+* `bypass_rule` - (Optional) Set the bypass rules for this IPSec VPN Service. `sources` and `destinations` are list of `ipv4-cidr-block`, `Action` value must be set to `BYPASS` or `PROTECT`.
```
annakhm, line 44: bypass_rule properties need to be documented in separate lines (with 2 space indentation)

2ez4szliu: done. Also added same documentation to ipsec_vpn_session

website/docs/r/policy_l2_vpn_service.html.markdown
2ez4szliu: done

annakhm (approved):
The rest of comments will be addressed in follow up PR

PR #774, opened 2022-07-29
Fix change detection in lb virtual server rules

annakhm:
We should perform change based on change in any rule, as
opposed to looking only at rule count.


2ez4szliu (approved):

ksamoray (approved):
LGTM

PR #779, opened 2022-08-05
Support replication mode in Segment resources

annakhm:
This attribute was added in NSX 3.0.0
In addition, remove cidr requirement for segment subnet, as
its no longer required on the platform.


2ez4szliu (approved):

coteforbes:
Can we update the documentation in the registry to reflect this change?  Thanks!

annakhm:
Yes @coteforbes, we expect to release provider containing this fix next month

PR #795, opened 2022-09-28
Make sure to path sequence id in FW rules

annakhm:
Otherwise rule order may be compromised.

### DIFF
--- a/nsxt/policy_common.go
+++ b/nsxt/policy_common.go
@@ -393,7 +393,6 @@ func setPolicyRulesInSchema(d *schema.ResourceData, rules []model.Rule) error {
 func getPolicyRulesFromSchema(d *schema.ResourceData) []model.Rule {
 	rules := d.Get("rule").([]interface{})
 	var ruleList []model.Rule
-	seq := 0
 	for _, rule := range rules {
 		data := rule.(map[string]interface{})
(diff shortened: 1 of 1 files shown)


nsxt/policy_common.go
```diff
@@ -437,7 +437,10 @@ func getPolicyRulesFromSchema(d *schema.ResourceData) []model.Rule {
 			Services:             getPathListFromMap(data, "services"),
 			Scope:                getPathListFromMap(data, "scope"),
 			Profiles:             getPathListFromMap(data, "profiles"),
-			SequenceNumber:       &sequenceNumber,
+		}
+
+		if sequenceNumber > 0 {
+			elem.SequenceNumber = &sequenceNumber
```
ksamoray, line 443: So what would be the value of elem.SequenceNumber if sequenceNumber == 0? Wouldn't it be the same?
Should this just protect from negative values in sequenceNumber?

ksamoray (approved):
LGTM, just a nit/curiosity

nsxt/policy_common.go
annakhm: The value will be nil, which translates to not passing the value in request body. In this case NSX assigns the value, which is what we want.

PR #797, opened 2022-10-06
Remove test for default GW policy data source

annakhm:
This test was only relevant for early NSX versions, and VPN configuration breaks the tests since one extra GW policy is added.

### DIFF
--- a/nsxt/policy_common.go
+++ b/nsxt/policy_common.go
@@ -452,7 +452,6 @@ func getPolicyRulesFromSchema(d *schema.ResourceData) []model.Rule {
 		}
 
 		ruleList = append(ruleList, elem)
-		seq = seq + 1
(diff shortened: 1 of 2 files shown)


nsxt/policy_common.go
```diff
@@ -452,7 +452,6 @@ func getPolicyRulesFromSchema(d *schema.ResourceData) []model.Rule {
 		}
 
 		ruleList = append(ruleList, elem)
-		seq = seq + 1
```
ksamoray, line 455: Is this related to the test removal? Or just piggybacking on the other change?

annakhm: no, unrelated, just some code cleaning

2ez4szliu (approved):

PR #805, opened 2022-10-20
Disallow setting nsx_id attribute in policy rules

annakhm:
nsx_id should only be configurable in predefined policy resources, and read-only in other resources.


ksamoray:
Should this (or could it) be UT somehow? Otherwise LGTM

annakhm:
> Should this (or could it) be UT somehow? Otherwise LGTM

We usually don't test terraform core functionality, including negative tests for read only attrs.

ksamoray (approved):

PR #807, opened 2022-10-25
Doc fix for vpn session resources

2ez4szliu:

### DIFF
--- a/website/docs/r/policy_ipsec_vpn_session.markdown
+++ b/website/docs/r/policy_ipsec_vpn_session.markdown
@@ -71,7 +71,7 @@ The following arguments are supported:
 * `enabled` - (Optional) Boolean. Enable/Disable IPsec VPN session. Default is "true" (session enabled).
 * `service_path` - (Required) The path of the IPSec VPN service for the VPN session.
 * `dpd_profile_path` - (Optional) Policy path referencing Dead Peer Detection (DPD) profile. Default is set to system default profile.
-* `vpn_type` - (Required) `RouteBased` or `PolicyBased`. Policy Based VPN requires to define protect rules that match local and peer subnets. IPSec security associations is negotiated for each pair of local and peer subnet. A Route Based VPN is more flexible, more powerful and recommended over policy based VPN. IP Tunnel port is created and all traffic routed via tunnel port is protected. Routes can be configured statically or can be learned through BGP. A route based VPN is must for establishing redundant VPN session to remote site.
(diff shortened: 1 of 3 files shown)


website/docs/r/policy_ipsec_vpn_session.markdown
```diff
@@ -80,9 +80,12 @@ The following arguments are supported:
 * `peer_id` - (Optional) Peer ID to uniquely identify the peer site. The peer ID is the public IP address of the remote device terminating the VPN tunnel. When NAT is configured for the peer, enter the private IP address of the peer.
 * `local_endpoint_path` - (Required) Policy path referencing Local endpoint. In VMC, Local Endpoints are pre-configured the user can refer to their path using `data nsxt_policy_ipsec_vpn_local_endpoint` and using the "Private IP1" or "Public IP1" values to refer to the private and public endpoints respectively.
 * `rule` - (Optional) Bypass rules for this IPSec VPN Session. `sources` and `destinations` are list of `ipv4-cidr-block`, `Action` value must be set to `BYPASS` or `PROTECT`. Only applied for Policy Based VPN Session. 
+  * `nsx_id` - (Optional) The NSX ID of the bypass rule for this IPSec VPN Session. If set, this ID will be used to create the resource.
```
annakhm, line 83: Allowing to specify `nsx_id` in sub-resource (rule in this case) has potential for issues, since there is no `ForceNew` functionality specific to the sub-resource. It would be painful to implement partial update for rules if we choose to do so - we'll need to implement `ForceNew` functionality ourselves on rule level. To avoid these kind of challenges, I suggest to make `nsx_id` read-only.

website/docs/r/policy_ipsec_vpn_session.markdown
```diff
@@ -71,7 +71,7 @@ The following arguments are supported:
...
+* `vpn_type` - (Required) `RouteBased` or `PolicyBased`. Policy Based VPN requires to define protect rules that match local and peer subnets. IPSec security associations is negotiated for each pair of local and peer subnet. For PolicyBased Session, `rule` must be specified with `sources`, `destination` and `action`. A Route Based VPN is more flexible, more powerful and recommended over policy based VPN. IP Tunnel port is created and all traffic routed via tunnel port is protected. Routes can be configured statically or can be learned through BGP. A route based VPN is must for establishing redundant VPN session to remote site. For RouteBased VPN session, `ip_addresses` amd `prefix_length` must be specified to create the tunnel interface and its subnet.
```
annakhm, line 74: some nits

`associations` -> `association`  
`amd` -> `and`
`A route based VPN is must` -> `A route based VPN is a must`

website/docs/r/policy_ipsec_vpn_session.markdown
2ez4szliu: Thanks Anna, i've changed the nsx_id in IPSecVpnRule schema to use getFlexNsxIDSchema and set readOnly to true

website/docs/r/policy_ipsec_vpn_session.markdown
2ez4szliu: Fixed

website/docs/r/policy_ipsec_vpn_session.markdown
annakhm: Thanks, consequently this documentation line needs to be removed or moved to `attributes` under `rule`

annakhm (approved):

PR #812, opened 2022-11-03
Tier-1s support for VPN sessions

2ez4szliu:


nsxt/utils_test.go
```diff
@@ -459,6 +458,23 @@ resource "nsxt_policy_tier0_gateway" "test" {
...
+	ha_mode = "ACTIVE_STANDBY"
+}
+resource "nsxt_policy_tier1_gateway" "test" {
+	description               = "Acceptance Test"
+	display_name              = "%s"
+	locale_service {
+		edge_cluster_path = data.nsxt_policy_edge_cluster.%s.path
+	}
+	tier0_path                = nsxt_policy_tier0_gateway.test.path
+}`, getAccTestResourceName(), getAccTestResourceName(), edgeClusterName)
```
annakhm, line 475: If I'm not mistaken, using `getAccTestResourceName()` here would mean T0 and T1 names will change between test steps. While this does not affect test result, its an extra update that would increase overall test run time unnecessarily. I suggest assigning random gateway names only once, store these in variables and use those variables here. I believe same issue is present in other VPN tests.

nsxt/utils_test.go
```diff
@@ -459,6 +458,23 @@ resource "nsxt_policy_tier0_gateway" "test" {
 }`, getAccTestResourceName(), edgeClusterName)
 }
 
+func testAccNsxtPolicyTier1WithEdgeClusterForVPN(edgeClusterName string) string {
```
annakhm, line 461: do we need `edgeClusterName` parameter here? I don't think its ever anything other than `test`  (same for `testAccNsxtPolicyTier0WithEdgeClusterForVPN`)

nsxt/resource_nsxt_policy_ipsec_vpn_session_test.go
```diff
@@ -489,3 +632,84 @@ resource "nsxt_policy_ipsec_vpn_session" "test" {
...
+	tag {
+	scope = "scope1"
+	tag   = "tag1"
+	  }
+}`, attrMap["display_name"], attrMap["description"], attrMap["enabled"], attrMap["vpn_type"],
+			attrMap["authentication_mode"], attrMap["compliance_suite"], attrMap["peer_address"], attrMap["peer_id"],
+			attrMap["psk"], attrMap["connection_initiation_mode"], attrMap["sources"], attrMap["destinations"], attrMap["action"])
+}
+
+func testAccNsxtPolicyIPSecVpnSessionRouteBasedTemplateTier1(createFlow bool) string {
```
annakhm, line 679: This seems very similar to `testAccNsxtPolicyIPSecVpnSessionRouteBasedTemplate`, could we avoid the code duplication with extra parameter?

nsxt/utils_test.go
2ez4szliu: I agree, but it seems we have several function using the edgeClusterName parameter, and they are used many times, I think we can change this function definition in a seperate PR

nsxt/resource_nsxt_policy_ipsec_vpn_session_test.go
2ez4szliu: done

nsxt/utils_test.go
annakhm: sounds good

nsxt/resource_nsxt_policy_l2_vpn_session_test.go
```diff
@@ -207,13 +255,40 @@ func testAccNsxtPolicyL2VpnSessionCheckDestroy(state *terraform.State, displayNa
 	return nil
 }
 
-func testAccNsxtPolicyL2VpnSessionPreConditionTemplate() string {
-	return fmt.Sprintf(`
+func testAccNsxtPolicyL2VpnSessionPreConditionTemplate(isT0 bool) string {
+	resourceName := getAccTestResourceName()
```
annakhm, line 259: `resourceName` would still be different across different invocations of  `testAccNsxtPolicyL2VpnSessionPreConditionTemplate`, correct? This means that terraform will perform update for parent resources. My suggestion was to refer to a global variable that would only be initialized once.

annakhm (approved):

PR #814, opened 2022-11-04
Deprecate MP objects

annakhm:


GNUmakefile
```diff
@@ -66,5 +66,8 @@ website-lint-fix:
 	@misspell -w -source=text website/
 	@terrafmt fmt ./website --pattern '*.markdown'
 
+website-list-category:
```
ksamoray, line 69: Is that related to the deprecation? Or just piggybacking?

ksamoray:
data_source_nsxt_provider_info.go
resource_nsxt_vlan_logical_switch.go

Shouldn't these be deprecated too?

GNUmakefile
annakhm: Sort of related - this verifies website categories are consistent. I'll add it to commit description

annakhm:
> data_source_nsxt_provider_info.go resource_nsxt_vlan_logical_switch.go
> 
> Shouldn't these be deprecated too?

Thanks for noticing the vlan switch resource, added. provider info doesn't actually use MP api.

ksamoray (approved):

Issue #699, opened 2021-12-16
Add resource for 'user' created FQDN's under Context Profiles


tullydwyer:
Hi there,

We are trying to use 'user' created FQDN's in Context Profiles as per https://github.com/vmware/terraform-provider-nsxt/issues/698 . We are not able to find a resource for creating the FQDN's via the NSX-T provider.

Are we able to have a resource for creating 'user' created FQDN's the same as the web console functionality?

<img width="727" alt="Screen Shot 2021-12-16 at 3 20 07 pm" src="https://user-images.githubusercontent.com/7508115/146317081-1ebb3961-cfa3-46fd-963b-7df2fc8543f3.png">

### Is your feature request related to a problem? Please describe.

Yes, https://github.com/vmware/terraform-provider-nsxt/issues/698 . And being able to manage these FQDN's via Terraform.

### Describe the solution you'd like

A resource to create the FQDN's the same as the web console, e.g.:

<img width="736" alt="Screen Shot 2021-12-16 at 3 54 11 pm" src="https://user-images.githubusercontent.com/7508115/146317527-b4493058-3ecd-40d0-9c92-2905b6ac036d.png">

```terraform
resource "nsxt_policy_context_profile_fqdn" "example" {
  domain = "bob.bob.com"
} 
```

Output:
<img width="727" alt="Screen Shot 2021-12-16 at 3 20 07 pm" src="https://user-images.githubusercontent.com/7508115/146317485-8f9d4352-97b7-4423-93db-ff6edc227f7a.png">

Issue #721, opened 2022-03-04
Add a resource to deploy custom (user created) FQDNs in Context Profile


juanjojulian:
### Is your feature request related to a problem? Please describe.

I would like to be able to add Custom FQDNs to Context Profile in order to consume them afterwards from nsxt_policy_context_profile (support for user created FQDNs needs to be added to this resource also). The objective is to manage a nsxt_policy_security_policy with two rules that will act as a transparent proxy where the Custom FQDNs will be the Allow List.

### Describe the solution you'd like

create a "nsxt_policy_context_profile_custom_fqdns" to be able to add custom FQDNs to Context Profiles.

### Describe alternatives you've considered

Right now I'm using a "null_resource" with a "local-exec" provisioner that launches a "curl" request to the REST API:

```
locals {
  allowList = setsubtract(split("\n", file("./transparentProxy_allowList.txt")), split("\n", file("./transparentProxy_fqdnsPredefined.txt")))
}
resource "null_resource" "transparentProxy_updateFQDNs" {
  triggers = {
    always_run = timestamp()
  }
  provisioner "local-exec" {
    interpreter = ["bash", "-c"]
    command = "curl -k --user ${var.nsx_username}:${var.nsx_password} -H \"Content-Type: application/json\" --request PATCH https://mynsxt.host.org/policy/api/v1/infra/context-profiles/custom-attributes -d ' { \"key\" : \"DOMAIN_NAME\", \"value\" : ${jsonencode(local.allowList)}, \"datatype\" : \"STRING\", \"attribute_source\" : \"CUSTOM\" }'"
  }
}
```

PR #835, opened 2023-01-31
 Add FQDN or URL Custom support for context profile

ksamoray:
To allow addition of custom URLs or FQDNs to context profile objects.

Fixes: #699, #721

Signed-off-by: Kobi Samoray <ksamoray@vmware.com>

### DIFF
--- a/nsxt/resource_nsxt_policy_context_profile_custom_attribute.go
+++ b/nsxt/resource_nsxt_policy_context_profile_custom_attribute.go
@@ -1,0 +1,207 @@
+/* Copyright © 2023 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/vmware/vsphere-automation-sdk-go/lib/vapi/std/errors"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
+	gm_infra "github.com/vmware/vsphere-automation-sdk-go/services/nsxt-gm/global_infra/context_profiles/custom_attributes"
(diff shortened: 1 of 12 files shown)


nsxt/resource_nsxt_policy_context_profile_custom_fqdn.go
```diff
@@ -0,0 +1,182 @@
...
+	return &schema.Resource{
+		Create: resourceNsxtPolicyContextProfileCustomFQDNCreate,
+		Read:   resourceNsxtPolicyContextProfileCustomFQDNRead,
+		Delete: resourceNsxtPolicyContextProfileCustomFQDNDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"fqdn": {
```
annakhm, line 28: Should we allow to configure CUSTOM_URL as part of same resource?

ksamoray: Yeah we could rename this resource to "custom_attribute" and have both. Would it make this more usable?

nsxt/resource_nsxt_policy_context_profile_custom_attribute.go
```diff
@@ -0,0 +1,209 @@
...
+	d.Set("attribute", attribute)
+	return nil
+}
+
+func resourceNsxtPolicyContextProfileCustomAttributeCreate(d *schema.ResourceData, m interface{}) error {
+	var err error
+	key := d.Get("key").(string)
+	attribute := d.Get("attribute").(string)
+	if d.Id() == "" {
+		d.SetId(makeCustomAttributeID(key, attribute))
```
annakhm, line 135: I believe we need to SetId at the end of Create, once we're the resource was created succesfully

ksamoray: OK

annakhm (approved):
Lets also add a test that uses custom attribute in context profile. Could be in follow up if you prefer.

ksamoray:
> Lets also add a test that uses custom attribute in context profile. Could be in follow up if you prefer.

Makes sense, I'll create an issue to track this follow up.

PR #837, opened 2023-02-15
Fix VPN data sources for newer NSX version

annakhm:
With latest NSX, search API result omits locale-service component, which data sources use in filtering.
This breaks vpn local endpoint data source when used with service_path. This change switches from search API to List API for such cases.
In addition, expose local_address in local enpoint data source.


ksamoray (approved):
LGTM

PR #838, opened 2023-02-22
Upgrade dependencies to fix security issues

annakhm:
In addition, upgrade golang to 1.19 in all workflows


.github/workflows/golangci-lint.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     steps:
       - uses: actions/setup-go@v3
         with:
-          go-version: 1.17
+          go-version: 1.18
```
ksamoray, line 11: Why 1.18 here and 1.19 everywhere else?
Otherwise LGTM

PR #843, opened 2023-02-27
Fix mixup in condition to address lb rules

annakhm:
A negation was missing in assesment of rule change, that relates to provider version upgrade workaround. When no change in rules is detected, we assume current rules should be preserved, for sake of rules that were added manually on NSX when rule feature was not present in proveder.

In addition, fix inconsistent documentation.


ksamoray (approved):
LGTM

Issue #841, opened 2023-02-24
nsxt_policy_security_policy rules don't work on Ethernet categories


dmbuil:
### Describe the bug

Terraform v1.3.6
on linux_amd64
+ provider registry.terraform.io/vmware/nsxt v3.2.9

When defining a rule on a distributed firewall section (nsxt_policy_security_policy), it cannot set rules, as the rule-type conflicts with the default values of the NSX-T provider of the rules' param `ip_version`.

Looks like Ethernet rules are not supported.

### Reproduction steps

On this example, no `ip_version` is set, but TF will populate its value to the default `IPv4_IPv6`:

```terraform
resource "nsxt_policy_security_policy" "my_section" {
  display_name = "my_section"

  category    = "Ethernet"
  locked      = false
  stateful    = false
  tcp_strict  = false

  rule {
    disabled      = false
    display_name  = "Allow ARP Broadcast"
    description   = "My cool description"
    source_groups = []
    destination_groups = [    ]
    services = []
    action   = "ALLOW"
    scope = [
      data.nsxt_policy_group.my_network.path
    ]
    logged     = false
    log_label  = ""
  }
```

As `IPv4_IPv6`
It fails with the following:

```
IP Protocol cannot be used in Layer2 (Ethernet) rules. (code 500201) IP Protocol cannot be used in Layer2 (Ethernet) rules. (code 500201) IP Protocol cannot be used in Layer2 (Ethernet) rules. (code 500201) IP Protocol cannot be used in Layer2 (Ethernet) rules. (code 500201) 
```

If the rule is added manually in both the GUI and TF, with the same params, you see that `ip_version` value is overwritten by TF, causing the issue:

```terraform
  rule {
    action             = "ALLOW"
    destination_groups = []
    direction          = "IN_OUT"
    display_name       = "New Rule"
    logged             = false
    scope              = []
    services           = []
    source_groups      = []
    ip_version         = null
  }
```

![2023-02-24_12h59_43](https://user-images.githubusercontent.com/22377986/221174368-bd900504-57f1-4301-9108-93738da3e479.png)

### Expected behavior

Ethernet type rules must be supported, by accepting "null" as a value for rules' `ip_version`.

PR #844, opened 2023-02-28
Allow creation of Ethernet category rules

ksamoray:
Creation of Ethernet rules fails when ip_protocol is not null

Fixes: #841


nsxt/policy_common.go
```diff
@@ -201,7 +201,6 @@ func getSecurityPolicyAndGatewayRulesSchema(scopeRequired bool, isIds bool, nsxI
 			Description:  "IP version",
 			Optional:     true,
 			ValidateFunc: validation.StringInSlice(securityPolicyIPProtocolValues, false),
-			Default:      model.Rule_IP_PROTOCOL_IPV4_IPV6,
```
annakhm, line 204: I believe this should be set to Computed, since it would be initialized by platform now.

annakhm (approved):

PR #846, opened 2023-03-01
Support session auth for policy resources

annakhm:
In addition, support ability to disable session auth for both MP and policy resources on provider level.
This change does not cover session reestablishment when xsrf token expires. This will be addressed in a follow-up.


ksamoray (approved):
LGTM

Issue #852, opened 2023-03-15
Updates on "nsxt_policy_tier1_gateway" ressource results in "Not Set" ha mode for tier-1 in nsx-t manager


m-garrido:
### Describe the bug

When we update a tier-1 gateway ressource by terraform, in nsx-t manager webui, the "ha mode" is set to "Not Set".
However, the tier-1 hasn't change is mode as is always in "Active-Standby" mode. In order to recover the value, edit/save without changes.

### Reproduction steps

Create a change in a "nsxt_policy_tier1_gateway" ressource and apply with terraform.
See in the webui nsx-t manager tier-1 the "ha mode" state.

### Expected behavior

In the webui nsx-t manager, the tier-1 must show the expected value for "ha mode".

### Additional context

The issue could be from nsx-t and not from terraform nsx-t provider.

RawKitn:
Hi Mika, thx for your help, we had drift yesterday and could not find out what causing this behaviour <3

PR #856, opened 2023-03-21
Add ha_mode attribute to T1 gateway

ksamoray:
Fixes: #852

### DIFF
--- a/nsxt/resource_nsxt_policy_tier1_gateway.go
+++ b/nsxt/resource_nsxt_policy_tier1_gateway.go
@@ -43,6 +43,10 @@ var poolAllocationValues = []string{
 	model.Tier1_POOL_ALLOCATION_LB_XLARGE,
 }
 
+var t1HaModeValues = []string{
+	model.Tier1_HA_MODE_ACTIVE,
+	model.Tier1_HA_MODE_STANDBY}
+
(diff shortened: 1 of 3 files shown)


ksamoray:
model.Tier1_HA_MODE_ACTIVE constant requires vSphere SDK upgrade

nsxt/resource_nsxt_policy_tier1_gateway.go
```diff
@@ -397,6 +409,7 @@ func policyTier1GatewayResourceToInfraStruct(d *schema.ResourceData, connector *
 		RouteAdvertisementRules: routeAdvertisementRules,
 		Ipv6ProfilePaths:        ipv6ProfilePaths,
 		ResourceType:            &t1Type,
+		HaMode:                  &haMode,
```
annakhm, line 412: I think this should require NSX version check (specifying this value in early versions would likely lead to an error)

annakhm: For NSX < 4.0.0 this would cause ```Error:  Failed to create Tier1: Json de-serialization error: property ha_mode is unrecognized. (code 287)```

annakhm (approved):

Issue #855, opened 2023-03-20
Missing group_type definition for nsxt_policy_group


laurentmas:
### Describe the bug

In VMC on AWS 1.20v2 with nsx 4.0.1.52.0.20682514 , a new argument has been added when we create a Management Gateway policy group.

Now by default groups are created with IP Addresses Only Type instead of Generic.

On terraform provider, group_type parameter is missing so we cannot specify the type when using it.
If we want to update a Management Group that has been created manually or by VMC, we cannot until we remove them

group_type:  (Array<string>)
Group type can be specified during create and update of a group.
Empty group type indicates a 'generic' group, ie group can
include any entity from the valid GroupMemberType.

### Reproduction steps

1. Manually creates a Group Test on management Gateway on VMC on AWS 1.20
2. Add an ipaddress into the group and save it.
3. Try to Update the Group using terraform by adding a new ip address; 
4. Terraform Apply

Terraform with fails with the following error :
│ Error:  Failed to update Group Test: Invalid Group Member type update. [IPAddress] type cannot be updated to Generic type. (code 500336)
...

### Expected behavior

We should be allowed to specify group_type as IPAddress and other supported group types.

If no group_type is specified, it must creates a generic group as the current behavior.

annakhm:
Hi @carsonpadawon, we are aware of this issue (its VMC-specific) and working on the fix. Thanks for reporting!

PR #857, opened 2023-03-21
Add group_type attribue to policy group

ksamoray:
Fixes: #855

### DIFF
--- a/nsxt/resource_nsxt_policy_group.go
+++ b/nsxt/resource_nsxt_policy_group.go
@@ -51,6 +51,11 @@ var externalMemberTypeValues = []string{
 	model.ExternalIDExpression_MEMBER_TYPE_PHYSICALSERVER,
 }
 
+var groupTypeValues = []string{
+	model.Group_GROUP_TYPE_IPADDRESS,
+	model.Group_GROUP_TYPE_ANTREA,
+}
+
 func resourceNsxtPolicyGroup() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceNsxtPolicyGroupCreate,
@@ -69,6 +74,12 @@ func resourceNsxtPolicyGroup() *schema.Resource {
 			"revision":     getRevisionSchema(),
 			"tag":          getTagsSchema(),
 			"domain":       getDomainNameSchema(),
+			"group_type": {
+				Type:         schema.TypeString,
+				Description:  "Indicates the group type",
+				ValidateFunc: validation.StringInSlice(groupTypeValues, false),
+				Optional:     true,
+			},
 			"criteria": {
 				Type:        schema.TypeList,
 				Description: "Criteria to determine Group membership",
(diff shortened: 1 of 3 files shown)


ksamoray:
@annakhm note that in the API, group_type is a list with a maximum of one member.
I implemented this as a string property as I thought that it'll make more sense. If you believe that implementing this as a list also in the provider, I can change that.

website/docs/r/policy_group.html.markdown
```diff
@@ -184,6 +184,7 @@ The following arguments are supported:
     * `distinguished_name` (Required for an `identity_group`) LDAP distinguished name (DN). A valid fully qualified distinguished name should be provided here. This value is valid only if it matches to exactly 1 LDAP object on the LDAP server.
     * `domain_base_distinguished_name` (Required for an `identity_group`) Identity (Directory) domain base distinguished name. This is the base distinguished name for the domain where this identity group resides. (e.g. dc=example,dc=com)
     * `sid` (Optional) Identity (Directory) Group SID (security identifier). A security identifier (SID) is a unique value of variable length used to identify a trustee. This field is only populated for Microsoft Active Directory identity store.
+* `group_type` - (Optional) Group type can be specified during create and update of a group. Empty group type indicates a 'generic' group, ie group can be one of IPAddress, ANTERA.
```
annakhm, line 187: ANTERA -> ANTREA?

ksamoray: Indeed. Thanks

nsxt/resource_nsxt_policy_group.go
```diff
@@ -844,12 +855,18 @@ func resourceNsxtPolicyGroupCreate(d *schema.ResourceData, m interface{}) error
...
+	if groupType != "" {
+		groupTypes = append(groupTypes, groupType)
+	}
 	obj := model.Group{
 		DisplayName:        &displayName,
 		Description:        &description,
 		Tags:               tags,
 		Expression:         expressionData,
 		ExtendedExpression: extendedExpressionList,
+		GroupType:          groupTypes,
```
annakhm, line 869: I believe version check is needed here since GroupType support was added recently in NSX

nsxt/resource_nsxt_policy_group.go
```diff
@@ -844,12 +855,18 @@ func resourceNsxtPolicyGroupCreate(d *schema.ResourceData, m interface{}) error
 	description := d.Get("description").(string)
 	tags := getPolicyTagsFromSchema(d)
 
+	var groupTypes []string
+	groupType := d.Get("group_type").(string)
+	if groupType != "" && nsxVersionHigherOrEqual("3.2.0") {
```
annakhm, line 860: I don't think this version check would work with NSX version <3.2, since empty list of group_types would still be sent to the platform, and NSX would not recognize this property. The version check needs to define whether `GroupType` is set in the struct.

nsxt/resource_nsxt_policy_group.go
```diff
@@ -969,12 +992,20 @@ func resourceNsxtPolicyGroupUpdate(d *schema.ResourceData, m interface{}) error
...
+	var groupTypes []string
+	groupType := d.Get("group_type").(string)
+	groupTypes = append(groupTypes, groupType)
 	obj := model.Group{
 		DisplayName:        &displayName,
 		Description:        &description,
 		Tags:               tags,
 		Expression:         expressionData,
 		ExtendedExpression: extendedExpressionList,
+		GroupType:          groupTypes,
```
annakhm, line 1004: In `Update` code path Group type is still set regardless of version

nsxt/resource_nsxt_policy_group_test.go
```diff
@@ -83,6 +83,52 @@ func TestAccResourceNsxtPolicyGroup_AddressCriteria(t *testing.T) {
 	})
 }
 
+func TestAccResourceNsxtPolicyGroup_GroupTypeIPAddressCriteria(t *testing.T) {
+	name := getAccTestResourceName()
+	testResourceName := "nsxt_policy_group.test"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
```
annakhm, line 91: Can we add a version check here?

PR #878, opened 2023-04-25
Segment security profile resource

annakhm:


nsxt/resource_nsxt_policy_segment_security_profile.go
```diff
@@ -0,0 +1,317 @@
...
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+			"dhcp_server_block_v6_enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+			"non_ip_traffic_block_enabled": {
```
ksamoray, line 69: Aren't we usually adding description to resource attributes?

annakhm: This file was generated by policygen, and descriptions in spec are often too long to appear here in the code. We do have them in documentation. I'm not sure description in the code has any practical purpose.

ksamoray: Makes sense. I wonder how much the provider image will shrink if we remove all these strings :)

ksamoray (approved):
LGTM

PR #884, opened 2023-05-03
Support enum delta in diffy tool

annakhm:
Diffy tool will now print out diff in enumeration values between two NSX specs


ksamoray (approved):
LGTM, IDK if platform would ever deprecate values from enums. If this is possible maybe it's worthwhile also to test baseline != target

annakhm:
I didn't see a way to deprecate in spec - enums are defined in a list. I think in order to remove items from the list you'll need to deprecate the attribute and redefine it with different enum. But not 100% sure here :)

PR #885, opened 2023-05-04
Multitenancy project data source

ksamoray:
Add a data source for multitenancy project object.

### DIFF
--- a/nsxt/data_source_nsxt_policy_project.go
+++ b/nsxt/data_source_nsxt_policy_project.go
@@ -1,0 +1,128 @@
+/* Copyright © 2023 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+	infra "github.com/vmware/vsphere-automation-sdk-go/services/nsxt/orgs"
+)
+
+const defaultOrgID = "default"
+
+func dataSourceNsxtPolicyProject() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNsxtPolicyProjectRead,
+
+		Schema: map[string]*schema.Schema{
+			"id":           getDataSourceIDSchema(),
+			"display_name": getDataSourceDisplayNameSchema(),
+			"description":  getDataSourceDescriptionSchema(),
+			"path":         getPathSchema(),
+			"short_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"site_info": {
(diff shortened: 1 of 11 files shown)


nsxt/data_source_nsxt_policy_project.go
```diff
@@ -0,0 +1,83 @@
...
+			"path":         getPathSchema(),
+		},
+	}
+}
+
+func dataSourceNsxtPolicyProjectRead(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+	client := infra.NewProjectsClient(connector)
+
+	objID := d.Get("id").(string)
```
annakhm, line 34: Hi Kobi, why not use search API here, like in https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/data_source_nsxt_policy_segment.go for example ? Are projects searchable?

nsxt/data_source_nsxt_policy_project.go
```diff
@@ -0,0 +1,83 @@
...
+
+func dataSourceNsxtPolicyProject() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNsxtPolicyProjectRead,
+
+		Schema: map[string]*schema.Schema{
+			"id":           getDataSourceIDSchema(),
+			"display_name": getDataSourceDisplayNameSchema(),
+			"description":  getDataSourceDescriptionSchema(),
+			"path":         getPathSchema(),
```
annakhm, line 25: We also need to expose associated edge clusters and T0, so that user could reference those in subsequent config

nsxt/data_source_nsxt_policy_project.go
ksamoray: Not sure why but search doesn't seem to work for me

ksamoray: I've added a comment - although it's doable with search API, it'll be quite cumbersome even after I'll add support for [multitenancy](https://github.com/vmware/terraform-provider-nsxt/compare/master...ksamoray:terraform-provider-nsxt:api-wrapper#diff-c5d8c0bd4f04ea7a2984c77a5b9c331564d1b48b518cb6a74fd54e897671a900R230) in  search API

annakhm (approved):

nsxt/data_source_nsxt_policy_project_test.go
```diff
@@ -0,0 +1,105 @@
...
+	connector, err := testAccGetPolicyConnector()
+	if err != nil {
+		return fmt.Errorf("Error during test client initialization: %v", err)
+	}
+	client := orgs.NewProjectsClient(connector)
+
+	// Find the object by name
+	objList, err := client.List(defaultOrgID, nil, nil, nil, nil, nil, nil, nil)
+	if err != nil {
+		return handleListError("IpiAddressBlock", err)
```
annakhm, line 86: Project?

PR #889, opened 2023-05-10
Add cache_size property to dns forwarder

annakhm:


ksamoray (approved):
Is there a value range for this attr?

annakhm:
> Is there a value range for this attr?

Yes, however since range can change in later versions I'd rather let platform return error (until we can update these range values with generator later on)

PR #890, opened 2023-05-14
Add project resource type

ksamoray:


nsxt/resource_nsxt_policy_project.go
```diff
@@ -0,0 +1,220 @@
...
+				Optional: true,
+				Computed: true,
+			},
+			"site_info": {
+				Type: schema.TypeList,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"edge_cluster_paths": {
+							Type: schema.TypeList,
+							Elem: &schema.Schema{
```
annakhm, line 46: we can use getElemPolicyPathSchema here?

nsxt/resource_nsxt_policy_project.go
```diff
@@ -0,0 +1,220 @@
...
+						},
+						"site_path": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+					},
+				},
+				Optional: true,
+			},
+			"tier0s": {
```
annakhm, line 59: `tier0_gateway_paths` for consistency?

nsxt/resource_nsxt_policy_project.go
ksamoray: Also for the T0 path?

nsxt/data_source_nsxt_policy_project.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+const defaultOrgID = "default"
+
+func dataSourceNsxtPolicyProject() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNsxtPolicyProjectRead,
+
+		Schema: map[string]*schema.Schema{
+			"id":           getDataSourceIDSchema(),
```
annakhm, line 22: should this be `nsx_id` (getNsxIDSchema) as in other resources?,

nsxt/resource_nsxt_policy_project.go
```diff
@@ -0,0 +1,207 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"site_info": {
+				Type: schema.TypeList,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"edge_cluster_paths": getElemPolicyPathSchemaWithFlags(true, false, false),
+						"site_path":          getElemPolicyPathSchemaWithFlags(true, false, false),
```
annakhm, line 43: Is this a list or a single element?

ksamoray: edge_cluster_paths should be a list indeed

nsxt/resource_nsxt_policy_project.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+func resourceNsxtPolicyProjectUpdate(d *schema.ResourceData, m interface{}) error {
+
+	id := d.Id()
+	if id == "" {
+		return fmt.Errorf("Error obtaining Project ID")
+	}
+
+	headers := map[string]string{"enable_partial_patch": "true"}
```
annakhm, line 185: Is this header still needed?

ksamoray: In theory we do: I've added this to handle that issue, where NSX was overriding short_id on update.
The NSX bug is fixed but there are still released NSX versions which support projects but do not have this fix.
However - tests seem to pass without it - so I'll remove it :)

nsxt/resource_nsxt_policy_project.go
```diff
@@ -0,0 +1,210 @@
...
+			edgeClusterPaths = append(edgeClusterPaths, ecp.(string))
+		}
+		sitePath := data["site_path"].(string)
+		obj := model.SiteInfo{
+			EdgeClusterPaths: edgeClusterPaths,
+			SitePath:         &sitePath,
+		}
+		siteInfos = append(siteInfos, obj)
+	}
+	var tier0s []string
```
annakhm, line 97: Lets use `getStringListFromSchemaList` helper here?

nsxt/resource_nsxt_policy_project.go
annakhm: We'll officially claim multitenancy support starting from next release, so the fix will be in

ksamoray: Cool, I'll leave `getPolicyConnectorWithHeaders()` just in case we'll even need it?

annakhm (approved):

PR #891, opened 2023-05-15
Fix issue when creating IPSecVpnSession with Compliance suite

2ez4szliu:
Change tunnel_profile_path to use the computedPolicyPath so that IPSecVpnSession with compliance suite can be configured correctly.

The MR also made enhancement to the docs.


nsxt/resource_nsxt_policy_ipsec_vpn_session_test.go
```diff
@@ -175,6 +204,47 @@ func TestAccResourceNsxtPolicyIPSecVpnSessionRouteBased_basic(t *testing.T) {
 	})
 }
```
annakhm, line 206: lets follow naming convention here - `withComplianceSuite`

nsxt/resource_nsxt_policy_ipsec_vpn_session_test.go
```diff
@@ -273,6 +343,49 @@ func TestAccResourceNsxtPolicyIPSecVpnSessionPolicyBased_basic(t *testing.T) {
...
 
+func TestAccResourceNsxtPolicyIPSecVpnSessionPolicyBased_with_compliance_suite(t *testing.T) {
```
annakhm, line 346: same as above

nsxt/resource_nsxt_policy_ipsec_vpn_session_test.go
```diff
@@ -485,6 +598,7 @@ func testAccNSXPolicyIPSecVpnSessionImporterGetID(s *terraform.State) (string, e
 }
 
 func testAccNsxtPolicyIPSecVpnSessionPreConditionTemplate(isT0 bool) string {
```
annakhm, line 600: Tests using certificate should only run if testAccEnvDefined(t, "NSXT_TEST_CERTIFICATE_NAME"), otherwise looks like certificate data source will fail

nsxt/resource_nsxt_policy_ipsec_vpn_session_test.go
2ez4szliu: done

nsxt/resource_nsxt_policy_ipsec_vpn_session_test.go
2ez4szliu: done

nsxt/resource_nsxt_policy_ipsec_vpn_session_test.go
2ez4szliu: done

PR #895, opened 2023-05-17
Implement multi-tenancy support for relevant resources

ksamoray:
Some of the resource have been added support to multi-tenancy.
Support is implemented using a generated API wrapper, which abstracts the calls to the various NSX APIs (local manager/global manager/multi tenancy).


nsxt/utils.go
```diff
@@ -662,3 +662,25 @@ func handlePagination(lister func(*paginationInfo) error) (int64, error) {
 
 	return total, nil
 }
+
+func getContextSchema(forceNew bool) *schema.Schema {
```
annakhm, line 666: When do we expect forceNew to be false?
Since context triggers object path change, seems to me ForceNew should always be defined

ksamoray: Indeed. Thanks :)

ksamoray: I'm guessing that the whole context - no matter what we'll add to it in the future, should be ForceNew as well.

api/infra.go
```diff
@@ -0,0 +1,78 @@
...
+		client = client1.NewInfraClient(connector)
+
+	default:
+		return nil
+	}
+	return &InfraClientContext{Client: client, ClientType: sessionContext.ClientType, ProjectID: sessionContext.ProjectID}
+}
+
+func (c InfraClientContext) Get(basePathParam *string, filterParam *string, typeFilterParam *string) (model0.Infra, error) {
+	var emptyOutput model0.Infra
```
annakhm, line 35: Do we need the distinction between `obj` and `emptyOutput`?

api/infra.go
```diff
@@ -0,0 +1,78 @@
+//nolint:revive
+package nsxt
```
annakhm, line 2: I would suggest adding a header that indicates this file is autogenerated

nsxt/utils.go
```diff
@@ -662,3 +662,22 @@ func handlePagination(lister func(*paginationInfo) error) (int64, error) {
...
+func getContextSchema() *schema.Schema {
+	return &schema.Schema{
+		Type:        schema.TypeList,
+		Description: "Resource context",
+		Optional:    true,
+		MaxItems:    1,
+		ForceNew:    true,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"project_id": {
```
annakhm, line 675: Not sure whether we should use `project_path` for consistency. This would eliminate the need to add `org` in future, but might be cumbersome if projects are hard-coded rather than specified by dependency.

nsxt/resource_nsxt_policy_service.go
```diff
@@ -504,24 +491,11 @@ func resourceNsxtPolicyServiceRead(d *schema.ResourceData, m interface{}) error
...
-	} else {
-		client := infra.NewServicesClient(connector)
-		var err error
-		obj, err = client.Get(id)
-		if err != nil {
-			return handleReadError(d, "Service", id, err)
-		}
+	client := infra.NewServicesClient(getSessionContext(d, m), connector)
+	var err error
+	obj, err = client.Get(id)
```
annakhm, line 496: We can save two lines of code here and in similar places by not defining `obj` and `err` and using `:=` on this line

api/infra.go
ksamoray: If Get returns an obj + err, do we still want to return anything? Or have a blank value?
I'm OK with dropping emptyOutput.

annakhm: I don't think object matters if error is not nil

nsxt/policy_search.go
```diff
@@ -206,3 +226,29 @@ func searchLMPolicyResources(connector client.Connector, query string) ([]*data.
...
 	}
 }
+
+func searchMultitenancyPolicyResources(connector client.Connector, org string, project string, query string) ([]*data.StructValue, error) {
+	client := lm_search.NewQueryClient(connector)
+	var results []*data.StructValue
+	var cursor *string
+	total := 0
+
+	// Make sure global objects are not found (path needs to start with infra)
```
annakhm, line 236: this comment is not relevant here

nsxt/resource_nsxt_policy_dhcp_v4_static_binding.go
```diff
@@ -68,24 +68,19 @@ func resourceNsxtPolicyDhcpV4StaticBinding() *schema.Resource {
...
+func getPolicyDchpStaticBindingOnSegment(context utl.SessionContext, id string, segmentPath string, connector client.Connector) (*data.StructValue, error) {
 	_, gwID, segmentID := parseSegmentPolicyPath(segmentPath)
-	if isGlobalManager {
-		client := gm_segments.NewDhcpStaticBindingConfigsClient(connector)
-		return client.Get(segmentID, id)
-	}
-	if gwID == "" {
-		// infra segment
-		client := segments.NewDhcpStaticBindingConfigsClient(connector)
+	if context.ClientType == utl.Global || gwID == "" {
```
annakhm, line 73: fixed segments might be supported on GM now. I suggest not to change the logic in this PR but lets check this later (it would simplify the code to only look at GW present in this condition)

ksamoray: OK. So should we open an issue to track this?

api/api_list.yaml
```diff
@@ -0,0 +1,999 @@
...
+      type: Local
+    - client: github.com/vmware/vsphere-automation-sdk-go/services/nsxt-gm/global_infra
+      model: github.com/vmware/vsphere-automation-sdk-go/services/nsxt-gm/model
+      type: Global
+    - client: github.com/vmware/vsphere-automation-sdk-go/services/nsxt/orgs/projects/infra
+      model: github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model
+      type: Multitenancy
+  model_name: PolicyContextProfile
+  obj_name: ContextProfile
+  var_name: policyContextProfileParam
```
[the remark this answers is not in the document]
ksamoray: Schema is explained at the top of the file. Specifically, `var_name` explained [here](https://github.com/vmware/terraform-provider-nsxt/blob/1106dfcbad347d033c024ba5855595f66a779a6f/api/api_list.yaml#L18)

api/utl/api_util.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func ConvertModelBindingType(obj interface{}, sourceType bindings.BindingType, destType bindings.BindingType) (interface{}, error) {
+	converter := bindings.NewTypeConverter()
+	dataValue, err := converter.ConvertToVapi(obj, sourceType)
+	if err != nil {
+		return nil, err[0]
+	}
+
+	gmObj, err := converter.ConvertToGolang(dataValue, destType)
```
annakhm, line 33: This is not specific to gm (global manager). Perhaps `convertedObj`?

annakhm:
I wonder if enforcement points relevant for /orgs? We have functions like `getPolicyEnforcementPointPath` that hard-code `infra` and `global-infra`

annakhm:
We would also need to craft a documentation guide for multitenancy use case, but this can be added in follow up

annakhm (approved):

PR #897, opened 2023-05-19
Remove `Policy` prefix from categories in docs

annakhm:
MP resources moved into Deprecated category, hence there should be no need to repeat the `Policy` keyword in category. In addition, some documentation polishing.


ksamoray (approved):
LGTM

PR #905, opened 2023-06-07
Multitenancy resources and data sources tests

ksamoray:
Testing to multitenancy support resources.


annakhm:
LGTM, I would suggest to also add multitenancy-specific update tests, such as add/remove/update context on resource

annakhm (approved):

PR #908, opened 2023-06-14
Multitenancy import

ksamoray:
Implement import functionality for supported resources

### DIFF
--- a/nsxt/policy_utils.go
+++ b/nsxt/policy_utils.go
@@ -4,6 +4,7 @@
 package nsxt
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"strings"
@@ -12,7 +13,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"github.com/vmware/vsphere-automation-sdk-go/lib/vapi/std/errors"
+	sdkerrors "github.com/vmware/vsphere-automation-sdk-go/lib/vapi/std/errors"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/realized_state"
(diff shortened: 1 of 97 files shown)


nsxt/resource_nsxt_policy_tier1_gateway_interface.go
```diff
@@ -316,14 +317,21 @@ func resourceNsxtPolicyTier1GatewayInterfaceDelete(d *schema.ResourceData, m int
 func resourceNsxtPolicyTier1GatewayInterfaceImport(d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
 	importID := d.Id()
 	s := strings.Split(importID, "/")
```
annakhm, line 319: cosmetics: lets move this below the path import section?

ksamoray: The split? It's used at line 323 to get the locale_service_id and then again in 328. So IDK if there's a better place for it.
Unless you mean something else.

annakhm: Correct, my bad, sorry. But it seems wrong to reference s[9] below? Doesn't index depend on whether tenant is present in path? And the user can provide path with smaller amount of tokens, that would still be policy path

nsxt/policy_utils.go
```diff
@@ -180,6 +196,39 @@ func nsxtDomainResourceImporter(d *schema.ResourceData, m interface{}) ([]*schem
...
+		return rd, err
+	}
+	return rd, nil
+}
+
+func nsxtPolicyPathResourceImporterHelper(d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
+	importID := d.Id()
+	if isPolicyPath(importID) {
+		pathSegs := strings.Split(importID, "/")
+		if len(pathSegs) < 4 {
```
annakhm, line 213: This is already validated in isPolicyPath

nsxt/policy_utils.go
```diff
@@ -180,6 +196,39 @@ func nsxtDomainResourceImporter(d *schema.ResourceData, m interface{}) ([]*schem
...
+			return nil, fmt.Errorf("invalid policy path %s", importID)
+		} else if strings.Contains(pathSegs[1], "infra") {
+			d.SetId(pathSegs[len(pathSegs)-1])
+		} else if pathSegs[1] == "orgs" && pathSegs[3] == "projects" {
+			// pathSegs[2] should contain the organization. Once we support multiple organization, it should be
+			// assigned into the context as well
+			contexts := make([]interface{}, 1)
+			ctxMap := make(map[string]interface{})
+			ctxMap["project_id"] = pathSegs[4]
```
annakhm, line 222: can we get a segmentation fault here if user provides bad path?

nsxt/resource_nsxt_policy_tier1_gateway_interface.go
ksamoray: Indeed. It's better to locate "locale_service" and grab the next list member.

nsxt/policy_utils.go
ksamoray: So maybe we should check for len(pathSegs < 5) for project case

annakhm (approved):

PR #909, opened 2023-06-20
Support `type` attribute on T1 for VMC

annakhm:
This attribute helps auto-configure route advertisements. In addition, support `DR only` mode for `ha_mode` setting.


ksamoray (approved):
LGTM

PR #910, opened 2023-06-25
Multitenancy documentation

ksamoray:
Add general multitenancy guide

### DIFF
--- a/website/docs/guides/multitenancy.html.markdown
+++ b/website/docs/guides/multitenancy.html.markdown
@@ -1,0 +1,124 @@
+---
+layout: "nsxt"
+page_title: "VMware NSX-T Terraform Provider support for Multi-tenancy"
+description: |-
+  The VMware NSX-T Terraform Provider support for Multi-tenancy
+---
+
+NSX-T Terraform Provider offers support for NSX-T [multi-tenancy feature](https://docs.vmware.com/en/VMware-NSX/4.1/administration/GUID-52180BC5-A1AB-4BC2-B1CE-666292505317.html).
+
+# NSX-T Project Creation and Reference
+
+NSX-T Project objects can be created and referenced with the `nsxt_policy_project` [resource](../r/policy_project.html.markdown) and [data source](../d/policy_project.html.markdown).
+
+For example, an NSX-T multi-tenancy Project could be created as below:
+
+```hcl
+resource "nsxt_policy_project" "test" {
+  display_name        = "test"
+  description         = "Terraform provisioned Project"
+  short_id            = "test"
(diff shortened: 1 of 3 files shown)


website/docs/guides/multitenancy.html.markdown
```diff
@@ -0,0 +1,125 @@
...
+
+NSX-T Project objects can be created and referenced with the `nsxt_policy_project` [resource](../r/policy_project.html.markdown) and [data source](../d/policy_project.html.markdown).
+
+For example, an NSX-T multi-tenancy Project could be created as below:
+
+```hcl
+resource "nsxt_policy_project" "test" {
+  display_name        = "test"
+  description         = "Terraform provisioned Project"
+  _default            = true
```
annakhm, line 20: not sure what `default` is here?

website/docs/r/policy_project.html.markdown
```diff
@@ -17,7 +17,6 @@ This resource is applicable to NSX Global Manager, NSX Policy Manager and VMC.
 resource "nsxt_policy_project" "test" {
```
annakhm, line 17: The line above - project should not be applicable for GM and VMC as of now

website/docs/r/policy_project.html.markdown
```diff
@@ -47,6 +45,7 @@ In addition to arguments listed above, the following attributes are exported:
 * `id` - ID of the resource.
 * `revision` - Indicates current revision number of the object as seen by NSX-T API server. This attribute can be useful for debugging.
 * `path` - The NSX path of the policy resource.
+* `_default` - The server will populate this field when returing the resource. Ignored on PUT and POST.true - the project is a default project. Default projects are non-editable, system create ones.
```
annakhm, line 48: I don't think it is present in the resource at all? I'm not sure its useful, but if you'd like to add it, I think it should be named `default` or `is_default` to follow convention (we have `is_default` in tz and security policy data sources)

ksamoray: I think that it's useless. Maybe it's better if I'll remove it...

ksamoray: But maybe in a separate PR as this one is regarding docs - I'll get rid of this in the resource as well.

annakhm: I don't see it in resource schema: https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/resource_nsxt_policy_project.go#L27, am I missing something?

ksamoray: Cool so I removed it already :)
I'll just fix the doc then. Thanks!!

annakhm (approved):

PR #911, opened 2023-06-27
Use MP API /node/version to retrieve node version

ksamoray:
As /node API doesn't support project users


salv-orlando (approved):
LGTM (reviewed provider.go and utils.go)

PR #917, opened 2023-07-09
Error use of multitenancy with T0 gateways

ksamoray:
T0 gateways do not have multitenancy support


annakhm:
LGTM but I would suggest to move the message string out to avoid duplication, and change `multitenancy` to `context` to be more precise

ksamoray:
> LGTM but I would suggest to move the message string out to avoid duplication, and change `multitenancy` to `context` to be more precise

Makes sense, I'll revise it.

annakhm (approved):

PR #919, opened 2023-07-11
Implement edge cluster resource

ksamoray:


ksamoray:
PR still misses documentation for the resource

nsxt/resource_nsxt_edge_cluster.go
```diff
@@ -0,0 +1,300 @@
...
+		Create: resourceNsxtEdgeClusterCreate,
+		Read:   resourceNsxtEdgeClusterRead,
+		Update: resourceNsxtEdgeClusterUpdate,
+		Delete: resourceNsxtEdgeClusterDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
+			"revision": getRevisionSchema(),
+			"description": {
```
annakhm, line 38: This is not a policy resource, but I think standard helpers are still applicable here `getDescriptionSchema`

nsxt/resource_nsxt_edge_cluster.go
```diff
@@ -0,0 +1,300 @@
...
+				Type:        schema.TypeString,
+				Description: "Description of this resource",
+				Optional:    true,
+			},
+			"display_name": {
+				Type:        schema.TypeString,
+				Description: "The display name of this resource. Defaults to ID if not set",
+				Optional:    true,
```
annakhm, line 46: Same as above - I think we should follow the convention and make this Required, and use `getDisplayNameSchema` here

nsxt/resource_nsxt_edge_cluster.go
```diff
@@ -0,0 +1,300 @@
...
+				Computed:    true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+			"deployment_type": {
+				Type:        schema.TypeString,
+				Description: "Edge cluster deployment type",
+				Computed:    true,
```
annakhm, line 62: Do those flags have a default?

nsxt/resource_nsxt_edge_cluster.go
```diff
@@ -0,0 +1,300 @@
...
+			},
+			"enable_inter_site_forwarding": {
+				Type:        schema.TypeBool,
+				Description: "Flag to enable inter site forwarding",
+				Optional:    true,
+				Computed:    true,
+			},
+			"member_node_type": {
```
annakhm, line 70: This is not an enum?

nsxt/resource_nsxt_edge_cluster.go
```diff
@@ -0,0 +1,300 @@
...
+				Type:        schema.TypeString,
+				Description: "Node type of the cluster members",
+				Computed:    true,
+			},
+			"members": {
```
annakhm, line 75: per convention for nested clauses, this should be named `member`

nsxt/resource_nsxt_edge_cluster.go
ksamoray: As a convention, do we apply the NSX defaults in the provider? Or let NSX apply them as we pass null?

ksamoray: BTW deployment_type is readonly: should we even include this in the schema? This is in general, not specific to deployment_type

nsxt/resource_nsxt_edge_cluster.go
ksamoray: Same applies here: this member_node_type is readonly

nsxt/resource_nsxt_edge_cluster.go
annakhm: We generally try to avoid `Computed` attributes wherever possible, so yes, we assign default on the provider side

annakhm: If we think the attribute can be useful to the user, we can add it as `Computed`

nsxt/resource_nsxt_edge_cluster.go
```diff
@@ -0,0 +1,291 @@
...
+	d.Set("deployment_type", obj.DeploymentType)
+	d.Set("enable_inter_site_forwarding", obj.EnableInterSiteForwarding)
+	d.Set("member_node_type", obj.MemberNodeType)
+	setMemberListInSchema(d, obj.Members)
+	setNodeRtepIPsInSchema(d, obj.NodeRtepIps)
+	return nil
+}
+
+func setNodeRtepIPsInSchema(d *schema.ResourceData, nodeRtepIPs []model.NodeRtepIpsConfig) error {
+	var expresionList []map[string]interface{}
```
annakhm, line 222: typo: `expression`

nsxt/resource_nsxt_edge_cluster.go
ksamoray: As I don't see any special use for this, I'll remove this attribute entirely

nsxt/resource_nsxt_edge_cluster.go
```diff
@@ -0,0 +1,291 @@
...
+	client := nsx.NewEdgeClustersClient(connector)
+
+	description := d.Get("description").(string)
+	displayName := d.Get("display_name").(string)
+	tags := getMPTagsFromSchema(d)
+
+	clusterProfileBindingList := getStringListFromSchemaList(d, "cluster_profile_ids")
+	var clusterProfileBindings []model.ClusterProfileTypeIdEntry
+	for i := range clusterProfileBindingList {
+		elem := model.ClusterProfileTypeIdEntry{
```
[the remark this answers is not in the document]
ksamoray: Done - I think that tests should include these but we'll need at least a data source for the profiles. Right?

nsxt/resource_nsxt_edge_cluster.go
```diff
@@ -0,0 +1,309 @@
...
+		Delete: resourceNsxtEdgeClusterDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
+			"revision":     getRevisionSchema(),
+			"description":  getDescriptionSchema(),
+			"display_name": getDisplayNameSchema(),
+			"tag":          getTagsSchema(),
+			"edge_ha_profile_id": {
```
annakhm, line 29: Is it possible to add multiple ha profiles? Same about below

nsxt/resource_nsxt_edge_cluster.go
```diff
@@ -0,0 +1,309 @@
...
+						},
+						"transport_node_id": {
+							Type:        schema.TypeString,
+							Description: "UUID of edge transport node",
+							Required:    true,
+						},
+					},
+				},
+			},
+			"node_rtep_ips": {
```
annakhm, line 82: According to the spec, this whole section should be Computed, is that correct?

annakhm:
Is `allocation_rule` section relevant? I don't see it in UI, but its in the spec

nsxt/resource_nsxt_edge_cluster.go
ksamoray: This is odd - the API has a list there and there are two types (bridge, edge). Yet I can't create a bridge profile using curl (server returns 201 but nothing is created), and UI lets me select only one profile.
Not sure what we do with this (nor where we gather more info).

nsxt/resource_nsxt_edge_cluster.go
ksamoray: Indeed. I'm not sure whether we need it or not too. What do you think?

ksamoray:
> Is `allocation_rule` section relevant? I don't see it in UI, but its in the spec
I couldn't figure out how to handle this either.

nsxt/resource_nsxt_edge_cluster.go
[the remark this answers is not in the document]
ksamoray: Definitely - isn't it on the list of resources which we should implement?

nsxt/resource_nsxt_edge_cluster.go
annakhm: I think those IPs could be used elsewhere in config, so lets leave them

nsxt/resource_nsxt_edge_cluster.go
```diff
@@ -0,0 +1,309 @@
...
+			"edge_ha_profile_id": {
+				Type:        schema.TypeList,
+				Description: "Edge high availability cluster profile Ids",
+				Optional:    true,
+				Computed:    true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+			"bridge_ha_profile_id": {
```
annakhm, line 38: Yes, bridge HA cluster profile is deprecated. Lets remove this

nsxt/resource_nsxt_edge_cluster.go
annakhm: Looks like only single HA profile can be assigned - hence TypeList is not needed

annakhm (approved):
Thanks Kobi!

PR #920, opened 2023-07-12
Implement compute manager resource

ksamoray:


nsxt/resource_nsxt_compute_manager.go
```diff
@@ -0,0 +1,550 @@
...
+							MaxItems:    1,
+							ExactlyOneOf: []string{
+								"credential.0.saml_login_credential",
+								"credential.0.session_login_credential",
+								"credential.0.username_password_login_credential",
+								"credential.0.verifiable_asymmetric_login_credential",
+							},
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"thumbprint": {
```
annakhm, line 69: Lets add `Sensitive:   true` to all secrets

nsxt/resource_nsxt_compute_manager.go
```diff
@@ -0,0 +1,550 @@
...
+						},
+					},
+				},
+			},
+			"origin_type": {
+				Type:        schema.TypeString,
+				Description: "Compute manager type like vCenter",
+				Required:    true,
+			},
+			"reverse_proxy_https_port": {
```
annakhm, line 204: should we add `ValidateFunc: validateSinglePort()` for this port?

nsxt/resource_nsxt_compute_manager.go
```diff
@@ -0,0 +1,550 @@
...
+				Type:        schema.TypeInt,
+				Description: "Proxy https port of compute manager",
+				Optional:    true,
+			},
+			"server": {
+				Type:        schema.TypeString,
+				Description: "IP address or hostname of compute manager",
+				Required:    true,
+			},
+			"set_as_oidc_provider": {
```
annakhm, line 214: Default value here?

nsxt/resource_nsxt_compute_manager_test.go
```diff
@@ -0,0 +1,159 @@
...
+	return fmt.Sprintf(`
+resource "nsxt_compute_manager" "test" {
+    description  = "Terraform test compute manager"
+    display_name = "%s"
+    tag {
+    	scope = "scope1"
+        tag   = "tag1"
+    }
+
+	server = "%s"
```
annakhm, line 148: lets format this

nsxt/resource_nsxt_compute_manager.go
```diff
@@ -0,0 +1,550 @@
...
+						},
+						"value": {
+							Type:        schema.TypeString,
+							Description: "Value",
+							Required:    true,
+						},
+					},
+				},
+			},
+			"origin_type": {
```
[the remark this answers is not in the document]
ksamoray: No - and indeed it's quite awkward

nsxt/resource_nsxt_compute_manager.go
ksamoray: I've added defaults to all the attributes which have default values in NSX. But I'm not sure I understand this: if TF assigns a default value, wouldn't it clutter the state file with stuff that wasn't assigned by the user?

annakhm: If NSX assigns default, it will make its way to the state anyway.

nsxt/resource_nsxt_compute_manager.go
```diff
@@ -0,0 +1,531 @@
...
+				Default:     false,
+			},
+			"credential": {
+				Type:        schema.TypeList,
+				Description: "Login credentials for the compute manager",
+				Required:    true,
+				MaxItems:    1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"saml_login_credential": {
```
annakhm, line 58: I would suggest to remove `credential` suffix here since its already present in parent name

nsxt/resource_nsxt_compute_manager.go
```diff
@@ -0,0 +1,531 @@
...
+							Type:        schema.TypeList,
+							Description: "A login credential specifying saml token",
+							Optional:    true,
+							MaxItems:    1,
+							ExactlyOneOf: []string{
```
annakhm, line 63: shouldn't this be defined under `credential` rather than under `saml_login_credential`?

nsxt/resource_nsxt_compute_manager.go
```diff
@@ -0,0 +1,531 @@
...
+						"username_password_login_credential": {
+							Type:        schema.TypeList,
+							Description: "A login credential specifying a username and password",
+							Optional:    true,
+							MaxItems:    1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"password": {
+										Type:        schema.TypeString,
+										Description: "The authentication password for login",
```
annakhm, line 115: I think `username` and `password` are required

nsxt/resource_nsxt_compute_manager.go
```diff
@@ -0,0 +1,531 @@
...
+								"credential.0.username_password_login_credential",
+								"credential.0.verifiable_asymmetric_login_credential",
+							},
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"thumbprint": {
+										Type:        schema.TypeString,
+										Description: "Thumbprint of the server",
+										Optional:    true,
+									},
```
annakhm, line 75: feels like this should be required

nsxt/resource_nsxt_compute_manager.go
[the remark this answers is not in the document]
annakhm: I wonder what else can it be. Perhaps we should default to `vcenter` as UI does

nsxt/resource_nsxt_compute_manager.go
ksamoray: I've looked at examples such as [this](https://stackoverflow.com/questions/68550233/how-can-i-fix-exactlyoneof-foo-references-unknown-attribute-bar-at-part-bar), it seemed correct as it is. It behaves as expected.

nsxt/resource_nsxt_compute_manager.go
ksamoray: For some reason they aren't [listed](https://vdc-download.vmware.com/vmwb-repository/dcr-public/612caf56-5bb1-4a0b-8d2b-12ccd8981904/37be75ba-aa56-42f4-99c3-f6c989ddbe57/api_includes/types_UsernamePasswordLoginCredential.html) as required.
Yet the UI forces to specify these.

annakhm: Can we check this against NSX to see if empty creds are rejected? Same regarding other types of creds.

website/docs/r/compute_manager.html.markdown
```diff
@@ -0,0 +1,88 @@
...
+  display_name = "test"
+  tag {
+    scope = "scope1"
+    tag   = "tag1"
+  }
+
+  server = "192.168.244.144"
+
+  credential {
+    username_password_login_credential {
```
annakhm, line 27: `credential` no longer needed here

website/docs/r/compute_manager.html.markdown
```diff
@@ -0,0 +1,88 @@
...
+    * `asymmetric_credential` - Asymmetric login credential.
+    * `credential_key` - Credential key.
+    * `credential_verifier` - Credential verifier.
+  * `extension_certificate` - (Optional) Specifies certificate for compute manager extension.
+    * `pem_encoded` - PEM encoded certificate data.
+    * `private_key` - Private key of certificate.
+  * `multi_nsx` - (Optional) Specifies whether multi nsx feature is enabled for compute manager.
+  * `origin_type` - Compute manager type like vCenter.
+  * `reverse_proxy_https_port` - (Optional) Proxy https port of compute manager.
+  * `server` - IP address or hostname of compute manager.
```
annakhm, line 66: Should be marked as  (Required)

website/docs/r/compute_manager.html.markdown
```diff
@@ -0,0 +1,88 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `display_name` - (Required) Display name of the resource.
+* `description` - (Optional) Description of the resource.
+* `tag` - (Optional) A list of scope + tag pairs to associate with this resource.
+* `access_level_for_oidc` - (Optional) Specifies access level to NSX from the compute manager.
+* `create_service_account` - (Optional) Specifies whether service account is created or not on compute manager.
+* `credential` - Login credentials for the compute manager. Should contain exactly one credential enlisted below:
```
annakhm, line 45: This is (Required) I think?

website/docs/r/compute_manager.html.markdown
```diff
@@ -0,0 +1,88 @@
...
+    * `username` - The username for login.
+  * `verifiable_asymmetric_login` - (Optional) A verifiable asymmetric login credential.
+    * `asymmetric_credential` - Asymmetric login credential.
+    * `credential_key` - Credential key.
+    * `credential_verifier` - Credential verifier.
+  * `extension_certificate` - (Optional) Specifies certificate for compute manager extension.
+    * `pem_encoded` - PEM encoded certificate data.
+    * `private_key` - Private key of certificate.
+  * `multi_nsx` - (Optional) Specifies whether multi nsx feature is enabled for compute manager.
+  * `origin_type` - Compute manager type like vCenter.
```
annakhm, line 64: This is Optional now, lets specify default as well.

website/docs/r/compute_manager.html.markdown
```diff
@@ -0,0 +1,88 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `display_name` - (Required) Display name of the resource.
+* `description` - (Optional) Description of the resource.
+* `tag` - (Optional) A list of scope + tag pairs to associate with this resource.
+* `access_level_for_oidc` - (Optional) Specifies access level to NSX from the compute manager.
```
annakhm, line 43: Lets add the enum values and default here

nsxt/resource_nsxt_compute_manager.go
ksamoray: I think that it's a waste of time - there's no way anyone could login without creds. I made these required already.

website/docs/r/compute_manager.html.markdown
ksamoray: General question of convention: (Optional) is optional, and (Required) is required, what about those who do not have any specification of neither?

nsxt/resource_nsxt_compute_manager.go
```diff
@@ -0,0 +1,533 @@
...
+									"password": {
+										Type:        schema.TypeString,
+										Description: "The authentication password for login",
+										Required:    true,
+										Sensitive:   true,
+									},
+									"thumbprint": {
+										Type:        schema.TypeString,
+										Description: "Thumbprint of the login server",
+										Required:    true,
```
annakhm, line 122: `thumbprint` is not required in UI

ksamoray: Indeed - instead the UI fetches the `thumbprint` and asks for verification.
with `curl` the transaction fails when the `thumbprint` is missing with the following error:

`"Compute manager server x.x.x.x thumbprint input empty is not valid. Current retrieved compute manager server thumbprint is xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx . If correct, please re-submit with this thumbprint"`

nsxt/resource_nsxt_compute_manager_test.go
```diff
@@ -0,0 +1,160 @@
...
+  tag {
+    scope = "scope1"
+    tag   = "tag1"
+  }
+
+  server = "%s"
+
+  credential {
+    username_password_login {
+      username = "%s"
```
annakhm, line 152: would the API fail if credentials are incorrect? If no, I suggest to use dummy credentials here and avoid basing this test on `testAccTestVCCredentials`

annakhm (approved):
Just one comment about test precheck that can be addressed later

nsxt/resource_nsxt_compute_manager_test.go
ksamoray: NSX validates the credentials and tests the connection. And it should be a vCenter host which isn't already connected to NSX (maybe this NSX instance?) so IDK how we run this in CI. Unless we keep an extra NSX node for that purpose.

PR #926, opened 2023-07-27
Add IP allocation import support

ksamoray:


nsxt/resource_nsxt_policy_ip_address_allocation.go
```diff
@@ -227,6 +228,14 @@ func resourceNsxtPolicyIPAddressAllocationDelete(d *schema.ResourceData, m inter
 func resourceNsxtPolicyIPAddressAllocationImport(d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
 	importID := d.Id()
 	s := strings.Split(importID, "/")
+	rd, err := nsxtPolicyPathResourceImporterHelper(d, m)
+	if err == nil {
+		d.Set("pool_path", importID[0:strings.Index(importID, "/ip-allocations/")])
```
annakhm, line 233: If `ip-allocations` is not present in `importID`, we should return an error

website/docs/r/policy_ip_address_allocation.html.markdown
```diff
@@ -113,5 +113,10 @@ An existing IP Allocation can be [imported][docs-import] into this resource, via
...
 terraform import nsxt_policy_ip_address_allocation.test POOL-ID/ID
 ```
-
 The above command imports IpAddressAllocation named `test` with the NSX IpAddressAllocation ID `ID` in IP Pool `POOL-ID`.
+
+```
+terraform import nsxt_policy_ip_address_allocation.test POLICY_PATH
+```
+The above command imports IpAddressAllocation named `test` with the NSX policy path `POLICY_PATH`.
+Note: for multitenancy projects only the later form is usable.
```
annakhm, line 122: Seems like there's only one form in this import

ksamoray: Why? Non-multitenancy resources could use pool-id/allocation-id as before

nsxt/resource_nsxt_policy_ip_address_allocation.go
annakhm: If user provides invalid path that does not include `ip-allocations` provider would crash. Indeed this problem is present in other places as well.

website/docs/r/policy_ip_address_allocation.html.markdown
annakhm: Sorry my bad, you're right

nsxt/resource_nsxt_policy_ip_address_allocation.go
ksamoray: I suggest to merge this and fix all of these altogether. I'll open an issue?

ksamoray: See PR #932

annakhm (approved):

PR #928, opened 2023-07-30
Tag arguments as required or optional

ksamoray:
Some arguments within documentation were missing these.


website/docs/d/policy_context_profile.html.markdown
```diff
@@ -39,7 +39,7 @@ data "nsxt_policy_context_profile" "demoprof" {
 * `id` - (Optional) The ID of Profile to retrieve.
 * `display_name` - (Optional) The Display Name prefix of the Profile to retrieve.
 * `context` - (Optional) The context which the object belongs to
-  * `project_id` - The ID of the project which the object belongs to
+  * `project_id` - (Optional) The ID of the project which the object belongs to
```
annakhm, line 42: If `context` is provided, shouldn't `project_id` be required, at least for now?
When we add extra parameters to the context, we can switch `project_id` to optional, this would not break backwards compatibility. What do you think?

ksamoray: Makes sense. In fact, also as of now it's tagged as required also in the schema so it'll be consistent with this as well.

annakhm (approved):

PR #929, opened 2023-07-30
Implement NSXT Manager Cluster resource

2ez4szliu:
Implement NSXT Manager cluster resource.
The resource can be used to join additional nodes into the cluster of the host node in terraform provider config.
Deletion of this resource will detach these nodes from the cluster.

### DIFF
--- a/nsxt/provider.go
+++ b/nsxt/provider.go
@@ -40,6 +40,8 @@ type commonProviderConfig struct {
 	MinRetryInterval       int
 	MaxRetryInterval       int
 	RetryStatusCodes       []int
+	Username               string
+	Password               string
 }
 
 type nsxtClients struct {
@@ -400,6 +402,7 @@ func Provider() *schema.Provider {
 			"nsxt_policy_project":                          resourceNsxtPolicyProject(),
 			"nsxt_edge_cluster":                            resourceNsxtEdgeCluster(),
 			"nsxt_compute_manager":                         resourceNsxtComputeManager(),
+			"nsxt_manager_cluster":                         resourceNsxtManagerCluster(),
 		},
 
 		ConfigureFunc: providerConfigure,
@@ -625,43 +628,16 @@ func configurePolicyConnectorData(d *schema.ResourceData, clients *nsxtClients) 
 		host = fmt.Sprintf("https://%s", host)
 	}
 
-	securityCtx := core.NewSecurityContextImpl()
 	securityContextNeeded := true
 	if clientAuthDefined && !clients.CommonConfig.RemoteAuth {
 		securityContextNeeded = false
 	}
-
 	if securityContextNeeded {
-		if len(vmcAccessToken) > 0 {
-			if vmcAuthHost == "" {
-				return fmt.Errorf("vmc auth host must be provided if auth token is provided")
-			}
-
-			apiToken, err := getAPIToken(vmcAuthHost, vmcAccessToken)
-			if err != nil {
-				return err
-			}
-
-			if vmcAuthMode == "Bearer" {
-				clients.CommonConfig.BearerToken = apiToken
-			} else {
-
-				securityCtx.SetProperty(security.AUTHENTICATION_SCHEME_ID, security.OAUTH_SCHEME_ID)
-				securityCtx.SetProperty(security.ACCESS_TOKEN, apiToken)
-			}
-		} else {
-			if username == "" {
-				return fmt.Errorf("username must be provided")
-			}
-
-			if password == "" {
-				return fmt.Errorf("password must be provided")
-			}
-
-			securityCtx.SetProperty(security.AUTHENTICATION_SCHEME_ID, security.USER_PASSWORD_SCHEME_ID)
-			securityCtx.SetProperty(security.USER_KEY, username)
-			securityCtx.SetProperty(security.PASSWORD_KEY, password)
+		securityCtx, err := getConfiguredSecurityContext(clients, vmcAccessToken, vmcAuthHost, vmcAuthMode, username, password)
+		if err != nil {
+			return err
 		}
+		clients.PolicySecurityContext = securityCtx
 	}
 
 	tlsConfig, err := getConnectorTLSConfig(d)
@@ -676,9 +652,6 @@ func configurePolicyConnectorData(d *schema.ResourceData, clients *nsxtClients) 
 
 	httpClient := http.Client{Transport: tr}
 	clients.PolicyHTTPClient = &httpClient
-	if securityContextNeeded {
-		clients.PolicySecurityContext = securityCtx
-	}
 	clients.Host = host
 	clients.PolicyEnforcementPoint = policyEnforcementPoint
 	clients.PolicyGlobalManager = policyGlobalManager
@@ -689,6 +662,41 @@ func configurePolicyConnectorData(d *schema.ResourceData, clients *nsxtClients) 
 		return nil
 	}
 	return initNSXVersion(getPolicyConnector(*clients))
+}
+
+func getConfiguredSecurityContext(clients *nsxtClients, vmcAccessToken string, vmcAuthHost string, vmcAuthMode string, username string, password string) (*core.SecurityContextImpl, error) {
+	securityCtx := core.NewSecurityContextImpl()
+	if len(vmcAccessToken) > 0 {
+		if vmcAuthHost == "" {
+			return nil, fmt.Errorf("vmc auth host must be provided if auth token is provided")
+		}
+
+		apiToken, err := getAPIToken(vmcAuthHost, vmcAccessToken)
+		if err != nil {
+			return nil, err
+		}
+
+		if vmcAuthMode == "Bearer" {
+			clients.CommonConfig.BearerToken = apiToken
+		} else {
+
+			securityCtx.SetProperty(security.AUTHENTICATION_SCHEME_ID, security.OAUTH_SCHEME_ID)
+			securityCtx.SetProperty(security.ACCESS_TOKEN, apiToken)
+		}
+	} else {
+		if username == "" {
+			return nil, fmt.Errorf("username must be provided")
+		}
+
+		if password == "" {
+			return nil, fmt.Errorf("password must be provided")
+		}
+
+		securityCtx.SetProperty(security.AUTHENTICATION_SCHEME_ID, security.USER_PASSWORD_SCHEME_ID)
+		securityCtx.SetProperty(security.USER_KEY, username)
+		securityCtx.SetProperty(security.PASSWORD_KEY, password)
+	}
+	return securityCtx, nil
 }
 
 type customHeaderProcessor struct {
@@ -820,6 +828,8 @@ func initCommonConfig(d *schema.ResourceData) commonProviderConfig {
 	maxRetries := d.Get("max_retries").(int)
 	retryMinDelay := d.Get("retry_min_delay").(int)
 	retryMaxDelay := d.Get("retry_max_delay").(int)
+	username := d.Get("username").(string)
+	password := d.Get("password").(string)
 
 	statuses := d.Get("retry_on_status_codes").([]interface{})
 	retryStatuses := make([]int, 0, len(statuses))
@@ -839,6 +849,8 @@ func initCommonConfig(d *schema.ResourceData) commonProviderConfig {
 		MinRetryInterval:       retryMinDelay,
 		MaxRetryInterval:       retryMaxDelay,
 		RetryStatusCodes:       retryStatuses,
+		Username:               username,
+		Password:               password,
 	}
 }
 
--- a/nsxt/resource_nsxt_manager_cluster.go
+++ b/nsxt/resource_nsxt_manager_cluster.go
@@ -1,0 +1,441 @@
+/* Copyright © 2023 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"log"
+	"math/rand"
+	"net"
+	"strings"
+	"time"
+
+	"golang.org/x/exp/slices"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx"
+	nsxModel "github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/model"
+)
+
+func resourceNsxtManagerCluster() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNsxtManagerClusterCreate,
+		Read:   resourceNsxtManagerClusterRead,
+		Update: resourceNsxtManagerClusterUpdate,
+		Delete: resourceNsxtManagerClusterDelete,
+
+		Schema: map[string]*schema.Schema{
+			"revision": getRevisionSchema(),
+			"node": {
+				Type:        schema.TypeList,
+				Description: "Nodes in the cluster",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:        schema.TypeString,
+							Description: "ID of the cluster node",
+							Computed:    true,
+						},
+						"fqdn": {
+							Type:        schema.TypeString,
+							Description: "FQDN for the cluster node",
+							Computed:    true,
+						},
+						"status": {
+							Type:        schema.TypeString,
+							Description: "Status of the cluster node",
+							Computed:    true,
+						},
+						"ip_address": {
+							Type:         schema.TypeString,
+							Description:  "IP address of the cluster node that will join the cluster",
+							ValidateFunc: validation.IsIPAddress,
+							Required:     true,
+						},
+						"password": {
+							Type:        schema.TypeString,
+							Description: "The password for login",
+							Required:    true,
+							Sensitive:   true,
+						},
+						"username": {
+							Type:        schema.TypeString,
+							Description: "The username for login",
+							Required:    true,
+						},
+					},
+				},
+				Required: true,
+			},
+		},
+	}
+}
+
+type NsxClusterNode struct {
+	ID        string
+	IPAddress string
+	UserName  string
+	Password  string
+	Fqdn      string
+	Status    string
+}
+
+func getClusterNodesFromSchema(d *schema.ResourceData) []NsxClusterNode {
+	nodes := d.Get("node").([]interface{})
+	var clusterNodes []NsxClusterNode
+	for _, node := range nodes {
+		data := node.(map[string]interface{})
+		id := data["id"].(string)
+		ipAddress := data["ip_address"].(string)
+		userName := data["username"].(string)
+		password := data["password"].(string)
+		nodeObj := NsxClusterNode{
+			ID:        id,
+			IPAddress: ipAddress,
+			UserName:  userName,
+			Password:  password,
+		}
+		clusterNodes = append(clusterNodes, nodeObj)
+	}
+	return clusterNodes
+}
+
+func setNodesInSchema(d *schema.ResourceData, nodes []NsxClusterNode) error {
+	// Retrieve node credential from schema and set them in the nodeList element
+	// This is because the nodes are obtained using cluster client, and does not
+	// contain any information about credential
+	nodeIPToCredentialMap := getIPtoCredentialMap(d)
+
+	var nodeList []map[string]interface{}
+	for _, node := range nodes {
+		_, nodeInSchema := nodeIPToCredentialMap[node.IPAddress]
+		elem := make(map[string]interface{})
+		elem["id"] = node.ID
+		elem["ip_address"] = node.IPAddress
+		// If node is not in schema but in cluster node list,
+		// it means the node needs to be removed.
+		// In this case we don't need its credential because
+		// remove api does not require node credential
+		if nodeInSchema {
+			elem["username"] = nodeIPToCredentialMap[node.IPAddress][0]
+			elem["password"] = nodeIPToCredentialMap[node.IPAddress][1]
+		}
+		elem["fqdn"] = node.Fqdn
+		elem["status"] = node.Status
+		nodeList = append(nodeList, elem)
+	}
+	return d.Set("node", nodeList)
+}
+
+func getIPtoCredentialMap(d *schema.ResourceData) map[string][]string {
+	// returns a map, whose key is IP address of a node,
+	// value is a slice of [node_login_username, node_login_password]
+	nodes := d.Get("node").([]interface{})
+	var idToCredentialMap = map[string][]string{}
+	for _, node := range nodes {
+		data := node.(map[string]interface{})
+		ipAddress := data["ip_address"].(string)
+		userName := data["username"].(string)
+		password := data["password"].(string)
+		cred := []string{userName, password}
+		idToCredentialMap[ipAddress] = cred
+	}
+	return idToCredentialMap
+}
+
+func resourceNsxtManagerClusterCreate(d *schema.ResourceData, m interface{}) error {
+	// Call Joincluster function on nodes that are not in the cluster
+	nodes := getClusterNodesFromSchema(d)
+	if len(nodes) == 0 {
+		return fmt.Errorf("At least a manager appliance must be provided to form a cluster")
+	}
+	clusterID, certSha256Thumbprint, hostIP, err := getClusterInfoFromHostNode(d, m)
+	if err != nil {
+		return err
+	}
+
+	for _, guestNode := range nodes {
+		err := joinNodeToCluster(clusterID, certSha256Thumbprint, guestNode, hostIP, d, m)
+		if err != nil {
+			return fmt.Errorf("Failed to join node %s: %s", guestNode.ID, err)
+		}
+	}
+	d.SetId(clusterID)
+	return resourceNsxtManagerClusterRead(d, m)
+}
+
+func getClusterInfoFromHostNode(d *schema.ResourceData, m interface{}) (string, string, string, error) {
+	// function return values are:
+	// clusterID, certSha256Thumbprint, hostIP, error
+	connector := getPolicyConnector(m)
+	client := nsx.NewClusterClient(connector)
+	c := m.(nsxtClients)
+	min := c.CommonConfig.MinRetryInterval
+	max := c.CommonConfig.MaxRetryInterval
+	maxRetries := c.CommonConfig.MaxRetries
+	hostIP := ""
+	for i := 0; i < maxRetries; i++ {
+		clusterConfig, err := client.Get()
+		if err != nil {
+			return "", "", "", handleReadError(d, "Cluster Config", "", err)
+		}
+		if hostIP == "" {
+			hostIP, err = getHostIPFromClusterConfig(m, clusterConfig)
+			if err != nil {
+				return "", "", "", err
+			}
+		}
+		clusterID := *clusterConfig.ClusterId
+		nodes := clusterConfig.Nodes
+		node := nodes[0]
+		apiListenAddr := node.ApiListenAddr
+		if apiListenAddr != nil {
+			certSha256Thumbprint := *apiListenAddr.CertificateSha256Thumbprint
+			return clusterID, certSha256Thumbprint, hostIP, nil
+		}
+		interval := (rand.Intn(max-min) + min)
+		time.Sleep(time.Duration(interval) * time.Millisecond)
+		log.Printf("[DEBUG]: Waited %d ms before retrying getting API Listen Address, attempt %d", interval, i+1)
+	}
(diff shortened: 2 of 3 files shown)


nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,173 @@
...
+				Type:        schema.TypeString,
+				Description: "ID of the NSXT cluster",
+				Computed:    true,
+			},
+			"nodes": {
+				Type:        schema.TypeList,
+				Description: "IP addresses of nodes in the cluster",
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validation.IsIPv4Address,
```
annakhm, line 40: I think IPv6 is also supported?

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,173 @@
...
+		},
+	}
+}
+
+func resourceNsxtManagerClusterCreate(d *schema.ResourceData, m interface{}) error {
+	nodes := getStringListFromSchemaList(d, "nodes")
+	if len(nodes) == 0 {
+		return fmt.Errorf("Please provide at least one node ip")
+	} else if len(nodes) == 1 {
+		log.Printf("[INFO] Only one node ip provided, skip joining nodes process")
```
annakhm, line 64: Should the user provide master node in the list?

2ez4szliu: I'm thinking if we want to assume that the host node in terraform provider config will be on this list.
For example nodes    = ["192.168.240.31", "192.168.240.32", "192.168.240.33"]
and in terrraform config we have 
provider "nsxt" {
  host                 = "192.168.240.31"
  username             = "admin"
  password             = "password"
  allow_unverified_ssl = true
  max_retries          = 2
}

Then the master node will be 192.168.240.31

If we want to use the provider to join nodes that does not include the host node for the provider, for example
["192.168.240.32", "192.168.240.33", "192.168.240.34"]
Then we should probably includes an argument master_node in the schema

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,173 @@
...
+		return nil
+	}
+	newClients.NsxtClient = newAPIClient
+	return newClients
+}
+
+func joinNodeToCluster(clusterID string, certSha256Thumbprint string, workerNodeIP string, masterNodeIP string, d *schema.ResourceData, m interface{}) error {
+	newNsxClients := getOverrideNsxtClient(workerNodeIP, d, m).(nsxtClients)
+	connector := getPolicyConnector(newNsxClients)
+	client := nsx.NewClusterClient(connector)
```
annakhm, line 130: This connector would connect to master node from provider config, and I think here we're supposed to connect to the node API endpoint?

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,173 @@
...
+package nsxt
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	api "github.com/vmware/go-vmware-nsxt"
```
annakhm, line 13: we should only use the new SDK

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,286 @@
...
+	return newClients, nil
+}
+
+func configureNewClient(newClient *nsxtClients, oldClient *nsxtClients, host string, username string, password string) error {
+	newClient.Host = host
+	securityCtx := core.NewSecurityContextImpl()
+	oldSecurityCtx := oldClient.PolicySecurityContext
+	oldSecurityCtxData := oldSecurityCtx.GetAllProperties()
+	authScheme, securityContextNeeded := oldSecurityCtxData[security.AUTHENTICATION_SCHEME_ID]
+	if securityContextNeeded {
```
ksamoray, line 184: Some code here is very similar to configurePolicyConnectorData(), worthwhile to consolidate these if possible

nsxt/resource_nsxt_manager_cluster.go
2ez4szliu: Done. changed to IsIPAddress.

nsxt/resource_nsxt_manager_cluster.go
2ez4szliu: So I think we will be using the host in provider config as master node, in .tf file for the resource we only provide the nodes that will be joining into the cluster

nsxt/resource_nsxt_manager_cluster.go
2ez4szliu: removed

vendor/github.com/vmware/go-vmware-nsxt/api_client.go
```diff
@@ -93,6 +93,10 @@ func GetContext(cfg *Configuration) context.Context {
 	return context.Background()
 }
 
+func (c *APIClient) GetCfg() *Configuration {
```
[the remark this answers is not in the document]
2ez4szliu: Not sure why it's not working, but i manually fixed it

nsxt/resource_nsxt_manager_cluster.go
2ez4szliu: I feel like the part that can be put into one function is the part of configuring security context, I have added a function for this part

nsxt/provider.go
```diff
@@ -627,35 +628,9 @@ func configurePolicyConnectorData(d *schema.ResourceData, clients *nsxtClients)
...
-			}
-
-			if password == "" {
-				return fmt.Errorf("password must be provided")
-			}
-
-			securityCtx.SetProperty(security.AUTHENTICATION_SCHEME_ID, security.USER_PASSWORD_SCHEME_ID)
-			securityCtx.SetProperty(security.USER_KEY, username)
-			securityCtx.SetProperty(security.PASSWORD_KEY, password)
+		err := configureSecurityContext(securityCtx, clients, vmcAccessToken, vmcAuthHost, vmcAuthMode, username, password)
```
annakhm, line 631: would it make sense to return `securityCtx` rather than passing it as a parameter?

annakhm: Also since VMC is not relevant for compute manager work, I'd suggest to leave the VMC part out of the function, which can be called `getBasicAuthSecurityContext` and become much simpler

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,323 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"revision": getRevisionSchema(),
+			"cluster_id": {
+				Type:        schema.TypeString,
+				Description: "ID of the NSXT cluster",
+				Computed:    true,
+			},
+			"nodes": {
```
annakhm, line 37: this should be `node` per convention

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,323 @@
...
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:        schema.TypeString,
+							Description: "ID of the cluster node",
+							Computed:    true,
+						},
+						"ip_address": {
+							Type:         schema.TypeString,
+							Description:  "IP addresses of the cluster node that will join the cluster of the host node",
```
annakhm, line 49: nit - plural is not needed here

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,323 @@
...
+					},
+				},
+				Required: true,
+			},
+		},
+	}
+}
+
+type NsxClusterNode struct {
+	ID        *string
```
annakhm, line 73: I would suggest not to use pointers here to keep things simple and safe

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,323 @@
...
+			UserName:  &userName,
+			Password:  &password,
+		}
+		clusterNodes = append(clusterNodes, nodeObj)
+	}
+	return clusterNodes
+}
+
+func setNodesInSchema(d *schema.ResourceData, nodes []NsxClusterNode) error {
+	nodeIPToCredentialMap := getIPtoCredentialMap(d)
```
annakhm, line 100: Could you please explain why is map needed here?

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+	d.SetId(clusterID)
+	d.Set("cluster_id", clusterID)
+	return resourceNsxtManagerClusterRead(d, m)
+}
+
+func getHostIPFromNsxtClient(client interface{}) (string, bool) {
+	c := client.(nsxtClients)
+	hostIP := c.Host
+	hostIP = strings.TrimPrefix(hostIP, "https://")
```
annakhm, line 158: This is not necessarily IP address. Can we resolve the host to IP programmatically if fqdn is specified?

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,323 @@
...
+	ip := net.ParseIP(hostIP)
+	v4 := ip.To4()
+	return hostIP, v4 != nil
+}
+
+func getClusterInfoFromHostNode(d *schema.ResourceData, m interface{}) (string, string, error) {
+	connector := getPolicyConnector(m)
+	client := nsx.NewClusterClient(connector)
+	clusterConfig, err := client.Get()
```
annakhm, line 167: Is default cluster always present?

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,323 @@
...
+	if err != nil {
+		return "", "", fmt.Errorf("Failed to get cluster info from host node: %s", err)
+	}
+	clusterID := *clusterConfig.ClusterId
+	nodes := clusterConfig.Nodes
+	node := nodes[0]
+	apiListenAddr := node.ApiListenAddr
```
annakhm, line 174: Related to above comment - can main node IP be retrieved from here if not specified in provider section?

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,323 @@
...
+	return newClients, nil
+}
+
+func configureNewClient(newClient *nsxtClients, oldClient *nsxtClients, host string, username string, password string) error {
+	newClient.Host = host
+	securityCtx := core.NewSecurityContextImpl()
+	oldSecurityCtx := oldClient.PolicySecurityContext
+	oldSecurityCtxData := oldSecurityCtx.GetAllProperties()
+	_, securityContextNeeded := oldSecurityCtxData[security.AUTHENTICATION_SCHEME_ID]
+	if securityContextNeeded {
```
annakhm, line 206: security context should always be needed here, since we don't support client cert for node joining, and remote authentication is also irrelevant

nsxt/resource_nsxt_manager_cluster.go
2ez4szliu: The credential(username/password) is something we used when initialize a new client on for the new node. However in read function, this information is not in the clusterConfig that we obtained fromthe cluster client, so when we setNodesInSchema we need to a way to first retrieve the username/password from old schema, then set them in the elem, that's why we use this map to store such information
`elem["username"] = nodeIPToCredentialMap[*node.IPAddress][0]`
`elem["password"] = nodeIPToCredentialMap[*node.IPAddress][1]`

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,323 @@
...
+		err := configureSecurityContext(securityCtx, newClient, "", "", "", username, password)
+		if err != nil {
+			return fmt.Errorf("Failed to configure new client with host %s: %s", host, err)
+		}
+	}
+	newClient.PolicyHTTPClient = oldClient.PolicyHTTPClient
+	if securityContextNeeded {
+		newClient.PolicySecurityContext = securityCtx
+	}
+	newClient.PolicyEnforcementPoint = oldClient.PolicyEnforcementPoint
```
annakhm, line 216: Those two lines should not be required. Lets avoid using `oldClient` unless absolutely necessary

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,323 @@
...
+		return resourceNsxtManagerClusterRead(d, m)
+	}
+	hostIP, _ := getHostIPFromNsxtClient(m)
+	clusterID, certSha256Thumbprint, err := getClusterInfoFromHostNode(d, m)
+	if err != nil {
+		return err
+	}
+
+	for _, guestNode := range nodes {
+		err := joinNodeToCluster(clusterID, certSha256Thumbprint, guestNode, hostIP, d, m)
```
annakhm, line 145: Lets add some logs here. It would be important to know which nodes were joined successfully and which failed.

nsxt/resource_nsxt_manager_cluster.go
2ez4szliu: Yes I think it can be, but cluster config contains information on all cluster nodes, how do we know which node is the host? Maybe by checking if its ip address is not equal to ips of nodes in the schema?

nsxt/resource_nsxt_manager_cluster.go
2ez4szliu: To my knowledge if the host is deployed with OVA template then it will always have a cluster ID, even if the cluster has only one member

nsxt/resource_nsxt_manager_cluster.go
2ez4szliu: done. removed this check

nsxt/resource_nsxt_manager_cluster.go
2ez4szliu: done

nsxt/resource_nsxt_manager_cluster.go
2ez4szliu: done

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,323 @@
...
+		Read:   resourceNsxtManagerClusterRead,
+		Update: resourceNsxtManagerClusterUpdate,
+		Delete: resourceNsxtManagerClusterDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"revision": getRevisionSchema(),
+			"cluster_id": {
```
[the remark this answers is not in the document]
2ez4szliu: removed this field

nsxt/provider.go
2ez4szliu: I think this is the provider's configurePolicyConnectorData function, and vmc parameters are still needed here?

nsxt/resource_nsxt_manager_cluster.go
2ez4szliu: done

nsxt/resource_nsxt_manager_cluster.go
annakhm: I don't think other nodes would be part of the cluster at this stage?

2ez4szliu: We are also using getClusterInfoFromHostNode in the update function, in that case there could be more nodes in the cluster

nsxt/resource_nsxt_manager_cluster.go
annakhm: Got it, thanks. I think we should add some computed info to those nodes, such as `status`, `fqdn`.

nsxt/resource_nsxt_manager_cluster.go
annakhm: Is that necessary? we can save needed info on the resource as Computed field

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,414 @@
...
+		clusterNodeIDs = append(clusterNodeIDs, *id)
+	}
+
+	clusterID, certSha256Thumbprint, hostIP, err := getClusterInfoFromHostNode(d, m)
+	isIPv4 := (net.ParseIP(hostIP)).To4() != nil
+	nodes := getClusterNodesFromSchema(d)
+	var nodesIDFromSchema []string
+	var nodesToAdd []NsxClusterNode
+	var nodesToRemove []string
+	for _, node := range nodes {
```
annakhm, line 362: You can get node changes from the schema with `oldNodes, newNodes := d.GetChange("node")`

nsxt/provider.go
```diff
@@ -210,6 +210,15 @@ func Provider() *schema.Provider {
 				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("NSXT_CA", nil),
 			},
+			// This config is set to true when need to create NSXT manager cluster resource with
+			// vms from vsphere provider as host. Skipping version check will allow terraform
+			// to apply the template while NSX host vm is still being configured.
+			"skip_version_check": {
```
annakhm, line 216: The version check should not be skipped, but performed once on demand whenever connector is first used.
In addition to version check, NSX connection is also made to NSX client in `configureNsxtClient` to retrieve session token. This should also happen on demand. I suggest to take this piece to a separate PR.

2ez4szliu: Got it, I have removed it from code

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,459 @@
...
+	return resourceNsxtManagerClusterRead(d, m)
+}
+
+func getClusterInfoFromHostNode(d *schema.ResourceData, m interface{}) (string, string, string, error) {
+	// function return values are:
+	// clusterID, certSha256Thumbprint, hostIP, error
+	connector := getPolicyConnector(m)
+	client := nsx.NewClusterClient(connector)
+	hostIP := ""
+	for i := 0; i < maxGetClusterConfigRetry; i++ {
```
2ez4szliu, line 187: The node's API Listen Address will show up after node management ip is up and the cluster config becomes available.
Since this is the only way to obtain the correct certsha256thumbprint, I added a retry logic here to retry when apiListenAddress is not up

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,459 @@
...
+
+	"golang.org/x/exp/slices"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx"
+	nsxModel "github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/model"
+)
+
+var MaxGetClusterConfigRetry = 20
```
annakhm, line 21: I suggest we use parameters configured on provider, like `commonProviderConfig.MaxRetries`

website/docs/r/manager_cluster.html.markdown
```diff
@@ -0,0 +1,40 @@
...
+layout: "nsxt"
+page_title: "NSXT: nsxt_manager_cluster"
+description: A resource to configure an NSXT cluster.
+---
+
+# nsxt_manager_cluster
+
+This resource provides a method for creating an NSXT cluster with several nodes
+This resource is supported with NSX 4.1.0 onwards.
+
```
annakhm, line 12: I think we need some details here, mentioning that the main node is specified in provider config

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,459 @@
...
+
+	}
+	d.Set("revision", clusterConfig.Revision)
+	setNodesInSchema(d, nodes)
+	return nil
+}
+
+func getIPFromNodeInfo(node nsxModel.ClusterNodeInfo, isIPv4 bool) string {
+	// After join node into cluster, apiListen address may take some time to become available
+	// if it's not available, we retrieve node ip from nodeEntityInfo
```
annakhm, line 359: Are we expecting this problem to be solved with retry? If yes, can this logic be removed?

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,459 @@
...
+		}
+	}
+	for _, node := range newNodes.([]interface{}) {
+		nodeMap := node.(map[string]interface{})
+		ip := nodeMap["ip_address"].(string)
+		if !slices.Contains(oldNodesIPs, ip) {
+			userName := nodeMap["username"].(string)
+			password := nodeMap["password"].(string)
+			ip := nodeMap["ip_address"].(string)
+			fqdn := nodeMap["fqdn"].(string)
```
annakhm, line 416: fqdn and status are computed, I don't think we need to pass those here?

nsxt/resource_nsxt_manager_cluster.go
2ez4szliu: Got it, in that case we need to first convert the client from interface type to nsxtclients, then user clients.CommonConfig.MaxRetries

nsxt/resource_nsxt_manager_cluster.go
2ez4szliu: yes I think you are right

nsxt/resource_nsxt_manager_cluster.go
2ez4szliu: So I think it can be solved, but here we don't need to retry because node ip is available in nodeEntityInfo, it could save some time.
Unlike getting API CertSha256Thumbprint, we have to retry to wait for APIListenAddress because it's the only way to obtain the correct thumbprint

website/docs/r/manager_cluster.html.markdown
2ez4szliu: Done. Also mentioned in the last few lines that import is not supported for this resource

nsxt/resource_nsxt_manager_cluster.go
annakhm: I would suggest to always grab the IP from https entity then, because its would mean less code and more deterministic behavior

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,455 @@
...
+				UserName:  userName,
+				Password:  password,
+			}
+			err = joinNodeToCluster(clusterID, certSha256Thumbprint, nodeObj, hostIP, d, m)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
```
annakhm, line 427: I wonder what should happen if username or password are updated on a node? I think according to this code no API calls would happen, which is probably correct, since we don't need to rejoin.

nsxt/resource_nsxt_manager_cluster.go
2ez4szliu: makes sense, I've made the changes

nsxt/resource_nsxt_manager_cluster.go
2ez4szliu: I think nothing will happen in that case, if they are already in cluster then any change in username/password will not affect them. If we want to delete them, it still does not matter because the remove node api does not require guest node credential

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,446 @@
...
+		userName := data["username"].(string)
+		password := data["password"].(string)
+		fqdn := data["fqdn"].(string)
+		status := data["status"].(string)
+		nodeObj := NsxClusterNode{
+			ID:        id,
+			IPAddress: ipAddress,
+			UserName:  userName,
+			Password:  password,
+			Fqdn:      fqdn,
```
annakhm, line 101: We don't need to include computed fields when reading from schema. Lets include the required minimum, I believe it would improve code readability.

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,446 @@
...
+		cred := []string{userName, password}
+		idToCredentialMap[ipAddress] = cred
+	}
+	return idToCredentialMap
+}
+
+func resourceNsxtManagerClusterCreate(d *schema.ResourceData, m interface{}) error {
+	// Call Joincluster function on nodes that are not in the cluster
+	nodes := getClusterNodesFromSchema(d)
+	if len(nodes) == 0 {
```
annakhm, line 155: Since `node` is required attribute in schema, this check seems redundant

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,446 @@
...
+		return fmt.Errorf("Please provide at least one node ip")
+	} else if len(nodes) == 1 {
+		log.Printf("[INFO] Only one node ip provided, skip joining nodes process")
```
annakhm, line 158: I think this is not correct any longer, since we're not including main node in the list anymore?

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,446 @@
...
+	}
+	return nil
+}
+
+func joinNodeToCluster(clusterID string, certSha256Thumbprint string, guestNode NsxClusterNode, masterNodeIP string, d *schema.ResourceData, m interface{}) error {
+	c, err := getNewNsxtClient(guestNode, d, m)
+	if err != nil {
+		return err
+	}
+	newNsxClients := c.(nsxtClients)
```
annakhm, line 297: I think log printouts are required for each joining node or node being removed. I would suggest `info` level logs for beginning/end of the process.

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,446 @@
...
+	client := nsx.NewClusterClient(connector)
+	c := m.(nsxtClients)
+	min := c.CommonConfig.MinRetryInterval
+	max := c.CommonConfig.MaxRetryInterval
+	maxRetries := c.CommonConfig.MaxRetries
+	hostIP := ""
+	for i := 0; i < maxRetries; i++ {
+		clusterConfig, err := client.Get()
+		if err != nil {
+			return "", "", "", fmt.Errorf("Failed to get cluster info from host node: %s", err)
```
annakhm, line 189: This error message would not be detailed enough, only error type would be printed out. Please refer to `handleReadError` for reference on printing error details. Same applies to other error messages below.

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,446 @@
...
+	_, err = client.Joincluster(joinClusterParams)
+	if err != nil {
+		return fmt.Errorf("Failed to join node to cluster: %s, node ip address: %s", err, guestNode.IPAddress)
+	}
+	return nil
+}
+
+func getHostCredential(m interface{}) (string, string) {
+	nsxtClient := m.(nsxtClients)
+	username := nsxtClient.NsxtClientConfig.UserName
```
annakhm, line 317: `nsxtClientConfig` is a struct used to initialize old sdk. We shouldn't rely on it, please use `CommonConfig` to store provider data intended for later use.

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,441 @@
...
+	client := nsx.NewClusterClient(connector)
+	nodes := getClusterNodesFromSchema(d)
+	force := "true"
+	gracefulShutdown := "true"
+	ignoreRepositoryIPCheckParam := "false"
+	for _, node := range nodes {
+		guestNodeID := node.ID
+		_, err := client.Removenode(guestNodeID, &force, &gracefulShutdown, &ignoreRepositoryIPCheckParam)
+		if err != nil {
+			return fmt.Errorf("Failed to remove node %s from cluster: %s", guestNodeID, err)
```
annakhm, line 437: Similar to above, lets use `handleDeleteError` here to provide more details to the user

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -0,0 +1,441 @@
...
+		nodeMap := node.(map[string]interface{})
+		ip := nodeMap["ip_address"].(string)
+		if !slices.Contains(newNodesIPs, ip) {
+			id := nodeMap["id"].(string)
+			force := "true"
+			gracefulShutdown := "true"
+			ignoreRepositoryIPCheckParam := "false"
+			_, err := client.Removenode(id, &force, &gracefulShutdown, &ignoreRepositoryIPCheckParam)
+			if err != nil {
+				return err
```
annakhm, line 391: `handleUpdateError` here and below

nsxt/resource_nsxt_manager_cluster.go
annakhm: Thanks. Logs are still missing in remove workflow.

salv-orlando: Got it

annakhm (approved):
The rest of comments will be addressed in a follow-up

PR #931, opened 2023-08-01
Validate order of rules is consistent with seq

annakhm:
In security policy and gateway policy, user can specify sequence numbers for the rules. This change validates their consistency with rule order upon apply and throws an error if validation fails. Before this change, inconsistent sequence number in rules would be applied and cause a permanent non-empty diff.

### DIFF
--- a/nsxt/policy_common.go
+++ b/nsxt/policy_common.go
@@ -409,6 +409,26 @@ func setPolicyRulesInSchema(d *schema.ResourceData, rules []model.Rule) error {
 	return d.Set("rule", rulesList)
 }
 
+func validatePolicyRuleSequence(d *schema.ResourceData) error {
+	rules := d.Get("rule").([]interface{})
+	latestNum := int64(0)
+	for _, rule := range rules {
+		data := rule.(map[string]interface{})
+		sequenceNumber := int64(data["sequence_number"].(int))
+		if sequenceNumber > 0 && sequenceNumber <= latestNum {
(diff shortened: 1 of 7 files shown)


nsxt/policy_common.go
```diff
@@ -409,6 +409,27 @@ func setPolicyRulesInSchema(d *schema.ResourceData, rules []model.Rule) error {
 	return d.Set("rule", rulesList)
 }
 
+func validatePolicyRuleSequence(d *schema.ResourceData) error {
+	rules := d.Get("rule").([]interface{})
+	latestNum := int64(0)
+	for _, rule := range rules {
+		data := rule.(map[string]interface{})
+		seq := data["sequence_number"].(int)
+		sequenceNumber := int64(seq)
```
ksamoray, line 418: Could be
`sequenceNumber := int64(data["sequence_number"].(int))`

as `seq` isn't used elsewhere

nsxt/policy_common.go
```diff
@@ -409,6 +409,27 @@ func setPolicyRulesInSchema(d *schema.ResourceData, rules []model.Rule) error {
...
+		if sequenceNumber > 0 && sequenceNumber <= latestNum {
+			return fmt.Errorf("When sequence_number is specified in a rule, it must be consistent with rule order. To avoid confusion, it is recommended to either specify sequence numbers in all rules, or none")
```
ksamoray, line 420: Style wise it's better not to start an error string with a capital letter

ksamoray (approved):

PR #933, opened 2023-08-03
Add Tier0 route map and prefix list data sources

annakhm:


nsxt/data_source_nsxt_policy_gateway_prefix_list.go
```diff
@@ -0,0 +1,40 @@
+/* Copyright © 2022 VMware, Inc. All Rights Reserved.
```
GraysonWu, line 1: Should we use 2023 here?

GraysonWu (approved):
LGTM.

PR #937, opened 2023-08-09
[WIP] Transport node resource

ksamoray:
Implement transport node resource


nsxt/resource_nsxt_transport_node.go
```diff
@@ -0,0 +1,1959 @@
...
+		},
+	}
+}
+
+func getNodeSchema(addlAttributes map[string]*schema.Schema, addExactlyOneOf bool) *schema.Schema {
+	elemSchema := map[string]*schema.Schema{
+		"external_id": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: "the Node maintained on the Node and used to recognize the Node",
```
annakhm, line 149: This description is confusing :)

nsxt/resource_nsxt_transport_node.go
```diff
@@ -0,0 +1,1959 @@
...
+								Type:        schema.TypeString,
+								Optional:    true,
+								Description: "Compute folder identifier in the specified vcenter server",
+							},
+							"compute_id": {
+								Type:        schema.TypeString,
+								Required:    true,
+								Description: "Cluster identifier or resourcepool identifier for specified vcenter server",
+							},
+							"data_network_ids": {
```
annakhm, line 253: If this can be segment path, perhaps we should call it `data_networks` to avoid confusion

nsxt/resource_nsxt_transport_node.go
```diff
@@ -0,0 +1,1959 @@
...
+								Description: "This flag represents whether IPv4 configuration is enabled or not",
+								Optional:    true,
+								Default:     true,
+							},
+							"management_network_id": {
+								Type:        schema.TypeString,
+								Required:    true,
+								Description: "Portgroup, logical switch identifier or segment path for management network connectivity",
+							},
+							"management_port_subnets": {
```
annakhm, line 290: `management_port_subnets` => `management_port_subnet`?

nsxt/resource_nsxt_transport_node.go
```diff
@@ -0,0 +1,1959 @@
...
+}
+
+func getHostSwitchProfileIDsSchema() *schema.Schema {
+	return &schema.Schema{
+		Type:        schema.TypeList,
+		Description: "Identifiers of host switch profiles to be associated with this host switch",
+		Optional:    true,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"key": {
```
annakhm, line 766: Those are de-facto `type` and `id`. Should we consider flattening those profiles out (as we do in other resources) for user convenience?

nsxt/resource_nsxt_transport_node.go
```diff
@@ -0,0 +1,1959 @@
...
+					Optional: true,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"default_gateway": {
+								Type:         schema.TypeString,
+								Required:     true,
+								Description:  "Gateway IP",
+								ValidateFunc: validateSingleIP(),
+							},
+							"ip_list": {
```
annakhm, line 821: we normally call list of IPs `ip_addresses` in the provider, lets do the same for consistency here?

nsxt/resource_nsxt_transport_node.go
```diff
@@ -0,0 +1,1959 @@
...
+					Optional: true,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"default_gateway": {
+								Type:         schema.TypeString,
+								Required:     true,
+								Description:  "Gateway IP",
+								ValidateFunc: validateSingleIP(),
+							},
+							"ip_mac_list": {
```
annakhm, line 852: this would be sub-clause, not a list

nsxt/resource_nsxt_transport_node.go
```diff
@@ -0,0 +1,1959 @@
...
+							"subnet_mask": {
+								Type:         schema.TypeString,
+								Required:     true,
+								Description:  "Subnet mask",
+								ValidateFunc: validateSingleIP(),
+							},
+						},
+					},
+				},
+				"static_ip_pool_spec": {
```
annakhm, line 883: any reason for this to be a list?

nsxt/resource_nsxt_transport_node.go
```diff
@@ -0,0 +1,1959 @@
...
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx"
```
annakhm, line 14: Looks like transport node APIs are exposed via policy API `/infra/sites/{site-id}/enforcement-points/{enforcementpoint-id}/host-transport-nodes/`
Ideally we should leverage policy API. If we're sticking to MP API, host switch profiles should also be exposed via MP API

nsxt/resource_nsxt_transport_node.go
ksamoray: AFAIK this is an id of a vSphere network object

nsxt/resource_nsxt_transport_node.go
ksamoray: Why? Per the [spec,](https://vdc-download.vmware.com/vmwb-repository/dcr-public/201a09cb-5610-49e6-9d31-7db0920e3a65/65f199d6-8516-4092-9b19-906559f972b5/api_includes/types_StaticIpMacListSpec.html) it could contain 32 items

nsxt/resource_nsxt_transport_node.go
ksamoray: I've defined `assigned_by_dhcp`, `static_ip_pool_spec` as lists as ExactlyOneOf demanded to have lists of MaxItems=1. Yet also the parent objects should be so and this is not the case for `transport_node_profile_sub_config`. I'm looking for some solution for this.

nsxt/resource_nsxt_transport_node.go
annakhm: Sorry, I didn't phrase this well. For hcl perspective, this would be a subclause rather than a list: 
```ip_mac {
    ...
   }```
so we shouldn't call it `list`. Maybe `ip_mac` or `ip_mac_pair`

File: nsxt/resource_nsxt_transport_node.go [hunk 9]
```diff
@@ -0,0 +1,1987 @@
...
+	client := nsx.NewTransportNodesClient(connector)
+
+	obj, err := getTransportNodeFromSchema(d)
+	if err != nil {
+		return err
+	}
+	obj1, err := client.Create(*obj)
+	if err != nil {
+		id := ""
+		if obj1.Id != nil {
```
annakhm, line 920: I don't think object would be sent back in case of error in `POST`, response will be error details

nsxt/resource_nsxt_transport_node.go
```diff
@@ -0,0 +1,1987 @@
...
+func getPreconfiguredHostSwitchSchema() *schema.Schema {
+	return &schema.Schema{
+		Type:        schema.TypeList,
+		MaxItems:    1,
+		Optional:    true,
+		Description: "Preconfigured host switch",
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"endpoint": {
+					Type:        schema.TypeList,
```
annakhm, line 558: Do we need this to be a list?

website/docs/r/transport_node.html.markdown
```diff
@@ -0,0 +1,276 @@
...
+
+```hcl
+resource "nsxt_transport_node" "test" {
+  description  = "Terraform-deployed edge node"
+  display_name = "tf_edge_node"
+  standard_host_switch {
+    host_switch_mode = "STANDARD"
+    host_switch_type = "NVDS"
+    ip_assignment {
+      static_ip_pool_id = data.nsxt_ip_pool.ipp1.id
```
annakhm, line 23: we should avoid suggesting deprecated objects in docs, but we can refine this later

website/docs/r/transport_node.html.markdown
```diff
@@ -0,0 +1,276 @@
...
+    * `syslog_server` - (Optional) List of Syslog servers.
+      * `log_level` - (Optional) Log level to be redirected. Accepted values - 'EMERGENCY', 'ALERT', 'CRITICAL', 'ERROR', 'WARNING', 'NOTICE', 'INFO' or 'DEBUG'. The default value is 'INFO'.
+      * `name` - (Optional) Display name of the syslog server.
+      * `port` - (Optional) Syslog server port. Defaults to 514.
+      * `protocol` - (Optional) Syslog protocol. Accepted values - 'TCP', 'UDP', 'TLS', 'LI', 'LI_TLS'. The default value is 'UDP'.
+      * `server` - (Required) Server IP or fqdn.
+* `remote_tunnel_endpoint` - (Optional) Configuration for a remote tunnel endpoint.
+  * `host_switch_name` - (Required) The host switch name to be used for the remote tunnel endpoint.
+  * `ip_assignment` - (Required) - Specification for IPs to be used with host switch virtual tunnel endpoints. Should contain exatly one of the below:
+    * `assigned_by_dhcp` - (Optional) Enables DHCP assignment. Should be set to true.
```
annakhm, line 240: I like the description :)

nsxt/resource_nsxt_transport_node.go
```diff
@@ -0,0 +1,1987 @@
...
+		MaxItems:    1,
+		Required:    true,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"assigned_by_dhcp": {
+					Type:        schema.TypeBool,
+					Optional:    true,
+					Description: "Enables DHCP assignment. Should be set to true",
+				},
+				"static_ip_list": {
```
annakhm, line 793: If this is single entry, should we call this `list`?

nsxt/resource_nsxt_transport_node.go
```diff
@@ -0,0 +1,1987 @@
...
+							"subnet_mask": {
+								Type:         schema.TypeString,
+								Required:     true,
+								Description:  "Subnet mask",
+								ValidateFunc: validateSingleIP(),
+							},
+						},
+					},
+				},
+				"static_ip_mac_list": {
```
annakhm, line 825: same as above

nsxt/resource_nsxt_transport_node.go
```diff
@@ -0,0 +1,1987 @@
...
+}
+
+func getTransportNodeFromSchema(d *schema.ResourceData) (*model.TransportNode, error) {
+	description := d.Get("description").(string)
+	displayName := d.Get("display_name").(string)
+	tags := getMPTagsFromSchema(d)
+	failureDomain := d.Get("failure_domain").(string)
+	hostSwitchSpec, err := getHostSwitchSpecFromSchema(d)
+	if err != nil {
+		return nil, handleCreateError("TransportNode", "", err)
```
annakhm, line 886: `handleCreateError` (which translates to `logAPIError`) is based on policy model error structure. We probably want to code similar functions (hopefully with code reuse) to handle MP model error structure. This can be handled in a separate PR.

nsxt/resource_nsxt_transport_node.go
ksamoray: I had the same thought - the fact that the API uses list here is strange and I assumed that it could change to multiple endpoints at some point - and then we'll have to support some legacy. What's your thought about this?

website/docs/r/transport_node.html.markdown
ksamoray: Can we use the policy IP pool in the MP transport node? How do we gather its id?

website/docs/r/transport_node.html.markdown
ksamoray: The API [AssignedByDhcp](https://vdc-download.vmware.com/vmwb-repository/dcr-public/201a09cb-5610-49e6-9d31-7db0920e3a65/65f199d6-8516-4092-9b19-906559f972b5/api_includes/types_AssignedByDhcp.html) object has nothing but resource type - which we do not add in TF. Any suggestion for how to make this more user-understandable? Or replace the bool with anything else?

nsxt/resource_nsxt_transport_node.go
ksamoray: I followed the API naming. However this assignment mode allows multiple IPs (ip_addresses), so technically the name reflects using multiple IPs, right?

nsxt/resource_nsxt_transport_node.go
ksamoray: Same - the API allows multiple pairs here as well.

website/docs/r/transport_node.html.markdown
annakhm: Realization data source should provide the realized ID

nsxt/resource_nsxt_transport_node.go
annakhm: Since this API is going to be rewritten with policy, I think it is unlikely it would evolve significantly.

nsxt/resource_nsxt_transport_node.go
annakhm: I see what you mean, still I would suggest not to use `list` because its not in line with naming conventions that is used so far in the provider. So `static_ip` or `static_ips` would be more consistent.

nsxt/resource_nsxt_transport_node.go
```diff
@@ -0,0 +1,1983 @@
...
+	n := 0
+	for _, iaType := range ipAssignmentTypes {
+		d = data[iaType]
+		if (isString(d) && d != nil) || (isSlice(d) && len(d.([]interface{})) > 0) || (isBool(d) && d.(bool)) {
+			t, d = iaType, data[iaType]
+			n++
+		}
+	}
+	if n > 1 {
+		return "", nil, fmt.Errorf("exactly one IP assignment is allowed for ip_assinment object")
```
annakhm, line 1355: typo `ip_assinment` => `ip_assignment`

website/docs/r/transport_node.html.markdown
ksamoray: OK. As I'm not sure how to use realization info data source for IP pool, I'll use DHCP for the example :)

nsxt/resource_nsxt_transport_node.go
ksamoray: Cool so this can be flattened.

annakhm (approved):

PR #940, opened 2023-08-10
Add PolicyTransportZone resource

wsquan171:
This change adds PolicyTransportZone as resource.

Compared to existing data source, the following changes are made:
- Added `site_path` and `enforcement_point_id` in case a TZ is to be created under non-default site / ep. When left out, the resource will still be CRUD'ed from default / default, or default / `nsxt.enforcement_point` in case of VMC
- `site_path`, `enforcement_point_id` and `transport_type` are made `ForceNew` as they are immutable.
- Added `uplink_teaming_policy_names` which is only supported on `VLAN_BACKED` transport zones.
- When importing, only policy path is accepted.

Fields not added due to missing from current SDK:
- `forwarding_mode`: Choice of `IPV4_ONLY`, `IPV6_ONLY`, `IPV4_AND_IPV6`
- Global manager does not support C/U/D of TZs. Thus only `infra` client is used.

Fields not added due to missing data source and unclear use case:
- `transport_zone_profile_paths`: List of `/infra/transport-zone-profiles`. This option is not even showing up on current UI.

### DIFF
--- a/nsxt/resource_nsxt_policy_transport_zone.go
+++ b/nsxt/resource_nsxt_policy_transport_zone.go
@@ -64,7 +64,7 @@ func resourceNsxtPolicyTransportZone() *schema.Resource {
 				Default:      defaultInfraSitePath,
 				ValidateFunc: validatePolicyPath(),
 			},
-			"enforcement_point_id": {
+			"enforcement_point": {
 				Type:        schema.TypeString,
 				Description: "ID of the enforcement point this Transport Zone belongs to",
 				Optional:    true,
@@ -154,7 +154,7 @@ func policyTransportZoneIDTuple(d *schema.ResourceData, m interface{}) (id, site
 		err = fmt.Errorf("error obtaining Site ID from site path %s", sitePath)
 		return
 	}
-	epID = d.Get("enforcement_point_id").(string)
+	epID = d.Get("enforcement_point").(string)
 	if epID == "" {
 		epID = getPolicyEnforcementPoint(m)
 	}
@@ -173,7 +173,7 @@ func resourceNsxtPolicyTransportZoneCreate(d *schema.ResourceData, m interface{}
 	if siteID == "" {
 		return fmt.Errorf("error obtaining Site ID from site path %s", sitePath)
 	}
-	epID := d.Get("enforcement_point_id").(string)
+	epID := d.Get("enforcement_point").(string)
 	if epID == "" {
 		epID = getPolicyEnforcementPoint(m)
 	}
@@ -217,7 +217,7 @@ func resourceNsxtPolicyTransportZoneRead(d *schema.ResourceData, m interface{}) 
 	}
 
 	d.Set("site_path", sitePath)
-	d.Set("enforcement_point_id", epID)
+	d.Set("enforcement_point", epID)
 	d.Set("display_name", obj.DisplayName)
 	d.Set("description", obj.Description)
 	setPolicyTagsInSchema(d, obj.Tags)
@@ -275,7 +275,7 @@ func resourceNsxtPolicyTransportZoneImporter(d *schema.ResourceData, m interface
 	if err != nil {
 		return nil, err
 	}
-	d.Set("enforcement_point_id", epID)
+	d.Set("enforcement_point", epID)
 	sitePath, err := getSitePathFromChildResourcePath(importID)
 	if err != nil {
 		return rd, err
--- a/nsxt/resource_nsxt_policy_transport_zone.go
+++ b/nsxt/resource_nsxt_policy_transport_zone.go
@@ -4,7 +4,6 @@
 package nsxt
 
 import (
-	"errors"
 	"fmt"
 	"log"
 	"strings"
@@ -16,6 +15,8 @@ import (
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/sites/enforcement_points"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 )
+
+var defaultInfraSitePath = "/infra/sites/default"
 
 func resourceNsxtPolicyTransportZone() *schema.Resource {
 	return &schema.Resource{
@@ -55,12 +56,13 @@ func resourceNsxtPolicyTransportZone() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
-			"site_id": {
-				Type:        schema.TypeString,
-				Description: "ID of the site this Transport Zone belongs to",
-				Optional:    true,
-				ForceNew:    true,
-				Default:     defaultSite,
+			"site_path": {
+				Type:         schema.TypeString,
+				Description:  "Path to the site this Transport Zone belongs to",
+				Optional:     true,
+				ForceNew:     true,
+				Default:      defaultInfraSitePath,
+				ValidateFunc: validatePolicyPath(),
 			},
 			"enforcement_point_id": {
 				Type:        schema.TypeString,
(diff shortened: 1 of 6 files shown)


nsxt/resource_nsxt_policy_transport_zone.go
```diff
@@ -0,0 +1,289 @@
...
+
+func policyTransportZoneIDTuple(d *schema.ResourceData, m interface{}) (id, siteID, epID string, err error) {
+	id = d.Id()
+	if id == "" {
+		err = fmt.Errorf("Error obtaining PolicyTransportZone ID")
+		return
+	}
+	siteID = d.Get("site_id").(string)
+	if siteID == "" {
+		siteID = defaultSite
```
annakhm, line 139: is there a reason not to specify this default in the schema instead?

wsquan171: Sure. Will make it default in schema.

nsxt/resource_nsxt_policy_transport_zone.go
```diff
@@ -0,0 +1,289 @@
...
+	d.Set("enforcement_point_id", epID)
+	d.Set("display_name", obj.DisplayName)
+	d.Set("description", obj.Description)
+	setPolicyTagsInSchema(d, obj.Tags)
+	d.Set("nsx_id", id)
+	d.Set("path", obj.Path)
+	d.Set("revision", obj.Revision)
+	d.Set("is_default", obj.IsDefault)
+	d.Set("transport_type", obj.TzType)
+	d.Set("UplinkTeamingPolicyNames", obj.UplinkTeamingPolicyNames)
```
annakhm, line 208: `UplinkTeamingPolicyNames` => `uplink_teaming_policy_names`

nsxt/resource_nsxt_policy_transport_zone.go
```diff
@@ -0,0 +1,289 @@
...
+			"display_name": getDataSourceDisplayNameSchema(),
+			"description":  getDataSourceDescriptionSchema(),
+			"path":         getPathSchema(),
+			"revision":     getRevisionSchema(),
+			"tag":          getTagsSchema(),
+			"is_default": {
+				Type:        schema.TypeBool,
+				Description: "Indicates whether the transport zone is default",
+				Optional:    true,
+				Computed:    true,
```
annakhm, line 41: Does the API specify default here?

nsxt/resource_nsxt_policy_transport_zone.go
annakhm: Why does this need to be computed, in case default is set whenever the user does not specify the value? Does platform ever override this value?

ksamoray:
Code LGTM, no doc or test?

nsxt/resource_nsxt_policy_transport_zone.go
```diff
@@ -0,0 +1,291 @@
...
+	displayName := d.Get("display_name").(string)
+	description := d.Get("description").(string)
+	tags := getPolicyTagsFromSchema(d)
+	isDefault := d.Get("is_default").(bool)
+	transportType := d.Get("transport_type").(string)
+	uplinkTeamingNames := getStringListFromSchemaList(d, "uplink_teaming_policy_names")
+
+	if len(uplinkTeamingNames) > 0 && transportType != model.PolicyTransportZone_TZ_TYPE_VLAN_BACKED {
+		// uplink_teaming_policy_names only valid for VLAN_BACKED TZ
+		return fmt.Errorf("Cannot use uplink_teaming_policy_names with transport_type %s", transportType)
```
ksamoray, line 113: nit: it is recommended not to capitalize error messages as these could be concatenated later on.

nsxt/resource_nsxt_policy_transport_zone.go
annakhm: Yes, I don't think enforcement point has a default

nsxt/resource_nsxt_policy_transport_zone.go
```diff
@@ -0,0 +1,291 @@
...
+			},
+			"uplink_teaming_policy_names": {
+				Type:        schema.TypeList,
+				Description: "Names of the switching uplink teaming policies that are supported by this transport zone.",
+				Optional:    true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+			"site_id": {
```
annakhm, line 58: Lets consider changing this to `site_path` for consistency (we already have `site_path` in TZ data source and other resources)

nsxt/resource_nsxt_policy_transport_zone.go
```diff
@@ -0,0 +1,291 @@
...
+
+func resourceNsxtPolicyTransportZoneCreate(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+
+	id := d.Get("nsx_id").(string)
+	if id == "" {
+		id = newUUID()
+	}
+	siteID := d.Get("site_id").(string)
+	if siteID == "" {
```
annakhm, line 157: This is no longer needed if default is specified in schema

nsxt/resource_nsxt_policy_transport_zone.go
```diff
@@ -0,0 +1,291 @@
...
+		if err != nil {
+			return nil, err
+		}
+		d.Set("site_id", siteID)
+		return rd, nil
+	} else if !errors.Is(err, ErrNotAPolicyPath) {
+		return rd, err
+	}
+
+	var siteID, epID, id string
```
annakhm, line 267: I believe for all new nested resources we should only support import by policy path. @ksamoray what do you think?

ksamoray: Makes sense. It won't take any functionality from the user, and will keep things tidy on our side.

wsquan171: Removed the all other import forms than via policy ID

nsxt/resource_nsxt_policy_transport_zone.go
wsquan171: done

nsxt/resource_nsxt_policy_transport_zone.go
wsquan171: Fixed all err msg in this file.

website/docs/r/policy_transport_zone.html.markdown
```diff
@@ -0,0 +1,67 @@
...
+# nsxt_policy_transport_zone
+
+This resource provides a method for the management of Policy based Transport Zones (TZ). A Transport Zone defines the scope to which a network can extend in NSX. For example an overlay based Transport Zone is associated with both hypervisors and logical switches and defines which hypervisors will be able to serve the defined logical switch. Virtual machines on the hypervisor associated with a Transport Zone can be attached to logical switches in that same Transport Zone.
+
+This data source is applicable to NSX Policy Manager.
+
+## Example Usage
+
+```hcl
+data "nsxt_policy_transport_zone" "overlay_transport_zone" {
```
annakhm, line 17: those should be resources, not data sources

wsquan171: fixed

website/docs/r/policy_transport_zone.html.markdown
```diff
@@ -0,0 +1,67 @@
+---
+subcategory: "Fabric"
+layout: "nsxt"
+page_title: "NSXT: nsxt_policy_transport_zone"
+description: A resource to configure Policy Transport Zone.
+---
+
+# nsxt_policy_transport_zone
+
+This resource provides a method for the management of Policy based Transport Zones (TZ). A Transport Zone defines the scope to which a network can extend in NSX. For example an overlay based Transport Zone is associated with both hypervisors and logical switches and defines which hypervisors will be able to serve the defined logical switch. Virtual machines on the hypervisor associated with a Transport Zone can be attached to logical switches in that same Transport Zone.
```
annakhm, line 10: I think we should use policy terminology here (segment rather than logical switch)

nsxt/resource_nsxt_policy_transport_zone.go
```diff
@@ -0,0 +1,285 @@
...
+			},
+			"site_path": {
+				Type:         schema.TypeString,
+				Description:  "Path to the site this Transport Zone belongs to",
+				Optional:     true,
+				ForceNew:     true,
+				Default:      defaultInfraSitePath,
+				ValidateFunc: validatePolicyPath(),
+			},
+			"enforcement_point_id": {
```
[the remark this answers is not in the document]
wsquan171: done

website/docs/r/policy_transport_zone.html.markdown
wsquan171: done

PR #946, opened 2023-08-21
Bump website linter action golang to v1.20

ksamoray:
This is needed as `make tools` fails due to change in requirements.

See example failure [here](https://github.com/vmware/terraform-provider-nsxt/actions/runs/5927217813/job/16070272629?pr=937)


ksamoray:
Fixed by PR #947

PR #948, opened 2023-08-21
Support on-demand connection init in the provider

annakhm:
When the new `on_demand_connection` provider flag is set to true, provider will not verify connection and pull NSX version on provider startup. Instead, those operations will be done for each provider instance when terraform objects are evaluated.
This setting is useful when NSX appliance is being deployed via same apply process, and not yet responsive when the provider is initialized. This setting is not supported for VMC and not compatible with deprecated objects implemented via the old SDK. It is also temporarily incompatible with license feature (this will be fixed in follow up PR)

### DIFF
--- a/nsxt/provider.go
+++ b/nsxt/provider.go
@@ -210,6 +210,12 @@ func Provider() *schema.Provider {
 				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("NSXT_CA", nil),
 			},
+			"on_demand_connection": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Avoid initializing NSX connection on startup",
+				DefaultFunc: schema.EnvDefaultFunc("NSXT_ON_DEMAND_CONNECTION", false),
+			},
 		},
 
 		DataSourcesMap: map[string]*schema.Resource{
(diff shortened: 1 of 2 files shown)


ksamoray (approved):

nsxt/provider.go
```diff
@@ -407,13 +413,19 @@ func Provider() *schema.Provider {
...
 func configureNsxtClient(d *schema.ResourceData, clients *nsxtClients) error {
+	onDemandConn := d.Get("on_demand_connection").(bool)
 	clientAuthCertFile := d.Get("client_auth_cert_file").(string)
 	clientAuthKeyFile := d.Get("client_auth_key_file").(string)
 	clientAuthCert := d.Get("client_auth_cert").(string)
 	clientAuthKey := d.Get("client_auth_key").(string)
 	vmcToken := d.Get("vmc_token").(string)
 	vmcAuthMode := d.Get("vmc_auth_mode").(string)
 
+	if onDemandConn {
```
wsquan171, line 424: Does it make more sense if we make vmc params taking precedence over this `on_demand_connection`, in that if vmc tokens / mode is set, `on_demand_connection` is ignored and turned off?

annakhm: I think its preferrable that user is aware the setting is not applicable, and corrects the configuration

wsquan171 (approved):

annakhm:
> @annakhm should we run a test cycle (e.g sanity) with this turned on?

I ran the tests locally. Indeed we should add an option to pass env variables to CI.

Issue #945, opened 2023-08-18
nsxt_nat_rule broken from NSX-T 4.0.1 onward


martinrohrbach:
### Describe the bug

From NSX-T 4.0.0 to 4.0.1, the "nat_pass" parameter for the /api/v1/logical-routers/<logical-router-id>/nat/rules/<rule-id> has seemingly been removed silently (at least I could not find it in the release notes, but I asked the support as well).

This essentially breaks the nsxt_nat_rule ressource as it defaults that parameter to true and the API rejects the corresponding call(s) for creation and/or update. 

`Error: Error during NatRule update: Status: 400 Bad Request, Body: {
  "httpStatus" : "BAD_REQUEST",
  "error_code" : 287,
  "module_name" : "common-services",
  "error_message" " "Json deserialzation error: property nat_pass is unrecognized."
}`

We're using the NSX-T provider 3.2.5, however, the code / behaviour hasn't changed in newer versions as far as I can see.  

### Reproduction steps

1. Create a nsxt_nat_rule resource and apply to NSX-T 4.0.1 or higher

### Expected behavior

1. The NAT rule is created without error

### Additional context

This also applies to existing NAT rules using this resource, after an update to 4.1.x we are unable to manage them and in fact are unable to use Terraform at all due to the NSX-T provider not working anymore. We've yet to find a workaround other than simply removing the resource type from the TF code as well as the state and managing the NAT rules manually. 

Moving to policy API is currently not an option because we need to create those rules for a manager API router which cannot be migrated to policy API due to other product limitations.

Would it be possible to leave the provider API as is but change the handling of the parameter "nat_pass" to transparently use "firewall_match" instead (as suggested in the API) for newer NSX-T versions?
https://vdc-repo.vmware.com/vmwb-repository/dcr-public/4e0249dc-f050-4e5f-8a47-bb247d9fa31f/96084528-a9f7-4e2b-866d-c5bbc1b090f7/api_includes/method_GetNatRule.html

We've also created a service request to investigate why this was changed silently (or if we simply missed this) and/or if this can be considered a bug to be reverted at some point (unlikely as it is).

martinrohrbach:
@ksamoray @annakhm Very much appreciated, will test this as soon as a new provider version is released.

PR #950, opened 2023-08-24
Update nsxt_nat_rule with firewall_match

ksamoray:
firewall_match attribute replace nat_pass in nsxt_nat_rule resource.

Fixes: #945

### DIFF
--- a/nsxt/resource_nsxt_nat_rule.go
+++ b/nsxt/resource_nsxt_nat_rule.go
@@ -5,23 +5,21 @@ package nsxt
 
 import (
 	"fmt"
-	"log"
-	"net/http"
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
-	"github.com/vmware/go-vmware-nsxt/manager"
-	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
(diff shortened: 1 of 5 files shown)


nsxt/resource_nsxt_nat_rule.go
```diff
@@ -245,41 +239,43 @@ func resourceNsxtNatRuleUpdate(d *schema.ResourceData, m interface{}) error {
 	//match_service := d.Get("match_service").(*NsServiceElement)
 	matchSourceNetwork := d.Get("match_source_network").(string)
 	natPass := d.Get("nat_pass").(bool)
+	firewallMatch := model.NatRule_FIREWALL_MATCH_MATCH_INTERNAL_ADDRESS
```
annakhm, line 242: do we know what version this was added in?

annakhm (commented):
LGTM

nsxt/resource_nsxt_nat_rule.go
ksamoray: nat_pass was removed at 4.0.0
firewall_match exists in 2.4 - it seemed old enough so I didn't check earlier versions...

ksamoray: Maybe I should keep the code which uses the old API for pre-4.0.0 environments?

annakhm: I think current code should be good, I would run the test against 3.0.0 to be sure

annakhm (approved):

PR #954, opened 2023-08-27
Policy host transport node profile resource

ksamoray:


nsxt/resource_nsxt_policy_host_transport_node_profile.go
```diff
@@ -0,0 +1,161 @@
...
+		Schema: map[string]*schema.Schema{
+			"nsx_id":       getNsxIDSchema(),
+			"path":         getPathSchema(),
+			"display_name": getDisplayNameSchema(),
+			"description":  getDescriptionSchema(),
+			"revision":     getRevisionSchema(),
+			"tag":          getTagsSchema(),
+			// host_switch_spec
+			"standard_host_switch":      getStandardHostSwitchSchema(),
+			"preconfigured_host_switch": getPreconfiguredHostSwitchSchema(),
```
annakhm, line 32: looks like this attr is ignored later on?

annakhm:
Did I understand correctly that in transport node resource user can either provide equivalent config, or define this profile as separate resource and specify it in transport node?

nsxt/resource_nsxt_policy_host_transport_node_profile.go
ksamoray: `getHostSwitchSpecFromSchema()` and `setHostSwitchSpecInSchema()` handle these two attributes (it's the same schema which is used by transport_node).

ksamoray:
> Did I understand correctly that in transport node resource user can either provide equivalent config, or define this profile as separate resource and specify it in transport node?

The simpler way to configure the host nodes (e.g ESXs) is to associate a transport_node_profile with the compute manager, and from there NSX will configure the nodes automatically.

annakhm (approved):

website/docs/r/policy_host_transport_node_profile.html.markdown
```diff
@@ -0,0 +1,120 @@
...
+            * `default_gateway` - (Required) Gateway IP.
+        * `static_ip_mac` - (Optional) IP and MAC assignment specification for Static IP List.
+            * `default_gateway` - (Required) Gateway IP.
+            * `subnet_mask` - (Required) Subnet mask.
+            * `ip_mac_pair` - (Required) List of IPs and MACs for transport node host switch virtual tunnel endpoints.
+                * `ip` - (Required) IP address.
+                * `mac` - (Required) MAC address.
+                * `static_ip_pool_id` - (Optional) IP assignment specification for Static IP Pool.
+    * `is_migrate_pnics` - (Optional) Migrate any pnics which are in use.
+    * `pnics` - (Optional) Physical NICs connected to the host switch.
```
annakhm, line 67: This should be `pnic`

website/docs/r/policy_host_transport_node_profile.html.markdown
```diff
@@ -0,0 +1,120 @@
...
+    host_switch_type = "NVDS"
+    ip_assignment {
+      assigned_by_dhcp = true
+    }
+    transport_zone_endpoint {
+      transport_zone_id = data.nsxt_transport_zone.tz1.id
+    }
+    host_switch_profile_id = [nsxt_uplink_host_switch_profile.hsw_profile1.id]
+    is_migrate_pnics       = false
+    pnics {
```
annakhm, line 29: `pnic`

website/docs/r/policy_host_transport_node_profile.html.markdown
```diff
@@ -0,0 +1,120 @@
...
+* `standard_host_switch` - (Optional) Standard host switch specification.
+    * `cpu_config` - (Optional) Enhanced Networking Stack enabled HostSwitch CPU configuration.
+        * `num_lcores` - (Required) Number of Logical cpu cores (Lcores) to be placed on a specified NUMA node.
+        * `numa_node_index` - (Required) Unique index of the Non Uniform Memory Access (NUMA) node.
+    * `host_switch_id` - (Optional) The host switch id. This ID will be used to reference a host switch.
+    * `host_switch_mode` - (Optional) Operational mode of a HostSwitch. Accepted values - 'STANDARD', 'ENS', 'ENS_INTERRUPT' or 'LEGACY'. The default value is 'STANDARD'.
+    * `host_switch_profile_id` - (Optional) Identifiers of host switch profiles to be associated with this host switch.
+    * `host_switch_type` - (Optional) Type of HostSwitch. Accepted values - 'NVDS' or 'VDS'. The default value is 'NVDS'.
+    * `ip_assignment` - (Required) - Specification for IPs to be used with host switch virtual tunnel endpoints. Should contain exatly one of the below:
+        * `assigned_by_dhcp` - (Optional) Enables DHCP assignment. Should be set to true.
```
annakhm, line 54: I don't get the `Should be set to true` part :)

ksamoray: [IpAssignmentSpec](https://vdc-download.vmware.com/vmwb-repository/dcr-public/201a09cb-5610-49e6-9d31-7db0920e3a65/65f199d6-8516-4092-9b19-906559f972b5/api_includes/types_IpAssignmentSpec.html) is a polymorphic schema with StaticIpListSpec, StaticIpPoolSpec containing members. Instead [AssignedByDhcp](https://vdc-download.vmware.com/vmwb-repository/dcr-public/201a09cb-5610-49e6-9d31-7db0920e3a65/65f199d6-8516-4092-9b19-906559f972b5/api_includes/types_AssignedByDhcp.html) has no members - it's a struct which has only resource_type. As we can't have a schema without any attributes in TF, I added a boolean. I'd appreciate a better suggestion really as it does look weird.

website/docs/r/policy_host_transport_node_profile.html.markdown
```diff
@@ -0,0 +1,120 @@
...
+* `ignore_overridden_hosts` - (Optional) Determines if cluster-level configuration should be applied on overridden hosts
+* `standard_host_switch` - (Optional) Standard host switch specification.
+    * `cpu_config` - (Optional) Enhanced Networking Stack enabled HostSwitch CPU configuration.
+        * `num_lcores` - (Required) Number of Logical cpu cores (Lcores) to be placed on a specified NUMA node.
+        * `numa_node_index` - (Required) Unique index of the Non Uniform Memory Access (NUMA) node.
+    * `host_switch_id` - (Optional) The host switch id. This ID will be used to reference a host switch.
+    * `host_switch_mode` - (Optional) Operational mode of a HostSwitch. Accepted values - 'STANDARD', 'ENS', 'ENS_INTERRUPT' or 'LEGACY'. The default value is 'STANDARD'.
+    * `host_switch_profile_id` - (Optional) Identifiers of host switch profiles to be associated with this host switch.
+    * `host_switch_type` - (Optional) Type of HostSwitch. Accepted values - 'NVDS' or 'VDS'. The default value is 'NVDS'.
+    * `ip_assignment` - (Required) - Specification for IPs to be used with host switch virtual tunnel endpoints. Should contain exatly one of the below:
```
annakhm, line 53: typo in `exatly`

website/docs/r/policy_host_transport_node_profile.html.markdown
annakhm: From terraform perspective polymorphic structure is hidden, but the flag should be only set to true if user wants `dhcp` feature. I would suggest to add `false` default for this flag, and explain in docs, that `static_ip` and `static_ip_mac` can only be specified if `dhcp` is set to false.

ksamoray: Cool I'll do this in a separate PR as it affects transport node as well.

ksamoray: Only one of assigned_by_dhcp/static_ip/static_mac/static_ip_pool can be used. It's enforced [here](https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/resource_nsxt_transport_node.go#L1360) as I couldn't use TF ExactlyOneOf capability. That's also documented.
So if I set a false default value - it'll always add assigned_by_dhcp which will create a mess IMO.
I think that I'll leave it as is - if it is set to `false` the transaction fails (as some IP assignment should be chosen and if assigned_by_dhcp is false, then there's no other assignment enabled).
And remove "should be set to true" from doc.

PR #955, opened 2023-08-27
Implement compute manager data source

ksamoray:


annakhm:
Search API not catching this object type?

ksamoray:
> Search API not catching this object type?

@annakhm  tried with search API, it didn't retrieve anything

annakhm (approved):

PR #957, opened 2023-08-30
Failure Domain support

GraysonWu:


nsxt/data_source_nsxt_failure_domain.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+	}
+
+	d.SetId(*obj.Id)
+	d.Set("display_name", obj.DisplayName)
+	d.Set("description", obj.Description)
+
+	preferPtr := obj.PreferredActiveEdgeServices
+	preferStr := ""
+	if preferPtr != nil {
```
ksamoray, line 90: Same here.

ksamoray (changes requested):

nsxt/data_source_nsxt_failure_domain.go
```diff
@@ -0,0 +1,99 @@
...
+func dataSourceNsxtFailureDomain() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNsxtFailureDomainRead,
+
+		Schema: map[string]*schema.Schema{
+			"id":           getDataSourceIDSchema(),
+			"display_name": getDataSourceDisplayNameSchema(),
+			"description":  getDataSourceDescriptionSchema(),
+			"preferred_active_edge_services": {
+				Type:        schema.TypeString,
```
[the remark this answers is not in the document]
GraysonWu: As the definition of `preferred_active_edge_services` [here](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/612caf56-5bb1-4a0b-8d2b-12ccd8981904/37be75ba-aa56-42f4-99c3-f6c989ddbe57/api_includes/types_FailureDomain.html ). `true`, `false` and `unset`, all three are valid cases. nil pointer in the SDK represents the `unset` state. So I use `TypeString` instead of `TypeBool` here.

annakhm:
I'm a bit confused on this one. Failure domain is used in transport node creation (perhaps somewhere else in future?), and it can be unspecified. If unspecified, isn't the behavior the same than using the failure domain with an `unset` flag value. 
If this is the case, perhaps we can revert to boolean flag without loss of functionality? Or do we still want to stick to the API?

GraysonWu:
> I'm a bit confused on this one. Failure domain is used in transport node creation (perhaps somewhere else in future?), and it can be unspecified. If unspecified, isn't the behavior the same than using the failure domain with an `unset` flag value. If this is the case, perhaps we can revert to boolean flag without loss of functionality? Or do we still want to stick to the API?

Sorry for the confusion. `unspecified` and `unset` are the same thing representing a nil pointer in the SDK. Do you think change the option strings to [`active`, `standby`, `no_preference`] and set the default value to `no_preference` can avoid the confusion?

nsxt/resource_nsxt_failure_domain.go
```diff
@@ -0,0 +1,154 @@
...
+	"log"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/model"
+)
+
+var activeEdgeServices = []string{
+	"true",
```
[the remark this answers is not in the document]
GraysonWu: Done.

website/docs/r/failure_domain.html.markdown
```diff
@@ -0,0 +1,55 @@
...
+
+This resource is supported with NSX 3.0.0 onwards.
+
+## Example Usage
+
+```hcl
+data "nsxt_failure_domain" "failure_domain" {
+  display_name            = "failuredomain1"
+  description             = "failuredomain"
+  preferred_edge_services = "true"
```
ksamoray, line 20: active

nsxt/data_source_nsxt_failure_domain.go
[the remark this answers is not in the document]
ksamoray: Understood. But shouldn't this be computed instead of optional?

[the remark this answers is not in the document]
GraysonWu: I think it's more optional than computed. Basically, it takes what users provide and goes with no preference, nil on API side, if users don't set this field.

website/docs/r/failure_domain.html.markdown
GraysonWu: Thanks. Done.

nsxt/data_source_nsxt_failure_domain.go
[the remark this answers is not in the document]
ksamoray: But this is a data source - so each field is either a search argument (and then required or optional) or an output (and then computed - as it's only set by TF). Maybe I'm missing something (sorry I'm new to TF...).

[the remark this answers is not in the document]
annakhm: I don't think this field is needed in the data source, but if present, it should be indeed Computed

nsxt/resource_nsxt_failure_domain.go
```diff
@@ -0,0 +1,156 @@
...
+	failureDomain := failureDomainSchemaToModel(d)
+	obj, err := client.Create(failureDomain)
+	if err != nil {
+		id := ""
+		if obj.Id != nil {
+			id = *obj.Id
+		}
+		return handleCreateError("Failure Domain", id, err)
+	}
+	log.Printf("[INFO] Creating Failure Domain with ID %s", *obj.Id)
```
annakhm, line 117: This log should probably move to location before API call, for consistency

website/docs/r/failure_domain.html.markdown
```diff
@@ -0,0 +1,55 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `display_name` - (Required) Display name of the resource.
+* `description` - (Optional) Description of the resource.
+* `preferred_edge_services` - Set preference for failure domain. Set preference for edge transport node failure domain which will be considered while doing auto placement of logical router, DHCP and MDProxy on edge node.
+  `active`: For preemptive failover mode, active edge cluster member allocation prefers this failure domain.
```
annakhm, line 35: Lets put those values in same line as general description, for sake of consistency. New line with offset generally describes nested attributes rather than enumeration values.

nsxt/data_source_nsxt_failure_domain.go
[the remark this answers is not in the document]
GraysonWu: My bad. I didn't notice we were discussing the data source instead of the resource. 
Yes, makes sense. Removed this field.

website/docs/r/failure_domain.html.markdown
GraysonWu: Thanks. Done.

nsxt/resource_nsxt_failure_domain.go
GraysonWu: The reason why I put it here is the ID can only be known after the API call, like the edge cluster resource. Let me know if you think moving it prior to the API call with `description_name` as a parameter is a better way.

annakhm: Makes sense, lets rename `Creating` to `Created` then

annakhm:
> > I'm a bit confused on this one. Failure domain is used in transport node creation (perhaps somewhere else in future?), and it can be unspecified. If unspecified, isn't the behavior the same than using the failure domain with an `unset` flag value. If this is the case, perhaps we can revert to boolean flag without loss of functionality? Or do we still want to stick to the API?
> 
> Sorry for the confusion. `unspecified` and `unset` are the same thing representing a nil pointer in the SDK. Do you think change the option strings to [`active`, `standby`, `no_preference`] and set the default value to `no_preference` can avoid the confusion?

Yes, thank you, I think the list [`active`, `standby`, `no_preference`] is more user friendly.
However let me rephrase my question: 
Lets say we have a resource that references failure domain we create using this resource:
```
{ 
  ..
  failure_domain_id = nsxt_failure_domain.test.id
}
```

If we don't specify `failure_domain_id`, is it any different, in terms of platform behavior, then specifying failure_domain with `no_preference`?

GraysonWu:
> > > I'm a bit confused on this one. Failure domain is used in transport node creation (perhaps somewhere else in future?), and it can be unspecified. If unspecified, isn't the behavior the same than using the failure domain with an `unset` flag value. If this is the case, perhaps we can revert to boolean flag without loss of functionality? Or do we still want to stick to the API?
> > 
> > 
> > Sorry for the confusion. `unspecified` and `unset` are the same thing representing a nil pointer in the SDK. Do you think change the option strings to [`active`, `standby`, `no_preference`] and set the default value to `no_preference` can avoid the confusion?
> 
> Yes, thank you, I think the list [`active`, `standby`, `no_preference`] is more user friendly. However let me rephrase my question: Lets say we have other resource that references the `failure_domain` resource:
> 
> ```
> { 
>   ..
>   failure_domain_id = nsxt_failure_domain.test.id
> }
> ```
> 
> If we don't specify `failure_domain_id`, is it any different, in terms of platform behavior, then specifying failure_domain with `no_preference`?

Ah, my bad now I understand your question. 
That's a good question. Based on my understanding, we can revert to the bool flag. 
I'm not sure if there are other resources, except `transport_node`, referencing `failure_domain`. According to the [API doc of `transport_node`](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/612caf56-5bb1-4a0b-8d2b-12ccd8981904/37be75ba-aa56-42f4-99c3-f6c989ddbe57/api_includes/types_TransportNode.html):
```
 In case failure domain is not
set by user explicitly, it will be always assigned with default
system created failure domain.
```
As for the default failure domain, I only found the description in [NSX Java SDK doc](https://vmware.github.io/vsphere-automation-sdk-java/nsx/nsx/com/vmware/nsx/model/FailureDomain.html):
```
You can't update preferred_active_edge_services flag for system generated default failure domain. It will be unset which means that failure domain can be used for both active and standby allocation.
```
Do you see any risks? @annakhm

nsxt/resource_nsxt_failure_domain.go
GraysonWu: Done.

nsxt/resource_nsxt_failure_domain.go
```diff
@@ -0,0 +1,156 @@
...
+
+	id := d.Id()
+	if id == "" {
+		return fmt.Errorf("Error obtaining FailureDomain ID")
+	}
+
+	client := nsx.NewFailureDomainsClient(connector)
+	obj, err := client.Get(id)
+	if err != nil {
+		return fmt.Errorf("error during Failure Domain read: %v", err)
```
ksamoray, line 59: Some errors are capitalized while others aren't - in the provider project there's no standard for this but let's be consistent at least at the PR level (I'd rather non-capitalized but as you want).

GraysonWu: Sounds good. Done.

annakhm:
> What do you think? @annakhm
Good point about importing a domain with `no_preference`. I would suggest leaving the PR as is.

annakhm (approved):

PR #958, opened 2023-08-30
Failure Domain in Edge Cluster

GraysonWu:
This PR should wait for PR #957 to be merged.

### DIFF
--- a/nsxt/resource_nsxt_edge_cluster.go
+++ b/nsxt/resource_nsxt_edge_cluster.go
@@ -8,9 +8,17 @@ import (
 	"log"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/model"
 )
+
+var failureDomainAllocationOptions = []string{
+	"enable",
+	"disable",
+}
 
 func resourceNsxtEdgeCluster() *schema.Resource {
 	return &schema.Resource{
@@ -95,6 +103,12 @@ func resourceNsxtEdgeCluster() *schema.Resource {
 					},
 				},
 			},
+			"failure_domain_allocation": {
+				Type:         schema.TypeString,
+				Description:  "Flag to enable/disable failure domain based allocation",
+				Optional:     true,
+				ValidateFunc: validation.StringInSlice(failureDomainAllocationOptions, false),
+			},
 		},
 	}
 }
@@ -107,14 +121,16 @@ func resourceNsxtEdgeClusterCreate(d *schema.ResourceData, m interface{}) error 
 	displayName := d.Get("display_name").(string)
 	tags := getMPTagsFromSchema(d)
 	clusterProfileBindings := getClusterProfileBindingsFromSchema(d)
-
 	members := getEdgeClusterMembersFromSchema(d)
+	allocationRules := getAllocationRulesFromSchema(d)
+
 	obj := model.EdgeCluster{
 		Description:            &description,
 		DisplayName:            &displayName,
 		Tags:                   tags,
 		ClusterProfileBindings: clusterProfileBindings,
 		Members:                members,
+		AllocationRules:        allocationRules,
 	}
 
 	log.Printf("[INFO] Creating Edge Cluster with name %s", displayName)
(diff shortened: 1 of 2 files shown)


nsxt/resource_nsxt_edge_cluster.go
```diff
@@ -216,6 +254,27 @@ func setNodeRtepIPsInSchema(d *schema.ResourceData, nodeRtepIPs []model.NodeRtep
...
+	for _, allocationRule := range allocationRules {
+		structValue := allocationRule.Action
+		actionInterface, errs := converter.ConvertToGolang(structValue, model.AllocationBasedOnFailureDomainBindingType())
+		if errs != nil {
+			continue
+		}
+		action := actionInterface.(model.AllocationBasedOnFailureDomain)
+		if action.Enabled != nil {
+			if *action.Enabled {
+				d.Set("failuer_domain_allocation", "true")
```
annakhm, line 268: typo here and below

nsxt/resource_nsxt_edge_cluster.go
```diff
@@ -8,10 +8,18 @@ import (
...
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/model"
 )
 
+var enableFailureDomainAllocation = []string{
+	"true",
```
annakhm, line 19: If we have three options here rather than two, I would suggest `enable`, `disable` and `no_preference` options here as well (and rename attr to `failure_domain_allocation`)

GraysonWu: Discussion in https://github.com/vmware/terraform-provider-nsxt/pull/957#issuecomment-1718370583 makes me uncertain if we need three options here. 
I didn't find any doc mentioning what allocation rule and behavior will be used if users don't provide one. If `no_preference` and `disable` are the same then we could use a boolean here.

nsxt/resource_nsxt_edge_cluster.go
GraysonWu: Thanks. Done.

nsxt/resource_nsxt_edge_cluster.go
```diff
@@ -95,6 +103,12 @@ func resourceNsxtEdgeCluster() *schema.Resource {
 					},
 				},
 			},
+			"failure_domain_allocation": {
+				Type:         schema.TypeString,
```
annakhm, line 107: If we only have 2 options, why not use `TypeBool`?

GraysonWu: Actually, we have three options: 1. enable 2. disable 3. unset. In API, there is a list to provide the allocation rules, which could be empty or contain a rule enabling failure domain allocation or a rule disabling failure domain allocation. It's too complicated for users to just enable the failure domain, so we simplify it in our terraform schema here. 

I'm not sure what the difference is between an empty list and a disabling failure domain rule in the actual behavior. The reason I use string here is similar to this discussion https://github.com/vmware/terraform-provider-nsxt/pull/957#issuecomment-1718370583  if we use bool here, then when a user imports an edge cluster without failure domain preference, we will automatically add a disable failure domain allocation rule inside.

nsxt/resource_nsxt_edge_cluster.go
salv-orlando: While @GraysonWu is correct, using a boolean might create a problem in case of future API changes where `no_preference` will end up having different semantics than `disable`

annakhm: I don't think behavior changes are a big concern, since this API will be replaced by policy API in future

nsxt/resource_nsxt_edge_cluster.go
annakhm: If we're unsure about platform behavior, lets keep the three options, but rename them in a way that would not cause confusion regarding property type (its not intuitive for terraform user to wrap `true` and `false` values in quotes)

GraysonWu: Make sense. Done.

website/docs/r/edge_cluster.html.markdown
```diff
@@ -31,6 +31,7 @@ The following arguments are supported:
 * `description` - (Optional) Description of the resource.
 * `tag` - (Optional) A list of scope + tag pairs to associate with this resource.
 * `edge_ha_profile_id` - (Optional) Edge high availability cluster profile ID.
+* `failure_domain_allocation` - (Optional) Flag to enable failure domain based allocation. Enable placement algorithm to consider failure domain of edge transport nodes and place active and standby contexts in different failure domains.
```
annakhm, line 34: We need to provide supported values here, otherwise the user might think its hcl bool

GraysonWu: Done.

annakhm (approved):

PR #959, opened 2023-09-02
Add realized_id attribute to policy fabric objects

annakhm:
In addition, remove readonly field from write flow in edge cluster resource.


ksamoray:
We need this also in nsxt_policy_ip_pool. Maybe both data source and resource as well.

annakhm:
`nsxt_policy_ip_pool` is a regular policy object, that is realized asyncronously, it wouldn't have `realized_id` on the main API endpoint. We need to pull it from realization API (which is what the realization data source does).
Instead, fabric resources seem to be realized synchronously and can return the realized id right away.

ksamoray (approved):

PR #960, opened 2023-09-02
Rewrite the license feature using new SDK

annakhm:
Remove restriction of on_demand_connection setting with license feature.


nsxt/provider.go
```diff
@@ -42,6 +49,7 @@ type commonProviderConfig struct {
 	RetryStatusCodes       []int
 	Username               string
 	Password               string
+	LicenseDiff            listDiff
```
ksamoray, line 52: Why persisting here the diffs instead of the current license list? Not sure I understand how this works.

annakhm: We want to apply/remove only the changes in license, not the whole list. So we can either persist the old+new license list, or just persist the changes

ksamoray (approved):

PR #965, opened 2023-09-07
Implement host transport node collection resource

ksamoray:


nsxt/resource_nsxt_policy_host_transport_node_collection.go
```diff
@@ -0,0 +1,277 @@
...
+							Required:    true,
+							Description: "sub-cluster Id",
+						},
+					},
+				},
+			},
+			"transport_node_profile_id": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Transport Node Profile Path",
```
annakhm, line 90: is this path or id?

ksamoray: This is a path - as this is a policy resource, it requires paths as object references. I followed the API naming where this attribute is called `transport_node_profile_id` but maybe it makes more sense to name it `transport_node_profile_path`.

website/docs/r/policy_host_transport_node_collection.html.markdown
```diff
@@ -0,0 +1,62 @@
...
+data "vsphere_compute_cluster" "compute_cluster" {
+  provider      = vsphere.vsphere-nsxcfg-data
+  name          = "Compute-Cluster"
+  datacenter_id = data.vsphere_datacenter.datacenter.id
+}
+
+resource "nsxt_policy_host_transport_node_collection" "htnc1" {
+  display_name                = "HostTransportNodeCollection1"
+  compute_collection_id       = data.vsphere_compute_cluster.compute_cluster.id
+  transport_node_profile_path = nsxt_policy_host_transport_node_profile.tnp.id
```
annakhm, line 24: shouldn't this be `nsxt_policy_host_transport_node_profile.tnp.path`?

nsxt/resource_nsxt_policy_host_transport_node_collection.go
```diff
@@ -0,0 +1,277 @@
...
+		Schema: map[string]*schema.Schema{
+			"nsx_id":       getNsxIDSchema(),
+			"path":         getPathSchema(),
+			"display_name": getDisplayNameSchema(),
+			"description":  getDescriptionSchema(),
+			"revision":     getRevisionSchema(),
+			"tag":          getTagsSchema(),
+			"site_path": {
+				Type:         schema.TypeString,
+				Description:  "Path to the site this Transport Zone belongs to",
```
annakhm, line 36: `Transport Zone` => `resource` ?

nsxt/resource_nsxt_policy_host_transport_node_collection.go
```diff
@@ -0,0 +1,277 @@
...
+				Optional:     true,
+				ForceNew:     true,
+				Default:      defaultInfraSitePath,
+				ValidateFunc: validatePolicyPath(),
+			},
+			"enforcement_point": {
+				Type:        schema.TypeString,
+				Description: "ID of the enforcement point this Transport Zone belongs to",
```
annakhm, line 44: Transport Zone => resource ?

nsxt/resource_nsxt_policy_host_transport_node_collection.go
```diff
@@ -0,0 +1,277 @@
...
+)
+
+func resourceNsxtPolicyHostTransportNodeCollection() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNsxtPolicyHostTransportNodeCollectionCreate,
+		Read:   resourceNsxtPolicyHostTransportNodeCollectionRead,
+		Update: resourceNsxtPolicyHostTransportNodeCollectionUpdate,
+		Delete: resourceNsxtPolicyHostTransportNodeCollectionDelete,
+		Importer: &schema.ResourceImporter{
+			State: nsxtPolicyPathResourceImporter,
```
annakhm, line 24: I don't think this can work out of the box here, since enforcement point and site need to be set before Read is called?

annakhm:
Is it too complex to add a test for this? perhaps with env variable indicating VC compute ID?

ksamoray:
> Is it too complex to add a test for this? perhaps with env variable indicating VC compute ID?

It means that we'll need a cluster on VC which isn't configured with NSX. We can create such with the LM topology but it'll require an additional ESXi node per topology.

annakhm (approved):

PR #967, opened 2023-09-13
Fix rule order bug

annakhm:
NSX can not guarantee rule order unless sequence_number is provided, or `revise` API is called.
This change auto-assigns sequence number to all rules in list, for which it was not provided by the user.
It takes care of provider upgrade by making sure rules without assigned sequence number are updated even if terraform does not detect a change in them.

### DIFF
--- a/nsxt/policy_common.go
+++ b/nsxt/policy_common.go
@@ -440,6 +440,7 @@ func validatePolicyRuleSequence(d *schema.ResourceData) error {
 func getPolicyRulesFromSchema(d *schema.ResourceData) []model.Rule {
 	rules := d.Get("rule").([]interface{})
 	var ruleList []model.Rule
+	lastSequence := int64(0)
 	for _, rule := range rules {
 		data := rule.(map[string]interface{})
 		displayName := data["display_name"].(string)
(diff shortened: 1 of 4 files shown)


nsxt/policy_common.go
```diff
@@ -469,6 +470,11 @@ func getPolicyRulesFromSchema(d *schema.ResourceData) []model.Rule {
 		}
 
 		resourceType := "Rule"
+		if sequenceNumber == 0 {
+			sequenceNumber = lastSequence + 1
```
ksamoray, line 474: Would it make sense to set sequence numbers with some gap between them? e.g 10, 20... etc so later a rule can be inserted between without recreating the whole list?

ksamoray (approved):
LGTM, just a question for understanding how it works.

nsxt/policy_common.go
annakhm: From terraform perspective the list is ordered, which means if rule 3.5 is inserted between rule 3 and rule 4, this will result in diff for rule 4, and a new rule 5. As a result, we will update all rules starting from the inserted rule, hence we don't need to bother about gaps in sequence numbers.

PR #976, opened 2023-09-19
Improve Compute Manager data source

annakhm:
Since usually only one compute manager is configured, it is useful to retrieve its ID without providing any filters. This PR enables this.

### DIFF
--- a/nsxt/data_source_nsxt_compute_manager.go
+++ b/nsxt/data_source_nsxt_compute_manager.go
@@ -18,7 +18,7 @@ func dataSourceNsxtComputeManager() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"id":           getDataSourceIDSchema(),
-			"display_name": getDisplayNameSchema(),
+			"display_name": getDataSourceExtendedDisplayNameSchema(),
 			"description":  getDescriptionSchema(),
 			"server": {
 				Type:        schema.TypeString,
@@ -45,8 +45,6 @@ func dataSourceNsxtComputeManagerRead(d *schema.ResourceData, m interface{}) err
 			return fmt.Errorf("failed to read ComputeManager %s: %v", objID, err)
 		}
 		obj = objGet
-	} else if objName == "" {
-		return fmt.Errorf("error obtaining Compute Manager ID or name during read")
 	} else {
 		// Get by full name/prefix
 		objList, err := client.List(nil, nil, nil, nil, nil, nil, nil)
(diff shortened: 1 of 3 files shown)


nsxt/data_source_nsxt_compute_manager.go
```diff
@@ -57,6 +55,11 @@ func dataSourceNsxtComputeManagerRead(d *schema.ResourceData, m interface{}) err
 		var perfectMatch []model.ComputeManager
 		var prefixMatch []model.ComputeManager
 		for _, objInList := range objList.Results {
+			if len(objName) == 0 {
+				// We want to grab single compute manager
+				perfectMatch = append(perfectMatch, objInList)
```
GraysonWu, line 60: So if there are multiple compute managers, an empty `display_name` will cause an error. I'm wondering if we should add a more informative err msg 
 indicating empty display_name can only be used with single compute manager to reduce the confusion or change the behavior here to just grab the first compute manager or just treat it as a match all for prefixMatch. What do you think?

nsxt/data_source_nsxt_compute_manager_test.go
```diff
@@ -11,24 +11,44 @@ import (
...
+		},
+	})
+}
+
+func TestAccDataSourceNsxtComputeManager_single(t *testing.T) {
+	testResourceName := "data.nsxt_compute_manager.test"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck: func() {
+			testAccOnlyLocalManager(t)
```
ksamoray, line 42: Not sure whether we need `testAccTestMP()` for these resources or not. Other than this LGTM

ksamoray (approved):

nsxt/data_source_nsxt_compute_manager_test.go
annakhm: I think we should rename `testAccTestMP` to `testAccTestDeprecated`. I think we should include the new fabric resources in the main test suite

annakhm: Thanks!

nsxt/data_source_nsxt_compute_manager.go
annakhm: I don't think we should decide which manager to grab without further input from the user. 
Failing the data source because multiple objects are matching is a behavior present in most data sources in the provider. Regarding error messages - perhaps we should unify these messages across all data sources for consistency? 
I can add more info in the doc for the case of compute manager.

PR #978, opened 2023-09-20
Fix crash in transport node resource

annakhm:
In addition, mark computed fields as such. to avoid non-empty diff problem. Sinsitive attributes (such as `root_password`) will be addressed in follow-up PR.

### DIFF
--- a/nsxt/resource_nsxt_transport_node.go
+++ b/nsxt/resource_nsxt_transport_node.go
@@ -99,6 +99,7 @@ func resourceNsxtTransportNode() *schema.Resource {
 			"failure_domain": {
 				Type:        schema.TypeString,
 				Optional:    true,
+				Computed:    true,
 				Description: "Id of the failure domain",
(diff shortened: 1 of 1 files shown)


nsxt/resource_nsxt_transport_node.go
```diff
@@ -157,6 +159,7 @@ func getNodeSchema(addlAttributes map[string]*schema.Schema, addExactlyOneOf boo
 		"ip_addresses": {
 			Type:        schema.TypeList,
 			Optional:    true,
+			Computed:    true,
```
GraysonWu, line 162: I'm not sure about this. Is this field used by users to locate the node?

GraysonWu: Ah, I see what you mean. When read it back this field might contain more IP addresses.

ksamoray (approved):

PR #983, opened 2023-09-21
Add User Management Resources

### LINKED ISSUE #943: Management of aaa resources (roles, role bindings, ...)

I want to manage roles and role bindings (ldap user -> role) with terraform.

Important api paths are:
* [GET /policy/api/v1/aaa/roles/{role}](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/201a09cb-5610-49e6-9d31-7db0920e3a65/65f199d6-8516-4092-9b19-906559f972b5/api_includes/method_GetRoleInfo.html)
* [PUT /policy/api/v1/aaa/roles/{role}](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/201a09cb-5610-49e6-9d31-7db0920e3a65/65f199d6-8516-4092-9b19-906559f972b5/api_includes/method_CreateOrUpdateRole.html)
* 

* [GET /policy/api/v1/aaa/role-bindings](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/201a09cb-5610-49e6-9d31-7db0920e3a65/65f199d6-8516-4092-9b19-906559f972b5/api_includes/method_GetAllRoleBindings.html)
* [POST /policy/api/v1/aaa/role-bindings](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/201a09cb-5610-49e6-9d31-7db0920e3a65/65f199d6-8516-4092-9b19-906559f972b5/api_includes/method_CreateRoleBinding.html)
* https://vdc-repo.vmware.com/vmwb-repository/dcr-public/201a09cb-5610-49e6-9d31-7db0920e3a65/65f199d6-8516-4092-9b19-906559f972b5/api_includes/method_UpdateRoleBinding.html
* [DELETE /policy/api/v1/aaa/role-bindings/{binding-id}](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/201a09cb-5610-49e6-9d31-7db0920e3a65/65f199d6-8516-4092-9b19-906559f972b5/api_includes/method_DeleteRoleBinding.html)

Add resources for aaa apis like roles and role bindings.

```terraform
resource "nsxt_aaa_role" "my_role" {
  name = "my_role"
  ...
}
```

```terraform
resource "nsxt_aaa_role_binding" "my_role_binding" {
  name = "my_role_binding"
  ...
}
```

```terraform
resource "nsxt_aaa_role_binding" "my_role_binding" {
  name = "my_role_binding"
  ...
}
```

```terraform
data "nsxt_ldap_identity_source" "my_ldap" { # needed for ldap role bindings
  ...
}

_No response_

### Additional context

Issue https://github.com/vmware/terraform-provider-nsxt/issues/815 goes in the same direction for a Identity Principal Resource which is part of the aaa apis, too.

### PR DESCRIPTION
**Pull request by wsquan171**
This PR adds the following 3 resources related to user management:
- Local users (`/api/v1/node/users`)
- Roles (`policy/api/v1/aaa/roles`)
- Role bindings (`policy/api/v1/aaa/role-bindings`)

#### Users ####
- As password is not returned on GET, and it's possible for terraform created users to change the password out of terraform, password updates from config is not checked. Password resets could be done by deactivating then reactivate the same account.
- SSH keys (`api/v1/node/users/{userid}/ssh-keys`) and auth policy (`/api/v1/node/aaa/auth-policy`) not included

#### Roles ####
- Role validation API doesn't return error on invalid set of feature permissions. Instead, a list of "recommended" permissions will be returned. An error is raised reporting such missing permissions if the list is not empty.
- Multi-tenancy API endpoints supports GET only, thus not included.

#### Role Bindings ####
- For local users, role bindings is owned by NSX and created upon user creation. Role Binding Create and Delete is not allowed. Such role bindings must be imported into terraform for modifications.
- To remove an entire `path` of a set of roles from a binding, the path needs to be marked for delete in the `PUT` call, with non-empty role in the list. A random role is picked from old state for the call to make NSX API validation pass.
- Full CRUD for remote user bindings are tested with LDAP.
- Acc test requires env `NSXT_TEST_LDAP_USER` to be set to a valid LDAP user, and NSX is enabled with LDAP identity source.

Partially resolve #943

### DIFF
--- a/nsxt/resource_nsxt_node_user.go
+++ b/nsxt/resource_nsxt_node_user.go
@@ -1,0 +1,240 @@
+/* Copyright © 2023 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	nsxModel "github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/model"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/node"
+)
+
+func resourceNsxtUsers() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNsxtNodeUserCreate,
+		Read:   resourceNsxtNodeUserRead,
+		Update: resourceNsxtNodeUserUpdate,
+		Delete: resourceNsxtNodeUserDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"full_name": {
+				Type:        schema.TypeString,
+				Description: "Full name for the user",
+				Required:    true,
+			},
+			"last_password_change": {
+				Type:        schema.TypeInt,
+				Description: "Number of days since password was last changed",
+				Computed:    true,
+			},
+			"password": {
+				Type:        schema.TypeString,
+				Description: "Old password for the user",
+				Sensitive:   true,
+				Optional:    true,
+			},
+			"password_change_frequency": {
+				Type:         schema.TypeInt,
+				Description:  "Number of days password is valid before it must be changed",
+				Optional:     true,
+				Default:      90,
+				ValidateFunc: validation.IntBetween(0, 9999),
+			},
+			"password_change_warning": {
+				Type:         schema.TypeInt,
+				Description:  "Number of days before user receives warning message of password expiration",
+				Optional:     true,
+				Default:      7,
+				ValidateFunc: validation.IntBetween(0, 9999),
+			},
+			"password_reset_required": {
+				Type:        schema.TypeBool,
+				Description: "Boolean value that states if a password reset is required",
+				Computed:    true,
+			},
+			"active": {
+				Type:        schema.TypeBool,
+				Description: "Boolean value that states if the user account is activated",
+				Optional:    true,
+				Default:     true,
+			},
+			"status": {
+				Type:        schema.TypeString,
+				Description: "User status",
+				Computed:    true,
+			},
+			"user_id": {
+				Type:        schema.TypeInt,
+				Description: "Numeric id for the user",
+				Computed:    true,
+			},
+			"username": {
+				Type:         schema.TypeString,
+				Description:  "User login name",
+				ValidateFunc: validateUsername(),
+				Required:     true,
+			},
+		},
+	}
+}
+
+func validateUsername() schema.SchemaValidateFunc {
+	return func(i interface{}, k string) (s []string, es []error) {
+		v, ok := i.(string)
+		if !ok {
+			es = append(es, fmt.Errorf("expected type of %s to be string", k))
+			return
+		}
+
+		r := regexp.MustCompile("^[a-zA-Z][a-zA-Z0-9@-_.\\-]*$")
+		if ok := r.MatchString(v); !ok {
+			es = append(es, fmt.Errorf("must be a valid username matching: ^[a-zA-Z][a-zA-Z0-9@-_.\\-]*$"))
+			return
+		}
+
+		if len(v) < 1 || len(v) > 32 {
+			es = append(es, fmt.Errorf("expected length of %s to be in the range (%d - %d), got %s", k, 1, 32, v))
+			return
+		}
+		return
+	}
+}
+
+func resourceNsxtNodeUserCreate(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+	client := node.NewUsersClient(connector)
+
+	fullName := d.Get("full_name").(string)
+	passwordChangeFrequency := int64(d.Get("password_change_frequency").(int))
+	passwordChangeWarning := int64(d.Get("password_change_warning").(int))
+	username := d.Get("username").(string)
+	password := d.Get("password").(string)
+
+	userProp := nsxModel.NodeUserProperties{
+		FullName:                &fullName,
+		PasswordChangeFrequency: &passwordChangeFrequency,
+		PasswordChangeWarning:   &passwordChangeWarning,
+		Username:                &username,
+	}
+	if len(password) > 0 {
+		userProp.Password = &password
+	}
+
+	user, err := client.Createuser(userProp)
+	if err != nil {
+		return handleCreateError("User", username, err)
+	}
+	d.Set("user_id", user.Userid)
+	d.SetId(strconv.Itoa(int(*user.Userid)))
+
+	return resourceNsxtNodeUserRead(d, m)
+}
+
+func resourceNsxtNodeUserRead(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+	client := node.NewUsersClient(connector)
+	id := d.Id()
+	if id == "" {
+		return fmt.Errorf("error obtaining logical object id")
+	}
+	user, err := client.Get(id)
+	if err != nil {
+		return handleReadError(d, "User", id, err)
+	}
+
+	// Password not return on GET
+	d.Set("full_name", user.FullName)
+	d.Set("last_password_change", user.LastPasswordChange)
+	d.Set("password_change_frequency", user.PasswordChangeFrequency)
+	d.Set("password_change_warning", user.PasswordChangeWarning)
+	d.Set("password_reset_required", user.PasswordResetRequired)
+	d.Set("status", user.Status)
+	d.Set("user_id", user.Userid)
+	d.Set("username", user.Username)
+	if *user.Status == nsxModel.NodeUserProperties_STATUS_NOT_ACTIVATED {
+		d.Set("active", false)
+	} else {
+		d.Set("active", true)
+	}
+	return nil
+}
+
+func resourceNsxtNodeUserUpdate(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+	client := node.NewUsersClient(connector)
+	id := d.Id()
+	if id == "" {
+		return fmt.Errorf("error obtaining logical object id")
+	}
+	password := d.Get("password").(string)
+	active := d.Get("active").(bool)
+	status := d.Get("status").(string)
+
+	// Handle user status change first
+	// Password reset can be achieved by deactivating then re-activate the account
+	if status == nsxModel.NodeUserProperties_STATUS_NOT_ACTIVATED && active {
+		if len(password) == 0 {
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: nsxt/resource_nsxt_node_user.go [hunk 1]
```diff
@@ -0,0 +1,238 @@
...
+			"password_change_warning": {
+				Type:         schema.TypeInt,
+				Description:  "Number of days before user receives warning message of password expiration",
+				Optional:     true,
+				Default:      7,
+				ValidateFunc: validation.IntBetween(0, 9999),
+			},
+			"password_reset_required": {
+				Type:        schema.TypeBool,
+				Description: "Boolean value that states if a password reset is required\t",
```
annakhm, line 66: Cosmetics - remove `\t` at the end

nsxt/resource_nsxt_node_user.go
```diff
@@ -0,0 +1,238 @@
...
+				Computed:    true,
+			},
+			"status": {
```
annakhm, line 69: This is marked as ReadOnly in API spec, so it should be Computed only (not Optional), and `ValidateFunc` is not needed

nsxt/resource_nsxt_node_user.go
```diff
@@ -0,0 +1,238 @@
...
+		PasswordChangeWarning:   &passwordChangeWarning,
+		Username:                &username,
+	}
+	if len(password) > 0 {
+		userProp.Password = &password
+	}
+
+	user, err := client.Createuser(userProp)
+	if err != nil {
+		return fmt.Errorf("failed to create Nsxt Node user: %s", err.Error())
```
annakhm, line 135: Please use handleCreateError here for consistency, same below

nsxt/resource_nsxt_node_user.go
```diff
@@ -0,0 +1,238 @@
...
+				Computed:    true,
+			},
+			"status": {
+				Type:         schema.TypeString,
+				Description:  "User status",
+				Optional:     true,
+				ValidateFunc: validation.StringInSlice(nodeUserStatus, false),
+				Computed:     true,
+			},
+			"userid": {
```
annakhm, line 76: Lets rename to `user_id` for consistency?

nsxt/resource_nsxt_node_user.go
annakhm: After looking further, I think this attribute is used to activate/deactivate users? I think its a bit confusing since its not clear why would the user specify `PASSWORD_EXPIRED` value.. Perhaps to make this more intuitive, we can offer boolean `active` attribute instead? And leave `status` as readonly..

nsxt/resource_nsxt_node_user.go
```diff
@@ -0,0 +1,238 @@
...
+func resourceNsxtNodeUserRead(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+	client := node.NewUsersClient(connector)
+	id := d.Id()
+	if id == "" {
+		return fmt.Errorf("error obtaining logical object id")
+	}
+	user, err := client.Get(id)
+	if err != nil {
+		if isNotFoundError(err) {
```
annakhm, line 152: This behavior is covered with `handleReadError`

annakhm:
We have policy API for roles, but not for users?

nsxt/resource_nsxt_node_user.go
wsquan171: done

nsxt/resource_nsxt_node_user.go
wsquan171: Done. Added bool `active` for activation / deactivation of users, and made`status` to be read-only

nsxt/resource_nsxt_node_user.go
wsquan171: done

nsxt/resource_nsxt_node_user.go
wsquan171: done

nsxt/resource_nsxt_node_user.go
wsquan171: done

wsquan171:
> We have policy API for roles, but not for users?

Unfortunately, yes. Local (node) users are only supported with management API.

nsxt/resource_nsxt_node_user.go
```diff
@@ -0,0 +1,240 @@
...
+	return func(i interface{}, k string) (s []string, es []error) {
+		v, ok := i.(string)
+		if !ok {
+			es = append(es, fmt.Errorf("expected type of %s to be string", k))
+			return
+		}
+
+		r := regexp.MustCompile("^[a-zA-Z][a-zA-Z0-9@-_.\\-]*$")
+		if ok := r.MatchString(v); !ok {
+			es = append(es, fmt.Errorf("must be a valid username matching: ^[a-zA-Z][a-zA-Z0-9@-_.\\-]*$"))
```
annakhm, line 100: I don't think we should expect users to understand regex..

nsxt/resource_nsxt_node_user.go
```diff
@@ -0,0 +1,240 @@
...
+	id := d.Id()
+	if id == "" {
+		return fmt.Errorf("error obtaining logical object id")
+	}
+	user, err := client.Get(id)
+	if err != nil {
+		return handleReadError(d, "User", id, err)
+	}
+
+	// Password not return on GET
```
annakhm, line 154: Would there be a non-empty diff after apply since password is not set in state?

nsxt/resource_nsxt_node_user.go
wsquan171: removed regex from the error message.

nsxt/resource_nsxt_node_user.go
wsquan171: Per terraform dev guide:

> If the Update callback returns with or without an error, the full state is saved. If the ID becomes blank, the resource is destroyed (even within an update, though this shouldn't happen except in error scenarios).

Verified that by changing only the password, `terraform plan` returns empty action plan after the initial apply, and reports:

> No changes. Your infrastructure matches the configuration.

annakhm: Thanks for verifying. Since `password` is not nested attribute, no problem indeed.

nsxt/resource_nsxt_node_user.go
```diff
@@ -0,0 +1,240 @@
...
+		return fmt.Errorf("error obtaining logical object id")
+	}
+
+	// Workaround a bug where the delete response body is marked as non-json content type by NSX,
+	// triggering SDK to raise it as internal server error.
+	// Error Data is not complying with ApiErrorBindingType, making the conversion result empty.
+	// A second deletion attempt should be deemed as NotFound, which is then treated as successful.
+	var err error
+	for i := 0; i < 2; i++ {
+		err = client.Delete(id)
```
annakhm, line 231: does the first delete always return error?

nsxt/resource_nsxt_policy_role_binding.go
```diff
@@ -0,0 +1,320 @@
...
+		Create: resourceNsxtPolicyUserManagementRoleBindingCreate,
+		Read:   resourceNsxtPolicyUserManagementRoleBindingRead,
+		Update: resourceNsxtPolicyUserManagementRoleBindingUpdate,
+		Delete: resourceNsxtPolicyUserManagementRoleBindingDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"display_name": getDataSourceDisplayNameSchema(),
```
annakhm, line 43: is there a reason we don't allow controlling `nsx_id` here and in the other policy resource?

nsxt/resource_nsxt_policy_role.go
```diff
@@ -0,0 +1,266 @@
...
+	return false, logAPIError("Error retrieving resource", err)
+}
+
+func getFeaturePermissionFromSchema(d *schema.ResourceData) []nsxModel.FeaturePermission {
+	features := d.Get("feature").([]interface{})
+	featurePermissions := make([]nsxModel.FeaturePermission, 0, len(features))
+	for _, feature := range features {
+		data := feature.(map[string]interface{})
+		featureID := data["feature"].(string)
+		featureDescription := data["feature_description"].(string)
```
annakhm, line 102: since `feature_description` and `feature_name` are Computed-only, we don't need those here I think

nsxt/resource_nsxt_policy_role.go
```diff
@@ -0,0 +1,266 @@
...
+			id, missingPerm)
+	}
+
+	obj := nsxModel.RoleWithFeatures{
+		DisplayName: &displayName,
+		Description: &description,
+		Tags:        tags,
+		Role:        &role,
+		Features:    features,
+	}
```
annakhm, line 179: is `revision` missing here for update?

nsxt/resource_nsxt_policy_role.go
```diff
@@ -0,0 +1,266 @@
...
+		elem["permission"] = permission.Permission
+		permissionList = append(permissionList, elem)
+	}
+	err := d.Set("feature", permissionList)
+	if err != nil {
+		log.Printf("[WARNING] Failed to set feature in schema: %v", err)
+	}
+}
+
+func policyUserManagementRolePatch(id string, d *schema.ResourceData, m interface{}) error {
```
annakhm, line 143: Since the actual API call is `Update` and not `Patch`, I would suggest to change function name to improve readability

nsxt/resource_nsxt_policy_role_binding.go
```diff
@@ -0,0 +1,320 @@
...
+			"description":  getDataSourceDescriptionSchema(),
+			"revision":     getRevisionSchema(),
```
annakhm, line 45: I'm guessing `path` is not exposed in this API as well?

wsquan171: Yes. Neither role or role bindings exports policy path.

nsxt/resource_nsxt_node_user.go
wsquan171: Essentially, the first DELETE call actually went through, with HTTP status 200. It's just the returned HTTP header `content-type`: `text/html; charset=UTF-8` would always trigger sdk to report it back as internal server error, and such error type casting made checking status code very difficult on our end.

The second call will make NSX return 404, which is ignored on delete calls.

This issue still exists on NSX 4.1.2 rtqa.

nsxt/resource_nsxt_policy_role.go
wsquan171: done

nsxt/resource_nsxt_policy_role.go
wsquan171: added

nsxt/resource_nsxt_policy_role.go
wsquan171: Renamed.

annakhm (approved):

annakhm:
Thanks Shawn!

PR #986, opened 2023-09-27
Change default EdgeCluster name

GraysonWu:


website/docs/r/dhcp_server_ip_pool.html.markdown
```diff
@@ -14,7 +14,7 @@ Provides a resource to configure IP Pool for logical DHCP server on NSX-T manage
 
 ```hcl
 data "nsxt_edge_cluster" "edgecluster" {
-  display_name = "edgecluster1"
+  display_name = "EDGECLUSTER1"
```
annakhm, line 17: can we use the env variable here and below?

ksamoray:
In the new topologies which I deploy with Terraform, the names are actually lowercased :)
If we want to use uppercase naming, I'll change that as well.

salv-orlando:
These are not NSX defaults, but rather default settings for the process used to deploy NSX. As @ksamoray mentioned, we will shortly start using topologies deployed via Terraform and defaulting to lower case names.

Also, any NSX user should be able to run these tests on their own topology and have a different name for the edge cluster to use in tests.
If the above makes sense, perhaps these constants should become variables that are read from a config file.

As this might take time for the time being can use two values depending on backend NSX version?

ksamoray:
@salv-orlando these values are just default names which are used in the testing environments, the CI sets these with environment vars and I already changed the topology to use uppercase values as the GM testing deployments use.

website/docs/r/dhcp_server_ip_pool.html.markdown
salv-orlando: Let's just not make any change to the doc file. It's not relevant to this PR

website/docs/r/logical_dhcp_server.html.markdown
```diff
@@ -14,7 +14,7 @@ Provides a resource to configure logical DHCP server on NSX-T manager
 
 ```hcl
 data "nsxt_edge_cluster" "edgecluster" {
-  display_name = "edgecluster1"
+  display_name = "EDGECLUSTER1"
```
salv-orlando, line 17: as above

website/docs/r/dhcp_server_ip_pool.html.markdown
GraysonWu: Changed back.

website/docs/r/logical_dhcp_server.html.markdown
GraysonWu: Done.

nsxt/utils_test.go
```diff
@@ -27,7 +27,8 @@ import (
 // Default names or prefixed of NSX backend existing objects used in the acceptance tests.
 // Those defaults can be overridden using environment parameters
 const tier0RouterDefaultName string = "PLR-1 LogicalRouterTier0"
-const edgeClusterDefaultName string = "edgecluster1"
+const edgeClusterDefaultName string = "EDGECLUSTER1"
+const computeCollectionDefaultName string = "Cluster-1"
```
ksamoray, line 31: For local manager environment, this should be Compute-Cluster if I recall (it's the name of the VC cluster which we deploy).

ksamoray: BTW the value is set within the Jenkins node but it's better to have a correct default value :)

GraysonWu: Not sure how that env is deployed. In the test topology `Policy-L2L3-OnPrem` it's called `Cluster-1`.

ksamoray (approved):

PR #987, opened 2023-09-27
Rename uplink host switch profile for consistency

annakhm:
In order to avoid confusion, objects based on policy API should contain _policy prefix. We omitted the prefix for sake of shorter names in some fabric resources, however this causes confusion in dependency management, since policy resources relay on 'path' rather than 'id'.


salv-orlando (approved):
LGTM!

PR #989, opened 2023-10-04
Fix login credential type identifier parsing

ksamoray:
Values used for type identifier parsing were incorrect

### DIFF
--- a/nsxt/resource_nsxt_compute_manager.go
+++ b/nsxt/resource_nsxt_compute_manager.go
@@ -398,7 +398,7 @@ func resourceNsxtComputeManagerRead(d *schema.ResourceData, m interface{}) error
 
 func setCredentialValuesInSchema(d *schema.ResourceData, credential *data.StructValue) error {
 	converter := bindings.NewTypeConverter()
-	elem := make(map[string]interface{})
+	parentElem := getElemOrEmptyMapFromSchema(d, "credential")
 
 	base, errs := converter.ConvertToGolang(credential, model.LoginCredentialBindingType())
 	if errs != nil {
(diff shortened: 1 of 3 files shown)


nsxt/resource_nsxt_compute_manager.go
```diff
@@ -449,7 +449,7 @@ func setCredentialValuesInSchema(d *schema.ResourceData, credential *data.Struct
 		return errors.New("no valid credential found")
 	}
 
-	d.Set("credential", elem)
+	d.Set("credential", []interface{}{elem})
```
salv-orlando, line 452: was this wrong before and we did not realize only because elem was an empty map?

ksamoray: Indeed. And the list was empty as the constants which were used in the switch case were wrong as well. So values were taken from the state.

nsxt/resource_nsxt_compute_manager.go
```diff
@@ -435,7 +435,7 @@ func setCredentialValuesInSchema(d *schema.ResourceData, credential *data.Struct
 		elem["thumbprint"] = credEntry.Thumbprint
 		elem["password"] = credEntry.Password
```
annakhm, line 436: Would we see non-empty diff here since sensitive data is not sent back by NSX?

ksamoray: Yep - in transport node the polymorphic objects are recreated with make(map[]interface{}), while here the same map is being populated with the GET values. So passwords are retained from the state.
I'm implementing a similar mechanism for transport node.

annakhm: Sorry - I'm still missing something.. I see `elem` is initialized as an empty map, and then populated with `GET` values. I'm missing the part where passwords are coming from the state..

annakhm (approved):

PR #992, opened 2023-10-05
Refactor manager cluster resource

annakhm:
Main changes:
1. Resolve provider host address, if needed, rather than relying on comparing host names in cluster config and in provider config
2. Rely on nodes IP version rather than host IP version, which is unknown when not specified in provider config
3. Refactor Read function to update only Computed fields in state

### DIFF
--- a/nsxt/resource_nsxt_manager_cluster.go
+++ b/nsxt/resource_nsxt_manager_cluster.go
@@ -102,71 +102,28 @@ func getClusterNodesFromSchema(d *schema.ResourceData) []NsxClusterNode {
 	return clusterNodes
 }
 
-func setNodesInSchema(d *schema.ResourceData, nodes []NsxClusterNode) error {
-	// Retrieve node credential from schema and set them in the nodeList element
-	// This is because the nodes are obtained using cluster client, and does not
-	// contain any information about credential
-	nodeIPToCredentialMap := getIPtoCredentialMap(d)
-
-	var nodeList []map[string]interface{}
-	for _, node := range nodes {
-		_, nodeInSchema := nodeIPToCredentialMap[node.IPAddress]
-		elem := make(map[string]interface{})
-		elem["id"] = node.ID
-		elem["ip_address"] = node.IPAddress
-		// If node is not in schema but in cluster node list,
-		// it means the node needs to be removed.
-		// In this case we don't need its credential because
-		// remove api does not require node credential
-		if nodeInSchema {
-			elem["username"] = nodeIPToCredentialMap[node.IPAddress][0]
-			elem["password"] = nodeIPToCredentialMap[node.IPAddress][1]
-		}
-		elem["fqdn"] = node.Fqdn
-		elem["status"] = node.Status
-		nodeList = append(nodeList, elem)
-	}
-	return d.Set("node", nodeList)
-}
-
-func getIPtoCredentialMap(d *schema.ResourceData) map[string][]string {
-	// returns a map, whose key is IP address of a node,
-	// value is a slice of [node_login_username, node_login_password]
-	nodes := d.Get("node").([]interface{})
-	var idToCredentialMap = map[string][]string{}
-	for _, node := range nodes {
-		data := node.(map[string]interface{})
-		ipAddress := data["ip_address"].(string)
-		userName := data["username"].(string)
-		password := data["password"].(string)
-		cred := []string{userName, password}
-		idToCredentialMap[ipAddress] = cred
-	}
-	return idToCredentialMap
-}
-
 func resourceNsxtManagerClusterCreate(d *schema.ResourceData, m interface{}) error {
 	// Call Joincluster function on nodes that are not in the cluster
 	nodes := getClusterNodesFromSchema(d)
 	if len(nodes) == 0 {
 		return fmt.Errorf("At least a manager appliance must be provided to form a cluster")
 	}
-	clusterID, certSha256Thumbprint, hostIP, err := getClusterInfoFromHostNode(d, m)
+	clusterID, certSha256Thumbprint, hostIPs, err := getClusterInfoFromHostNode(d, m)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -314,6 +258,34 @@ func getHostCredential(m interface{}) (string, string) {
 	return username, password
 }
 
+func getMatchingIPVersion(ip string, hostIPs []string) string {
+	needIPv4 := (net.ParseIP(ip)).To4() != nil
+
+	for _, hostIP := range hostIPs {
+		isIPv4 := (net.ParseIP(hostIP)).To4() != nil
```
salv-orlando, line 265: are 100% sure hostIp is always a valid Ip address? otherwise we might first need to check ParseIP result.

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -314,6 +258,34 @@ func getHostCredential(m interface{}) (string, string) {
...
+		if needIPv4 == isIPv4 {
```
salv-orlando, line 266: The code looks good but I find some difficulties to read it. It's not easy to understand that if needIPv4=False and isIPv4=False then we want to return an IPv6 address?

Do you agree or is it just me getting older?

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -314,6 +258,34 @@ func getHostCredential(m interface{}) (string, string) {
...
+	return ""
+}
+
+func isMatchingNode(node nsxModel.ClusterNodeInfo, address string) bool {
+	for _, entity := range node.Entities {
+		if entity.Port != nil {
+			if (entity.IpAddress != nil) && (*entity.IpAddress == address) {
+				return true
+			}
+			if (entity.Ipv6Address != nil) && (*entity.Ipv6Address == address) {
```
salv-orlando, line 280: shall we convert string to IPs to avoid equality check to fail due to differences in Ipv6 short/full representation?

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -175,75 +132,58 @@ func getClusterInfoFromHostNode(d *schema.ResourceData, m interface{}) (string,
 	min := c.CommonConfig.MinRetryInterval
 	max := c.CommonConfig.MaxRetryInterval
 	maxRetries := c.CommonConfig.MaxRetries
-	hostIP := ""
+	hostIPs := []string{}
 	for i := 0; i < maxRetries; i++ {
 		clusterConfig, err := client.Get()
 		if err != nil {
-			return "", "", "", handleReadError(d, "Cluster Config", "", err)
+			return "", "", hostIPs, err
```
salv-orlando, line 139: if this fails after the first iteration we might return a non empty host IP list.
Is that expected?

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -175,75 +132,58 @@ func getClusterInfoFromHostNode(d *schema.ResourceData, m interface{}) (string,
...
+				return "", "", hostIPs, err
 			}
+
+			log.Printf("[DEBUG]: Host resolved to IP addresses %v", hostIPs)
 		}
 		clusterID := *clusterConfig.ClusterId
 		nodes := clusterConfig.Nodes
 		node := nodes[0]
 		apiListenAddr := node.ApiListenAddr
 		if apiListenAddr != nil {
```
salv-orlando, line 153: Could it be better to first do this check, and only if it succeeds collect hostIPs? It seems while we wait for ApiListenAddr to populate we will keep repeating `resolveHostIPs`

nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -175,75 +132,58 @@ func getClusterInfoFromHostNode(d *schema.ResourceData, m interface{}) (string,
...
-	return "", "", "", fmt.Errorf("Failed to read ClusterConfig info from host node after %d attempts", maxRetries)
+	return "", "", hostIPs, fmt.Errorf("Failed to read ClusterConfig after %d attempts", maxRetries)
 }
 
-func getHostIPFromClusterConfig(client interface{}, clusterConfig nsxModel.ClusterConfig) (string, error) {
+func resolveHostIPs(client interface{}) ([]string, error) {
 	c := client.(nsxtClients)
 	host := c.Host
 	host = strings.TrimPrefix(host, "https://")
 	// Check if host is ip address or fqdn, if it's ip address then we are all set,
```
2ez4szliu, line 168: with the approach in resolveHostIPs I think this comment section is no longer needed

nsxt/resource_nsxt_manager_cluster.go
annakhm: We validate hostIP in `resolveHostIPs`, which is happening before this call

nsxt/resource_nsxt_manager_cluster.go
annakhm: Its a bit tricky to read, I agree, I've added a comment

nsxt/resource_nsxt_manager_cluster.go
annakhm: `resolveHostIPs` only happens on first iteration, if `len(hostIPs) == 0`, so it shouldn't be repeated..

nsxt/resource_nsxt_manager_cluster.go
annakhm: Updated the comment, thanks

nsxt/resource_nsxt_manager_cluster.go
annakhm: Should be ok since error is always checked, and other return values are disregarded when error is present

ksamoray (approved):
LGTM, if no other comments from reviewers, it's good to go.

PR #993, opened 2023-10-05
Fix searching by special chars in data sources

annakhm:
For data sources that rely on search API, functionality was broken when searched string contained special characters. This change fixes the issue by escaping special characters according to search API spec.


wsquan171 (approved):
LGTM

PR #994, opened 2023-10-06
Avoid applying lisence on every init

annakhm:
Previous code assumed GetChange SDK routine works for provider attributes, which it doesn't (provider attributes are not stored in state).
In order to avoid reapplying licenses on every init, this change pulls license keys from backend and only applies/removes the diff.

### DIFF
--- a/nsxt/provider.go
+++ b/nsxt/provider.go
@@ -34,12 +34,6 @@ import (
 
 var defaultRetryOnStatusCodes = []int{400, 409, 429, 500, 503, 504}
 
-// Struct to keep track of changes in schema List
-type listDiff struct {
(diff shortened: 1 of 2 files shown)


ksamoray:
Do we support a case where a user adds a license via UI? I think that in this case, this license will be removed.

annakhm:
> Do we support a case where a user adds a license via UI? I think that in this case, this license will be removed.

Good point Kobi. I was contemplating doing nothing if license is not specified. This will cause an issue when removing licenses that were specified earlier..

ksamoray (approved):
LGTM. Maybe the removal note should be more visible?

PR #997, opened 2023-10-10
On read, extract sensitive data from state

ksamoray:
NSX does not return sensitive data (e.g passwords) on GET operation, and therefore TF detects a diff.
But as the schema.Resource object contains the credentials from the state, we can keep these intact and then the value will remain as is and stored again in the state.
This is relevant to the transport_node resource as it contains multiple password attributes, which are as of now placed in empty maps during read - hence state values are ignored.

### DIFF
--- a/nsxt/resource_nsxt_transport_node.go
+++ b/nsxt/resource_nsxt_transport_node.go
@@ -1596,42 +1596,48 @@ func setNodeDeploymentInfoInSchema(d *schema.ResourceData, info *data.StructValu
 		return errs[0]
 	}
 	node := base.(model.Node)
-	elem := make(map[string]interface{})
+	var elem map[string]interface{}
 
-	// Set base object attributes
-	elem["external_id"] = node.ExternalId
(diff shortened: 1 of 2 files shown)


nsxt/resource_nsxt_transport_node.go
```diff
@@ -1596,68 +1596,94 @@ func setNodeDeploymentInfoInSchema(d *schema.ResourceData, info *data.StructValu
...
 	switch nodeType {
 	case model.Node__TYPE_IDENTIFIER:
-		d.Set("node", []map[string]interface{}{elem})
-
+		elems := d.Get("node").([]interface{})
+		elem = elems[0].(map[string]interface{})
+		d.Set("node", []interface{}{elem})
 	case model.EdgeNode__TYPE_IDENTIFIER:
+		elems := d.Get("edge_node").([]interface{})
+		elem = elems[0].(map[string]interface{})
```
annakhm, line 1610: This might not be present (and crush) with `import` command

annakhm (approved):

PR #1001, opened 2023-10-16
Validate rule sequence numbers on create only

annakhm:
Since sequence number is Computed in rule, bad sequence numbers may be stuck in state as a result of configurations with previous provider version.
Rather than throwing an error, the provider will auto-correct the sequence number while also logging this event.
For create flow, we leave sequence number validation since state for new object is clean.


wsquan171 (approved):
One nit otherwise lgtm

PR #1004, opened 2023-10-19
Fix release header template

wsquan171:
This PR is to fix the broken release note header rendering since [v3.2.0](https://github.com/vmware/terraform-provider-nsxt/releases/tag/v3.2.0), due to the command line option in `goreleaser` changed from `--release-header` to `--release-header-tmpl`, where the prior one is now treated as a static file.

Also addresses the following nits in our release action:
- Bumps `ghaction-import-gpg` action to latest to address deprecation warnings of `save-state` and `set-output`
- Uses `--clean` instead of `--rm-dist` per https://goreleaser.com/deprecations/#-rm-dist
- Sets job content write permission per guideline

Takes care of regex quoting of several instances as previously not being captured by PR golang lint jobs.

Testing done:
- Action run under personal repo: https://github.com/wsquan171/terraform-provider-nsxt/actions/runs/6567984764
- Release generated from ^ run: https://github.com/wsquan171/terraform-provider-nsxt/releases/tag/v0.0.2

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,10 @@ on:
   push:
     tags:
       - 'v*'
+
+permissions:
+  contents: write
+
 jobs:
   goreleaser:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


annakhm:
Thanks, releaser changes LGTM, but what is the motivation for changing regex quotes?

annakhm (approved):

PR #1007, opened 2023-10-20
Clarify documentation on sequence number in rules

annakhm:


website/docs/r/policy_intrusion_service_policy.html.markdown
```diff
@@ -85,7 +85,7 @@ In addition to arguments listed above, the following attributes are exported:
 * `rule`:
   * `revision` - Indicates current revision number of the object as seen by NSX-T API server. This attribute can be useful for debugging.
   * `path` - The NSX policy path for this rule.
-  * `sequence_number` - Sequence number of the this rule, is defined by order of rules in the list.
+  * `sequence_number` - Sequence number for this rule, as defined by order of rules in the list.
```
ksamoray, line 88: Why removed "as defined by order of rules in the list" on lines 172, 190 but left here?

annakhm: Here `sequence_number` is read only attribute, so it is only dictated by rule order. In gateway policy and security policy you can specify `sequence_number`

GraysonWu (approved):
LGTM

PR #1008, opened 2023-10-20
Support realized_id in ip pool data source

annakhm:
This id should be used in transport node resource

### DIFF
--- a/nsxt/data_source_nsxt_policy_ip_pool.go
+++ b/nsxt/data_source_nsxt_policy_ip_pool.go
@@ -4,13 +4,10 @@
 package nsxt
 
 import (
-	"fmt"
-	"strings"
-
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 
-	"github.com/vmware/terraform-provider-nsxt/api/infra"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 )
 
 func dataSourceNsxtPolicyIPPool() *schema.Resource {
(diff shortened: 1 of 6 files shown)


ksamoray:
`nsxt_policy_realization_info` data source is usable with IP pools, why do we need this specifically for IP pool?

annakhm:
> `nsxt_policy_realization_info` data source is usable with IP pools, why do we need this specifically for IP pool?

It is just more convenient.. we expose this attribute in resource, makes sense to do this in data source as well

ksamoray:
> > `nsxt_policy_realization_info` data source is usable with IP pools, why do we need this specifically for IP pool?
> 
> It is just more convenient.. we expose this attribute in resource, makes sense to do this in data source as well

But policy_ip_pool resource doesn't have this attribute.

salv-orlando:
> > > `nsxt_policy_realization_info` data source is usable with IP pools, why do we need this specifically for IP pool?
> > 
> > 
> > It is just more convenient.. we expose this attribute in resource, makes sense to do this in data source as well
> 
> But policy_ip_pool resource doesn't have this attribute.

It's always good to be consistent indeed. Also, as the attribute in API spec is called realization_id, I would stay consistent also with naming

ksamoray:
> > > > `nsxt_policy_realization_info` data source is usable with IP pools, why do we need this specifically for IP pool?
> > > 
> > > 
> > > It is just more convenient.. we expose this attribute in resource, makes sense to do this in data source as well
> > 
> > 
> > But policy_ip_pool resource doesn't have this attribute.
> 
> It's always good to be consistent indeed. Also, as the attribute in API spec is called realization_id, I would stay consistent also with naming

We already have `realized_id` attributes in other objects. It's better to remain consistent within the provider than being consistent with NSX.

annakhm:
> > > > > `nsxt_policy_realization_info` data source is usable with IP pools, why do we need this specifically for IP pool?
> > > > 
> > > > 
> > > > It is just more convenient.. we expose this attribute in resource, makes sense to do this in data source as well
> > > 
> > > 
> > > But policy_ip_pool resource doesn't have this attribute.
> > 
> > 
> > It's always good to be consistent indeed. Also, as the attribute in API spec is called realization_id, I would stay consistent also with naming
> 
> We already have `realized_id` attributes in other objects. It's better to remain consistent within the provider than being consistent with NSX.

I've added this attribute to the resource as well

ksamoray (approved):
LGTM. Please rebase on top of master so tests will pass.
I think that maybe in a couple of more we can justify a `getRealizedIdSchema()` func :)

PR #1010, opened 2023-10-24
Add LDAP identity source resource

wsquan171:
This PR adds policy LDAP identity source resource. Both `ActiveDirectoryIdentitySource` and `OpenLdapIdentitySource` type LDAP are supported. The two resources have different resource type on NSX API, but shares the same attributes.

### DIFF
--- a/nsxt/resource_nsxt_policy_ldap_identity_source.go
+++ b/nsxt/resource_nsxt_policy_ldap_identity_source.go
@@ -1,0 +1,333 @@
+/* Copyright © 2023 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/aaa"
+	nsxModel "github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+)
+
+var ldapServerTypes = [](string){
+	nsxModel.LdapIdentitySource_RESOURCE_TYPE_ACTIVEDIRECTORYIDENTITYSOURCE,
+	nsxModel.LdapIdentitySource_RESOURCE_TYPE_OPENLDAPIDENTITYSOURCE,
+}
+
+func resourceNsxtPolicyLdapIdentitySource() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNsxtPolicyLdapIdentitySourceCreate,
+		Read:   resourceNsxtPolicyLdapIdentitySourceRead,
+		Update: resourceNsxtPolicyLdapIdentitySourceUpdate,
+		Delete: resourceNsxtPolicyLdapIdentitySourceDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"nsx_id":       getNsxIDSchema(),
+			"display_name": getDataSourceDisplayNameSchema(),
+			"description":  getDataSourceDescriptionSchema(),
+			"revision":     getRevisionSchema(),
+			"tag":          getTagsSchema(),
+			"type": {
+				Type:         schema.TypeString,
+				Description:  "Indicates the type of LDAP server",
+				Required:     true,
+				ValidateFunc: validation.StringInSlice(ldapServerTypes, false),
+			},
+			"domain_name": {
+				Type:        schema.TypeString,
+				Description: "Authentication domain name",
+				Required:    true,
+			},
+			"base_dn": {
+				Type:        schema.TypeString,
+				Description: "DN of subtree for user and group searches",
+				Required:    true,
+			},
+			"alternative_domain_names": {
+				Type:        schema.TypeList,
+				Description: "Additional domains to be directed to this identity source",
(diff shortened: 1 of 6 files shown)


nsxt/resource_nsxt_policy_ldap_identity_source.go
```diff
@@ -0,0 +1,333 @@
...
+				Required:    true,
+				MaxItems:    1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"bind_identity": {
+							Type:        schema.TypeString,
+							Description: "Username or DN for LDAP authentication",
+							Optional:    true,
+						},
+						"certificates": {
```
annakhm, line 77: is this certificate path?

wsquan171: No. It's the public cert of the servers to tell NSX to trust. Has to be a string of pem. No path allowed.

nsxt/resource_nsxt_policy_ldap_identity_source_test.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+var accTestPolicyLdapIdentitySourceUpdateAttributes = map[string]string{
+	"display_name": getAccTestResourceName(),
+	"description":  "terraform updated",
+}
+
+func TestAccResourceNsxtPolicyLdapIdentitySource_basic(t *testing.T) {
+	testResourceName := "nsxt_policy_ldap_identity_source.test"
+	ldapType := nsxModel.LdapIdentitySource_RESOURCE_TYPE_ACTIVEDIRECTORYIDENTITYSOURCE
```
annakhm, line 27: Can we test both types?

nsxt/resource_nsxt_policy_ldap_identity_source.go
```diff
@@ -0,0 +1,333 @@
...
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"nsx_id":       getNsxIDSchema(),
+			"display_name": getDataSourceDisplayNameSchema(),
+			"description":  getDataSourceDescriptionSchema(),
+			"revision":     getRevisionSchema(),
+			"tag":          getTagsSchema(),
+			"type": {
```
annakhm, line 41: Should this be defined as `ForceNew`, or can resource type be updated on existing object?

nsxt/resource_nsxt_policy_ldap_identity_source.go
```diff
@@ -0,0 +1,333 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/aaa"
+	nsxModel "github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+)
+
+var ldapServerTypes = [](string){
+	nsxModel.LdapIdentitySource_RESOURCE_TYPE_ACTIVEDIRECTORYIDENTITYSOURCE,
```
annakhm, line 21: I would suggest to offer more user friendly values than `OpenLdapIdentitySource` - perhaps `OpenLdap` and `ActiveDirectory`.

nsxt/resource_nsxt_policy_ldap_identity_source.go
```diff
@@ -0,0 +1,333 @@
...
+		Description:            &description,
+		Revision:               &revision,
+		Tags:                   tags,
+		DomainName:             &domainName,
+		BaseDn:                 &baseDn,
+		AlternativeDomainNames: altDomainNames,
+		LdapServers:            ldapServers,
+		ResourceType:           serverType,
+	}
+	dataValue, errs := converter.ConvertToVapi(obj, nsxModel.LdapIdentitySourceBindingType())
```
annakhm, line 217: Would this work for the other server type? I think even it if works now, SDK might add validations in future that would fail if structs diverge. I would suggest to convert to the correct type even if those are identical for now.

nsxt/resource_nsxt_policy_ldap_identity_source_test.go
wsquan171: Do you think we make 2 sets of ENVs (one set for open LDAP another for AD)? Or just add another ENV var to select which type of server to test against?

nsxt/resource_nsxt_policy_ldap_identity_source.go
wsquan171: Done

nsxt/resource_nsxt_policy_ldap_identity_source.go
wsquan171: Done

nsxt/resource_nsxt_policy_ldap_identity_source.go
wsquan171: Converting using proper binding types now.

nsxt/resource_nsxt_policy_ldap_identity_source_test.go
annakhm: Would the same set of ENV variable work for both types? In this case, I would suggest to reuse same vars for both types

wsquan171: `bind_identity` will be different. It all boils down to how we expect this acc test to be run in our pipelines in the future. If we want to cover both in one go then we need 2 servers (openldap + ad) anyways, so 2 sets of VARs. Otherwise the VAR names makes sense for both server types. Just different values should be set to them.

ksamoray:
Do we need some infrastructure to test this?
We could setup an LDAP server or such on the CI jumphost VM if that helps.

annakhm (approved):

wsquan171:
> Do we need some infrastructure to test this? We could setup an LDAP server or such on the CI jumphost VM if that helps.

Kobi, I agree that it's worth testing. I can follow up with test related changes when we have a better idea regarding how we add openldap or AD over ldap in our testing env. Merging this PR for now.

PR #1013, opened 2023-10-25
Rename nsxt_uplink_host_switch_profile

ksamoray:
Previous PR renamed this resource to
nsxt_policy_uplink_host_switch_profile but left a few places which were unchanged.


ksamoray:
@annakhm I'm not sure if you left those on purpose. Also I didn't rename the func names and filename - do you think that we should?

annakhm:
> @annakhm I'm not sure if you left those on purpose. Also I didn't rename the func names and filename - do you think that we should?

I didn't rename the funcs on purpose, but I missed the changes you propose here, thanks!

PR #1015, opened 2023-10-26
Support tenant user creds in T0 and edge cluster

annakhm:
Tenant user should be able to pull T0 gateway and edge cluster that are shared with the project, using corresponding data sources. Get API would not be available for tenant user, so this change switches data source implementation to search API.


ksamoray (approved):
LGTM

PR #1020, opened 2023-10-31
Implement ODS runbook invocation resource

ksamoray:


nsxt/resource_nsxt_policy_ods_runbook_invocation.go
```diff
@@ -0,0 +1,202 @@
...
+		Update: resourceNsxtPolicyODSRunbookInvocationUpdate,
+		Delete: resourceNsxtPolicyODSRunbookInvocationDelete,
+		Importer: &schema.ResourceImporter{
+			State: nsxtPolicyPathResourceImporter,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"nsx_id": getNsxIDSchema(),
+			"path":   getPathSchema(),
+			// Due to a bug, invocations with a display_name specification fail, and when there's none set, NSX assigns
```
annakhm, line 28: So the user can't specify display_name?

nsxt/resource_nsxt_policy_ods_runbook_invocation.go
```diff
@@ -0,0 +1,202 @@
...
+						},
+						"value": {
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: "Value",
+						},
+					},
+				},
+			},
+			"runbook_path": {
```
annakhm, line 58: `getPolicyPathSchema`?

annakhm: Shouldn't this and below be defined as `ForceNew`?

nsxt/resource_nsxt_policy_ods_runbook_invocation.go
```diff
@@ -0,0 +1,202 @@
...
+		argData["key"] = arg.Key
+		argData["value"] = arg.Value
+		argList = append(argList, argData)
+	}
+	d.Set("argument", argList)
+
+	return nil
+}
+
+func resourceNsxtPolicyODSRunbookInvocationUpdate(d *schema.ResourceData, m interface{}) error {
```
annakhm, line 182: If the user changes description, this will result in constant non-empty diff. Do tags/description have any value for this resource, perhaps we can remove them?

nsxt/resource_nsxt_policy_ods_runbook_invocation.go
ksamoray: For now, he can't. The runbook won't execute, it's a bug which they intend to fix obviously.

nsxt/resource_nsxt_policy_ods_runbook_invocation.go
ksamoray: Makes sense, if anyone ever needs them we can reevaluate how to handle these.

PR #1021, opened 2023-11-06
Add Principle Identity resource

wsquan171:
This PR adds support of Principle Identity resource. A principle identity is identified as a pair of PI `name` and `cert`. If an API is called with SSL client cert matching with a registered PI, the caller is recognized as the PI, granted with its roles and associated privileges.

Currently, the only non-deprecated API on NSX are:
- `createWithPem`: PI's cert pem is in-lined as part of PI creation call. The inlined pem has to be unique / new to NSX. NSX then creates a certificate in the background, and populates the MP cert id in `certificate_id` in response. This cert, however, will not be removed by NSX upon PI deletion.
- `updateCert`: If a PI's cert has expired, it can be updated by using this API call to replace the current `certificate_id` with another cert that is already in the system. The issue is:
  - NSX UI no longer shows MP API for cert management. This means if user imports their own cert on UI they have no way to find the underlying ID. Even if figured out, the cert belongs to internal PI `nsx_policy` and will cause trouble to be used for PI
  - our tf provider currently only supports cert as a data source
  - Other metadatas (`tags`, `display_name`, `description`) can also be updated with this API if and only if the `certificate_id` is changed and valid
- `delete`

As a result, PI is in general immutable. This is also reflected on NSX UI. In this implementation, I decided to:
- Make all fields `ForceNew`
- Leave out `display_name`, `description` and `revision`. As they're either not used or won't be reflected even on UI.
- Use `createWithPem` for PI creation and leave out `updateCert` to avoid res ownership issues and confusion. Cert updates are handled as PI recreation
- Deletes the NSX-imported pem when the PI is deleted.

Also made some changes to role binding resource to reuse some common arguments. Added TLS cert generation for test.

Principle Identity is also removed from role binding resource, as bindings created for PI is immutable from that API as well.

### DIFF
--- a/nsxt/resource_nsxt_policy_role_binding.go
+++ b/nsxt/resource_nsxt_policy_role_binding.go
@@ -6,7 +6,6 @@ package nsxt
 import (
 	"fmt"
 	"log"
-	"regexp"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
@@ -88,27 +87,12 @@ func getRolesForPathSchema(forceNew bool) *schema.Schema {
 					Description: "Path of the entity in parent hierarchy.",
 					Required:    true,
 				},
-				"role": {
-					Type:        schema.TypeList,
+				"roles": {
+					Type:        schema.TypeSet,
 					Description: "Applicable roles",
 					Required:    true,
-					Elem: &schema.Resource{
-						Schema: map[string]*schema.Schema{
-							"role": {
-								Type:        schema.TypeString,
-								Description: "Short identifier for the role",
-								Required:    true,
-								ValidateFunc: validation.StringMatch(
-									regexp.MustCompile(
-										`^[_a-z0-9-]+$`),
-									"Must be a valid role identifier matching: ^[_a-z0-9-]+$"),
-							},
-							"role_display_name": {
-								Type:        schema.TypeString,
-								Description: "Display name for role",
-								Computed:    true,
-							},
-						},
+					Elem: &schema.Schema{
+						Type: schema.TypeString,
 					},
 				},
 			},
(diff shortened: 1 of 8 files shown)


nsxt/resource_nsxt_principal_identity.go
```diff
@@ -0,0 +1,179 @@
...
+		return fmt.Errorf("error obtaining logical object id")
+	}
+	certID := d.Get("certificate_id").(string)
+	err := piClient.Delete(id)
+	if err != nil {
+		return handleDeleteError("PrincipalIdentity", id, err)
+	}
+
+	// Clean up underlying cert imported by NSX if exists
+	if len(certID) > 0 {
```
annakhm, line 170: sounds like the cert is created automatically by the create API, but not deleted nu delete API?

martinrohrbach:
@wsquan171 @annakhm I'm gonna be a little bold here but since I noticed this PR I wanted to ask if you could maybe also expose the "is_protected" boolean in this resource? We do use PIs quite a lot but at the same time set is_protected to false so we can still manipulate objects from the admin GUI if need be.

I believe this boolean cannot be updated after the PI user has been created. Our workaround is to recreate the PI user with the new setting, all objects that "belonged" to the old user are still working as intended afawcs.

nsxt/resource_nsxt_principal_identity.go
wsquan171: Right.

wsquan171:
> @wsquan171 @annakhm I'm gonna be a little bold here but since I noticed this PR I wanted to ask if you could maybe also expose the "is_protected" boolean in this resource? 

Sure. Wasn't aware of this use case and misread `is_protected` as `_protection` where the latter was a common read-only attr. Will add this.

nsxt/resource_nsxt_policy_role_binding.go
```diff
@@ -71,38 +69,44 @@ func resourceNsxtPolicyUserManagementRoleBinding() *schema.Resource {
...
+		ForceNew:    forceNew,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"path": {
+					Type:        schema.TypeString,
+					Description: "Path of the entity in parent hierarchy.",
+					Required:    true,
+				},
+				"role": {
+					Type:        schema.TypeList,
```
annakhm, line 92: I wonder if list of sub-clauses here is justified. Does role_display_name convey useful information?
If not, seems more user friendly to just offer a list of string roles? Unless we expect this object to grow in API

wsquan171: fixed

annakhm (approved):

PR #1025, opened 2023-11-09
Add FloodProtectionProfile support

### LINKED ISSUE #880: Add resource to manage the Distributed Firewall "Flood Protection Profiles"

We have several NSX-T instances, for which setting a hierarchy of several Flood Protection Profiles and maintain them is quite tiresome and tricky.

New resource to manage the Distributed Firewall General Setting "Flood Protection Profile".

Create the Flood Protection Profiles Manually.

### Additional context

_No response_

GraysonWu:
Fix #880


annakhm:
Is this object relevant for multitenancy use case?

ksamoray:
> Is this object relevant for multitenancy use case?

SDK has a multitenancy client [implementation](https://github.com/vmware/vsphere-automation-sdk-go/blob/master/services/nsxt/orgs/projects/infra/FloodProtectionProfilesClient.go) for this API, so it's worthwhile to implement this.
@GraysonWu LMK if you need help with this, generating a wrapper etc.

GraysonWu:
@annakhm @ksamoray Thanks for pointing it out. Added multi-tenancy support.

nsxt/resource_nsxt_policy_gateway_flood_protection_profile.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	if otherActiveConnLimit != 0 {
+		obj.OtherActiveConnLimit = &otherActiveConnLimit
+	}
+	if tcpHalfOpenConnLimit != 0 {
+		obj.TcpHalfOpenConnLimit = &tcpHalfOpenConnLimit
+	}
+	if udpActiveFlowLimit != 0 {
+		obj.UdpActiveFlowLimit = &udpActiveFlowLimit
+	}
```
salv-orlando, line 75: Minor comment (no need to address): I think it should be possible to have a common function to read/set attributes which are shared between the Gateway and Distributed profiles.

salv-orlando (commented):
I am unsure about whether we need some addiitional step to consume these profiles.
Do you think we need to implement also the FloodProtectionProfileBindingMap?

My question specifically is that in theory we can create multiple profiles. How do we know which profile should be applied to the firewall?

GraysonWu:
> I am unsure about whether we need some addiitional step to consume these profiles. Do you think we need to implement also the FloodProtectionProfileBindingMap?
> 
> My question specifically is that in theory we can create multiple profiles. How do we know which profile should be applied to the firewall?

Ah, good catch. We do need that binding map to create group-profile binding to apply the profile to groups. Let me add it.

ksamoray:
LGTM, pls rebase this though as there are conflicts.

ksamoray (approved):

annakhm:
Sorry I don't find the binding maps in this PR - should they be added as part of this PR?

GraysonWu:
> Sorry I don't find the binding maps in this PR - should they be added as part of this PR?

My bad. I thought I had pushed it...

nsxt/resource_nsxt_policy_gateway_flood_protection_profile_binding.go
```diff
@@ -0,0 +1,288 @@
...
+				Optional:      true,
+				ConflictsWith: []string{"context"},
+				ExactlyOneOf:  []string{"tier1_id"},
+			},
+			"tier1_id": {
+				Type:        schema.TypeString,
+				Description: "The ID of the tier1 gateway",
+				Optional:    true,
+			},
+			"locale_service_id": {
```
annakhm, line 55: May I suggest changing this to `parent_path`? This could be either T0 path, T1 path or locale service path.

GraysonWu: Sounds good. Done.

Issue #1002, opened 2023-10-17
Specify Inter VRF transit Subnet for T0 gateway


nickliuibm:
### Is your feature request related to a problem? Please describe.

We expect to use NSX provider to create T0 gateway by specifying 
- Internal Transit Subnet
- T0-T1 transit Subnet
- Inter VRF Transit Subent

We find the arguments [internal_transit_subnets](https://registry.terraform.io/providers/vmware/nsxt/latest/docs/resources/policy_tier0_gateway#internal_transit_subnets) and [transit_subnets](https://registry.terraform.io/providers/vmware/nsxt/latest/docs/resources/policy_tier0_gateway#transit_subnets) in resource [nsxt_policy_tier0_gateway](https://registry.terraform.io/providers/vmware/nsxt/latest/docs/resources/policy_tier0_gateway) for the first two. 
But for the last one `Inter VRF Transit Subent`, we don't see an argument for it. 
<img width="922" alt="image" src="https://github.com/vmware/terraform-provider-nsxt/assets/39811332/ed836f05-191a-486b-9a1a-272d88f28509">

Can you let use know if NSX provider support specifying  `Inter VRF transit Subnet`

### Describe the solution you'd like

An available argument in resource `nsxt_policy_tier0_gateway` to support  specifying  `Inter VRF transit Subnet`

nickliuibm:
Hell, Any update? Thanks!

salv-orlando:
Hello @nickliuibm  - sorry for our delay.
`vrf_transit_subnets` is an attribute which was added in NSX 4.1.0 and not yet covered by the provider.

We'll take care of ensuring the Tier-0 resource is updated to keep up with the latest NSX release.

salv-orlando:
Hello @nickliuibm  - I believe you are referring to tunnel setting which has been added to localeservice:
`/infra/tier-0s/<tier-0-id>/locale-services/<locale-svc-id>/tunnels/<tunnel-id>`

salv-orlando:
Hello @nickliuibm  - If I understand correctly your ask is to support the new feature for creating N/B GRE tunnels terminated at the T0. Created Issue #1027 for this.
We will scope the work and target it to one of the upcoming releases.

PR #1026, opened 2023-11-10
Add vrf_transit_subnets to nsxt_policy_tier0_gateway (#1002)

salv-orlando:
The vrf_transit_subnets attribute is supported for NSX>=4.1.0.

### DIFF
--- a/nsxt/resource_nsxt_policy_tier0_gateway.go
+++ b/nsxt/resource_nsxt_policy_tier0_gateway.go
@@ -109,6 +109,16 @@ func resourceNsxtPolicyTier0Gateway() *schema.Resource {
 				Computed: true,
 				ForceNew: true, // Modification of transit subnet not allowed after Tier-0 deployment
 			},
+			"vrf_transit_subnets": {
+				Type:        schema.TypeList,
+				Description: "VRF transit subnets in CIDR format",
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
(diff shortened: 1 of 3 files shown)


ksamoray (approved):
LGTM

annakhm:
I think the part where this new attribute is set in schema is missing (in `resourceNsxtPolicyTier0GatewayRead`)?

salv-orlando:
@annakhm can you have another look, please?

nsxt/resource_nsxt_policy_tier0_gateway.go
```diff
@@ -109,6 +109,17 @@ func resourceNsxtPolicyTier0Gateway() *schema.Resource {
 				Computed: true,
 				ForceNew: true, // Modification of transit subnet not allowed after Tier-0 deployment
 			},
+			"vrf_transit_subnets": {
+				Type:        schema.TypeList,
+				MaxItems:    1,
```
annakhm, line 114: I'm guessing we expect the API to allow more subnets in future (such as ipv6 support?)
In this case, I would remove the `MaxItems` here to allow native support for this in the future, and rely on NSX validation in the time being.
If we don't expect the list to grow, we should remove the list here as well, to avoid confusion.

salv-orlando: sounds good

annakhm (approved):

PR #1030, opened 2023-11-14
Implement nsxt_edge_transport_node data source

ksamoray:


nsxt/data_source_nsxt_transport_node.go
```diff
@@ -0,0 +1,100 @@
...
+
+func dataSourceNsxtTransportNode() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNsxtTransportNodeRead,
+
+		Schema: map[string]*schema.Schema{
+			"id":           getDataSourceIDSchema(),
+			"display_name": getDataSourceExtendedDisplayNameSchema(),
+			"description":  getDescriptionSchema(),
+			"node_type": {
```
annakhm, line 24: Would it make sense to filter by node type as well?

ksamoray: Would that produce a unique result, other than in testing environments?

annakhm: Seems to me like such filtering could be helpful, but we can wait for the ask

annakhm (approved):

nsxt/data_source_nsxt_transport_node.go
ksamoray: Maybe it's better to wait and see if we want to convert this into nsxt_edge_transport_node as discussed? Or maybe we change only the resource?

annakhm: I don't really have a preference, either is fine I think

PR #1036, opened 2023-11-20
Implement policy resources for lb monitor profile

2ez4szliu:
Implement several lb monitor profile resources using policy API including:
- lb_http_monitor
- lb_https_monitor
- lb_icmp_monitor
- lb_passive_monitor
- lb_tcp_monitor
- lb_udp_monitor

### DIFF
--- a/nsxt/lb_utils.go
+++ b/nsxt/lb_utils.go
@@ -13,6 +13,7 @@ import (
 	"github.com/vmware/go-vmware-nsxt/loadbalancer"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 )
 
 // Helpers for common LB monitor schema settings
@@ -267,6 +268,123 @@ func resourceNsxtPolicyLBAppProfileDelete(d *schema.ResourceData, m interface{})
 	if err != nil {
 		return handleDeleteError("LBAppProfile", id, err)
 	}
-
 	return nil
+}
+
+func resourceNsxtPolicyLBMonitorProfileExistsWrapper(id string, connector client.Connector, isGlobalManager bool) (bool, error) {
+	client := infra.NewLbMonitorProfilesClient(connector)
+	_, err := client.Get(id)
+	if err == nil {
+		return true, nil
+	}
+
+	if isNotFoundError(err) {
+		return false, nil
+	}
+	msg := fmt.Sprintf("Error retrieving resource LBMonitorProfile")
+	return false, logAPIError(msg, err)
+}
+
+func resourceNsxtPolicyLBMonitorProfileDelete(d *schema.ResourceData, m interface{}) error {
+	id := d.Id()
+	if id == "" {
+		return fmt.Errorf("Error obtaining LBMonitorProfile ID")
+	}
+	connector := getPolicyConnector(m)
+	forceParam := true
+	client := infra.NewLbMonitorProfilesClient(connector)
+	err := client.Delete(id, &forceParam)
+	if err != nil {
+		return handleDeleteError("LBMonitorProfile", id, err)
+	}
+	return nil
+}
+
+func getLbServerSslSchema() *schema.Schema {
+	return &schema.Schema{
+		Type:     schema.TypeList,
+		MaxItems: 1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"certificate_chain_depth": {
+					Type:        schema.TypeInt,
+					Optional:    true,
+					Description: "Authentication depth is used to set the verification depth in the server certificates chain. format: int64",
+				},
+				"client_certificate_path": getPolicyPathSchema(false, false, "Client certificate path"),
+				"server_auth": {
+					Type:         schema.TypeString,
+					Optional:     true,
+					ValidateFunc: validation.StringInSlice(lBServerSslProfileBindingServerAuthValues, false),
+					Description:  "Server authentication mode.",
+				},
+				"server_auth_ca_paths": {
+					Type: schema.TypeList,
+					Elem: &schema.Schema{
+						Type: schema.TypeString,
+					},
+					Optional:    true,
+					Description: "If server auth type is REQUIRED, server certificate must be signed by one of the trusted Certificate Authorities (CAs), also referred to as root CAs, whose self signed certificates are specified.",
+				},
+				"server_auth_crl_paths": {
+					Type: schema.TypeList,
+					Elem: &schema.Schema{
+						Type: schema.TypeString,
+					},
+					Optional:    true,
+					Description: "A Certificate Revocation List (CRL) can be specified in the server-side SSL profile binding to disallow compromised server certificates.",
+				},
+				"ssl_profile_path": getPolicyPathSchema(false, false, "SSL profile path"),
+			},
+		},
+		Optional: true,
+	}
+}
+
+func getLbServerSslFromSchema(d *schema.ResourceData) *model.LBServerSslProfileBinding {
+	serverSslList := d.Get("server_ssl").([]interface{})
+	var serverSsl *model.LBServerSslProfileBinding
+	for _, item := range serverSslList {
+		data := item.(map[string]interface{})
+		certificateChainDepth := int64(data["certificate_chain_depth"].(int))
+		clientCertificatePath := data["client_certificate_path"].(string)
+		serverAuth := data["server_auth"].(string)
+		serverAuthCaPathsList := data["server_auth_ca_paths"].([]interface{})
+		serverAuthCrlPathsList := data["server_auth_crl_paths"].([]interface{})
+		var serverAuthCaPaths []string
+		for _, path := range serverAuthCaPathsList {
+			serverAuthCaPaths = append(serverAuthCaPaths, path.(string))
+		}
+		var serverAuthCrlPaths []string
+		for _, path := range serverAuthCrlPathsList {
+			serverAuthCrlPaths = append(serverAuthCrlPaths, path.(string))
+		}
+		sslProfilePath := data["ssl_profile_path"].(string)
+		obj := model.LBServerSslProfileBinding{
+			CertificateChainDepth: &certificateChainDepth,
+			ClientCertificatePath: &clientCertificatePath,
+			ServerAuth:            &serverAuth,
+			ServerAuthCaPaths:     serverAuthCaPaths,
+			ServerAuthCrlPaths:    serverAuthCrlPaths,
+			SslProfilePath:        &sslProfilePath,
+		}
+		serverSsl = &obj
+	}
+	return serverSsl
+}
+
+func setLbServerSslInSchema(d *schema.ResourceData, lBServerSsl model.LBServerSslProfileBinding) {
+	var serverSslList []map[string]interface{}
+	elem := make(map[string]interface{})
+	elem["certificate_chain_depth"] = lBServerSsl.CertificateChainDepth
+	elem["client_certificate_path"] = lBServerSsl.ClientCertificatePath
+	elem["server_auth"] = lBServerSsl.ServerAuth
+	elem["ssl_profile_path"] = lBServerSsl.SslProfilePath
+	elem["server_auth_ca_paths"] = lBServerSsl.ServerAuthCaPaths
+	elem["server_auth_crl_paths"] = lBServerSsl.ServerAuthCrlPaths
+	serverSslList = append(serverSslList, elem)
+	err := d.Set("server_ssl", serverSslList)
+	if err != nil {
+		log.Printf("[WARNING] Failed to set server_ssl in schema: %v", err)
+	}
 }
--- a/nsxt/policy_utils.go
+++ b/nsxt/policy_utils.go
@@ -421,3 +421,38 @@ func getElemOrEmptyMapFromMap(d map[string]interface{}, key string) map[string]i
 	}
 	return make(map[string]interface{})
 }
+
+func setPolicyLbHTTPHeaderInSchema(d *schema.ResourceData, attrName string, headers []model.LbHttpRequestHeader) {
+	var headerList []map[string]string
+	for _, header := range headers {
+		elem := make(map[string]string)
+		elem["name"] = *header.HeaderName
+		elem["value"] = *header.HeaderValue
+		headerList = append(headerList, elem)
+	}
+	d.Set(attrName, headerList)
+}
+
+func getPolicyLbHTTPHeaderFromSchema(d *schema.ResourceData, attrName string) []model.LbHttpRequestHeader {
+	headers := d.Get(attrName).(*schema.Set).List()
+	var headerList []model.LbHttpRequestHeader
+	for _, header := range headers {
+		data := header.(map[string]interface{})
+		name := data["name"].(string)
+		value := data["value"].(string)
+		elem := model.LbHttpRequestHeader{
+			HeaderName:  &name,
+			HeaderValue: &value}
+
+		headerList = append(headerList, elem)
+	}
+	return headerList
+}
+
+func getPolicyLbMonitorPortSchema() *schema.Schema {
+	return &schema.Schema{
+		Type:        schema.TypeInt,
+		Description: "If the monitor port is specified, it would override pool member port setting for healthcheck. A port range is not supported",
+		Optional:    true,
+	}
+}
--- a/nsxt/resource_nsxt_policy_lb_http_monitor_profile.go
+++ b/nsxt/resource_nsxt_policy_lb_http_monitor_profile.go
@@ -1,0 +1,181 @@
+/* Copyright © 2023 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+)
+
+func resourceNsxtPolicyLBHttpMonitorProfile() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNsxtPolicyLBHttpMonitorProfileCreate,
+		Read:   resourceNsxtPolicyLBHttpMonitorProfileRead,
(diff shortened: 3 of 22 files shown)


website/docs/r/policy_lb_icmp_monitor_profile.html.markdown
```diff
@@ -0,0 +1,64 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `display_name` - (Required) Display name of the resource.
+* `description` - (Optional) Description of the resource.
+* `tag` - (Optional) A list of scope + tag pairs to associate with this resource.
+* `nsx_id` - (Optional) The NSX ID of this resource. If set, this ID will be used to create the resource.
+* `data_length` - (Optional) The data size (in byte) of the ICMP healthcheck packet format: int64
```
salv-orlando, line 38: nit: in bytes [plural]

nsxt/policy_utils.go
```diff
@@ -402,3 +402,38 @@ func getElemOrEmptyMapFromMap(d map[string]interface{}, key string) map[string]i
...
+		headerList = append(headerList, elem)
+	}
+	return headerList
+}
+
+func getPolicyLbMonitorPortSchema() *schema.Schema {
+	return &schema.Schema{
+		Type:        schema.TypeInt,
+		Description: "If the monitor port is specified, it would override pool member port setting for healthcheck. A port range is not supported",
+		Optional:    true,
```
annakhm, line 437: We can use `validateSinglePort()` for validation here

nsxt/resource_nsxt_policy_lb_http_monitor_profile.go
```diff
@@ -0,0 +1,207 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"nsx_id":                getNsxIDSchema(),
+			"path":                  getPathSchema(),
+			"display_name":          getDisplayNameSchema(),
+			"description":           getDescriptionSchema(),
+			"revision":              getRevisionSchema(),
+			"tag":                   getTagsSchema(),
+			"request_body":          getLbMonitorRequestBodySchema(),
```
annakhm, line 35: would it make sense to group `request` attributes in a sub-clause, and same for response?

nsxt/resource_nsxt_policy_lb_http_monitor_profile.go
```diff
@@ -0,0 +1,207 @@
...
+
+	client := infra.NewLbMonitorProfilesClient(connector)
+	obj, err := client.Get(id)
+	if err != nil {
+		return handleReadError(d, "LBHttpMonitorProfile", id, err)
+	}
+
+	baseObj, errs := converter.ConvertToGolang(obj, model.LBHttpMonitorProfileBindingType())
+	if len(errs) > 0 {
+		return fmt.Errorf("Error converting LBHttpMonitorProfile %s", errs[0])
```
annakhm, line 152: I think this error that would appear if we try to import monitor of the wrong type. Lets make it more verbose, something like "lb monitor with id xxx is not of http type"

nsxt/resource_nsxt_policy_lb_https_monitor_profile.go
```diff
@@ -0,0 +1,316 @@
...
+		MaxItems: 1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"certificate_chain_depth": {
+					Type:        schema.TypeInt,
+					Optional:    true,
+					Description: "Authentication depth is used to set the verification depth in the server certificates chain. format: int64",
+				},
+				"client_certificate_path": {
+					Type:        schema.TypeString,
```
annakhm, line 71: Lets use `getPolicyPathSchema`

nsxt/resource_nsxt_policy_lb_https_monitor_profile.go
```diff
@@ -0,0 +1,316 @@
...
+				"server_auth": {
+					Type:         schema.TypeString,
+					Optional:     true,
+					ValidateFunc: validation.StringInSlice(lBServerSslProfileBindingServerAuthValues, false),
+					Description:  "Server authentication mode.",
+				},
+				"server_auth_ca_paths": {
+					Type: schema.TypeList,
+					Elem: &schema.Schema{
+						Type: schema.TypeString,
```
annakhm, line 84: `getPolicyPathSchema`

nsxt/resource_nsxt_policy_lb_https_monitor_profile.go
```diff
@@ -0,0 +1,316 @@
...
+					Optional:    true,
+					Description: "Server SSL profile defines reusable, application-independent server side SSL properties.",
+				},
+			},
+		},
+		Optional: true,
+	}
+}
+
+func resourceNsxtPolicyLBHttpsMonitorProfileExists(id string, connector client.Connector, isGlobalManager bool) (bool, error) {
```
annakhm, line 108: I think this function can be reused for all monitors?

nsxt/resource_nsxt_policy_lb_http_monitor_profile.go
```diff
@@ -0,0 +1,207 @@
...
+
+	err := resourceNsxtPolicyLBHttpMonitorProfilePatch(d, m, id)
+	if err != nil {
+		return handleUpdateError("LBHttpMonitorProfile", id, err)
+	}
+
+	return resourceNsxtPolicyLBHttpMonitorProfileRead(d, m)
+}
+
+func resourceNsxtPolicyLBHttpMonitorProfileDelete(d *schema.ResourceData, m interface{}) error {
```
annakhm, line 192: Seems like all *Delete functions can be reused across all monitors

nsxt/resource_nsxt_policy_lb_http_monitor_profile.go
```diff
@@ -0,0 +1,207 @@
...
+)
+
+func resourceNsxtPolicyLBHttpMonitorProfile() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNsxtPolicyLBHttpMonitorProfileCreate,
+		Read:   resourceNsxtPolicyLBHttpMonitorProfileRead,
+		Update: resourceNsxtPolicyLBHttpMonitorProfileUpdate,
+		Delete: resourceNsxtPolicyLBHttpMonitorProfileDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
```
annakhm, line 25: Lets use `nsxtPolicyPathResourceImporter` for consistency

website/docs/r/policy_lb_https_monitor_profile.html.markdown
```diff
@@ -0,0 +1,84 @@
...
+```hcl
+resource "nsxt_policy_lb_https_monitor_profile" "test" {
+  display_name          = "test"
+  description           = "Terraform provisioned LBHttpsMonitorProfile"
+  request_body          = "test"
+  request_method        = "POST"
+  request_url           = "test"
+  request_version       = "HTTP_VERSION_1_1"
+  response_body         = "test"
+  response_status_codes = 2
```
annakhm, line 25: I believe this should be a list, and it would be helpful to show values that make sense - 200 here, and a reasonable port value below

nsxt/resource_nsxt_policy_lb_https_monitor_profile_test.go
```diff
@@ -0,0 +1,258 @@
...
+	"header_value":            "test-update",
+	"certificate_chain_depth": "5",
+	"server_auth":             "test-update",
+}
+
+func TestAccResourceNsxtPolicyLBHttpsMonitorProfile_basic(t *testing.T) {
+	testResourceName := "nsxt_policy_lb_https_monitor_profile.test"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
```
annakhm, line 58: I don't think there is support for global manager, so we should add `testAccOnlyLocalManager(t)` here

website/docs/r/policy_lb_http_monitor_profile.html.markdown
```diff
@@ -0,0 +1,77 @@
...
+layout: "nsxt"
+page_title: "NSXT: nsxt_policy_lb_http_monitor_profile"
+description: A resource to configure a LBHttpMonitorProfile.
+---
+
+# nsxt_policy_lb_http_monitor_profile
+
+This resource provides a method for the management of a LBHttpMonitorProfile.
+
+This resource is applicable to NSX Global Manager, NSX Policy Manager and VMC.
```
annakhm, line 12: No global manager, and not certain about VMC

nsxt/policy_utils.go
2ez4szliu: I noticed that when validateSinglePort() is used, the port in Schema is a string, but here the schema for port is Int

nsxt/lb_utils.go
```diff
@@ -237,3 +240,123 @@ func resourceNsxtLbMonitorDelete(d *schema.ResourceData, m interface{}) error {
 	}
 	return nil
 }
+
+func resourceNsxtPolicyLBMonitorProfileExists(id string, connector client.Connector, isGlobalManager bool, resourceType string) (bool, error) {
```
annakhm, line 244: Is `isGlobalManager` seems redundant here, global manager is not supported.
Also my personal preference would be to define a single function for all *Exists* and *Delete* implementations for LB monitor, since `resourceType` is only used for printing errors. For printing purposes, this ca be just LbMonitor, without specifying exact type. But I don't insist :)

2ez4szliu: Makes sense, I have removed these 2 parameters from the function

nsxt/resource_nsxt_policy_lb_icmp_monitor_profile.go
```diff
@@ -0,0 +1,164 @@
...
+			"revision":     getRevisionSchema(),
+			"tag":          getTagsSchema(),
+			"data_length": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "The data size (in byte) of the ICMP healthcheck packet",
+			},
+			"fall_count":   getLbMonitorFallCountSchema(),
+			"interval":     getLbMonitorIntervalSchema(),
+			"monitor_port": getPolicyLbMonitorPortSchema(),
```
annakhm, line 42: Does port make sense for icmp monitor? Docs state that for icmp, port is not required.

website/docs/r/policy_lb_http_monitor_profile.html.markdown
```diff
@@ -0,0 +1,77 @@
...
+* `response_status_codes` - (Optional) The HTTP response status code should be a valid HTTP status code.
+* `fall_count` - (Optional) Mark member status DOWN if the healtcheck fails consecutively for fall_count times.
+* `interval` - (Optional) Active healthchecks are initiated periodically, at a configurable interval (in seconds), to each member of the Group.
+* `monitor_port` - (Optional) Typically, monitors perform healthchecks to Group members using the member IP address and pool_port. However, in some cases, customers prefer to run healthchecks against a different port than the pool member port which handles actual application traffic. In such cases, the port to run healthchecks against can be specified in the monitor_port value. For ICMP monitor, monitor_port is not required.
```
annakhm, line 54: The sentence in the end about ICMP monitor should be removed from other types of monitors, as its not applicable to them

website/docs/r/policy_lb_http_monitor_profile.html.markdown
```diff
@@ -0,0 +1,77 @@
...
+* `request_body` - (Optional) String to send as part of HTTP health check request body. Valid only for certain HTTP methods like POST.
+* `request_header` - (Optional) Array of HTTP request headers.
+  * `header_name` - (Optional) Name of HTTP request header
+  * `header_value` - (Optional) Value of HTTP request header
+* `request_method` - (Optional) Possible values are: `GET`, `OPTIONS`, `POST`, `HEAD`, `PUT`.
+* `request_url` - (Optional) For HTTP active healthchecks, the HTTP request url sent can be customized and can include query parameters.
+* `request_version` - (Optional) Possible values are: `HTTP_VERSION_1_0`, `HTTP_VERSION_1_1`, HTTP request version.
+* `response_body` - (Optional) If HTTP response body match string (regular expressions not supported) is specified (using LBHttpMonitor.response_body) then the healthcheck HTTP response body is matched against the specified string and server is considered healthy only if there is a match. If the response body string is not specified, HTTP healthcheck is considered successful if the HTTP response status code is 2xx, but it can be configured to accept other status codes as successful.
```
annakhm, line 50: Lets remove the `(using LBHttpMonitor.response_body)` part

nsxt/resource_nsxt_policy_lb_https_monitor_profile.go
```diff
@@ -0,0 +1,194 @@
...
+			"revision":                   getRevisionSchema(),
+			"tag":                        getTagsSchema(),
+			"request_body":               getLbMonitorRequestBodySchema(),
+			"request_header":             getLbHTTPHeaderSchema("Array of HTTP request headers"),
+			"request_method":             getLbMonitorRequestMethodSchema(),
+			"request_url":                getLbMonitorRequestURLSchema(),
+			"request_version":            getLbMonitorRequestVersionSchema(),
+			"response_body":              getLbMonitorResponseBodySchema(),
+			"response_status_codes":      getLbMonitorResponseStatusCodesSchema(),
+			"server_ssl_profile_binding": getLbServerSslProfileBindingSchema(),
```
annakhm, line 48: I suggest to call this `server_ssl` for consistency with other LB resources (f.e. virtual server)

nsxt/resource_nsxt_policy_lb_http_monitor_profile.go
```diff
@@ -0,0 +1,184 @@
...
+			"fall_count":            getLbMonitorFallCountSchema(),
+			"interval":              getLbMonitorIntervalSchema(),
+			"rise_count":            getLbMonitorRiseCountSchema(),
+			"timeout":               getLbMonitorTimeoutSchema(),
+			"monitor_port":          getPolicyLbMonitorPortSchema(),
+		},
+	}
+}
+
+func resourceNsxtPolicyLBHttpMonitorProfileExists(id string, connector client.Connector, isGlobalManager bool) (bool, error) {
```
annakhm, line 51: Can we make this function common for all profile types as well?

nsxt/resource_nsxt_policy_lb_icmp_monitor_profile.go
```diff
@@ -0,0 +1,190 @@
...
+			"nsx_id":       getNsxIDSchema(),
+			"path":         getPathSchema(),
+			"display_name": getDisplayNameSchema(),
+			"description":  getDescriptionSchema(),
+			"revision":     getRevisionSchema(),
+			"tag":          getTagsSchema(),
+			"data_length": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "The data size (in byte) of the ICMP healthcheck packet format: int64",
```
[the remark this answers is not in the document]
2ez4szliu: I think for all in64 parameters from SDK we convert them to Terraform TypeInt

nsxt/resource_nsxt_policy_lb_icmp_monitor_profile.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func resourceNsxtPolicyLBIcmpMonitorProfilePatch(d *schema.ResourceData, m interface{}, id string) error {
+	connector := getPolicyConnector(m)
+	converter := bindings.NewTypeConverter()
+
+	displayName := d.Get("display_name").(string)
+	description := d.Get("description").(string)
+	tags := getPolicyTagsFromSchema(d)
+	dataLength := int64(d.Get("data_length").(int))
```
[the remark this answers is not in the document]
2ez4szliu: Seems SDK use int64 for all int parameters...
I noticed that in the SDK, the comment next to some parameters says that this parameter is of format int32, however in the SDK code the parameter is still defined as an int64.
For example, Revision
```
// The _revision property describes the current revision of the resource. To prevent clients from overwriting each other's changes, PUT operations must include the current _revision of the resource, which clients should obtain by issuing a GET operation. If the _revision provided in a PUT request is missing or stale, the operation will be rejected. format: int32
Revision *int64
```

nsxt/resource_nsxt_policy_lb_icmp_monitor_profile.go
2ez4szliu: I think it is not required but can be configured via API, on NSX UI we cannot specify port for the Icmp monitor, but we can do this via API. Not sure if we want to keep it or not?

annakhm: I would remove it as it seems inapplicable for icmp

nsxt/resource_nsxt_policy_lb_http_monitor_profile.go
2ez4szliu: I think it's already calling the function resourceNsxtPolicyLBMonitorProfileExists which is common for all profile types?

nsxt/resource_nsxt_policy_lb_http_monitor_profile.go
2ez4szliu: Not sure if it's necessary since we've being using this format for the corresponding non-policy resource?

nsxt/resource_nsxt_policy_lb_http_monitor_profile.go
annakhm: Yes, but rather than having 6 wrapper functions replicas for each monitor type, we can have just one - we can call it `resourceNsxtPolicyLBMonitorProfileExistsWrapper` or similar

nsxt/resource_nsxt_policy_lb_http_monitor_profile.go
annakhm: Good point, if we implement resource migration in the future, it would be easier to keep similar format

nsxt/resource_nsxt_policy_lb_http_monitor_profile.go
2ez4szliu: Done. Now we only use one wrapper function `resourceNsxtPolicyLBMonitorProfileExistsWrapper` across all monitor types

website/docs/r/policy_lb_icmp_monitor_profile.html.markdown
```diff
@@ -0,0 +1,64 @@
...
+## Example Usage
+
+```hcl
+resource "nsxt_policy_lb_icmp_monitor_profile" "test" {
+  display_name = "test"
+  description  = "Terraform provisioned LBIcmpMonitorProfile"
+  data_length  = 2
+  fall_count   = 2
+  interval     = 2
+  monitor_port = 8080
```
annakhm, line 23: `monitor_port` needs to be removed for `icmp` profile, here and below in attribute description. Other than that LGTM, thanks

2ez4szliu: Thanks Anna, monitor port has been removed

annakhm (approved):

Issue #757, opened 2022-05-19
Add resource to manage the Distributed Firewall settings "Exclusion list"


harishkalyanpur:
### Is your feature request related to a problem? Please describe.

We cannot update distributed firewall exclusion list with terraform at the moment. This requires us to manually update.

### Describe the solution you'd like

New resource to manage the Distributed Firewall Settings "Exclusion List".

### Describe alternatives you've considered

Manually updating the exclusion list.

https://docs.vmware.com/en/VMware-Cloud-on-AWS/services/com.vmware.vmc-aws-networking-security/GUID-F6012765-7969-4ACC-92E0-CE716AAF92CE.html 

We are not sure If there is an internal API call available for the same.

Girdhar1982:
any update on this request.. is this functionality available?

vitality411:
I am also interested in this feature.

PR #1037, opened 2023-11-22
Implement DFW settings Exclusion list

ksamoray:
Fixes: #757


nsxt/resource_nsxt_policy_firewall_exclude_list_member.go
```diff
@@ -0,0 +1,127 @@
...
+func resourceNsxtPolicyFirewallExcludeListMember() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNsxtPolicyFirewallExcludeListMemberCreate,
+		Read:   resourceNsxtPolicyFirewallExcludeListMemberRead,
+		Delete: resourceNsxtPolicyFirewallExcludeListMemberDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
+			"member": {
```
annakhm, line 25: Would it make more sense to implement the list as one resource (with list of paths inside?)

ksamoray: We've considered - But then we'll run into the usual issue where this singleton NSX object is virtually indestructible using NSX API - TF doesn't really like these, I think. Is there any decent way to work around this?

ksamoray: We can - the question is if there is any advantage in handling this as a scalar, vs handling as a list which is more "natural" to terraform with full CRUD operation.
We've done this for context profile custom attributes and such.

ksamoray:
test-all

nsxt/resource_nsxt_policy_firewall_exclude_list_member.go
annakhm: My personal feeling is that a single string does not justify a dedicated resource. But I don't insist.
As long as we're sure order is not important in this list, we can do it as suggested here

annakhm (approved):

nsxt/resource_nsxt_policy_firewall_exclude_list_member.go
ksamoray: @annakhm there is a more serious problem which I should look into:
When two members are created at the same time, there is a concurrency problem - both retrieve the member list, append the new member and when submitting one change is overwritten.
I'm not sure if we have some mean to handle this (and we might have this elsewhere maybe).

ksamoray: I worked around the concurrency problem, and added a few test. I'd rather if you could have another look...

nsxt/resource_nsxt_policy_firewall_exclude_list_member.go
```diff
@@ -0,0 +1,144 @@
...
+func resourceNsxtPolicyFirewallExcludeListMemberDelete(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+	member := d.Get("member").(string)
+
+	doUpdate := func() error {
+		var obj model.PolicyExcludeList
+
+		client := security.NewExcludeListClient(getSessionContext(d, m), connector)
+		obj, err := client.Get()
+		if isNotFoundError(err) {
```
annakhm, line 124: We should silently swallow the error in `not found` case

annakhm:
I still feel like a singleton resource is better design, as it would save roundtrips and avoid concurrency problems. But I don't insist.

ksamoray:
> I still feel like a singleton resource is better design, as it would save roundtrips and avoid concurrency problems. But I don't insist.

IMO a singleton doesn't make sense in the API level as well, from the functional aspect, e.g:
- Why should this resource be managed from a single point?
- Should a user be able to create dependency between a exclude list member and other resources?
- Why would we confine an organization to maintain the exclude list from a single interface (UI, TF...)? As otherwise concurrency issues become worse.

salv-orlando:
@ksamoray shouldn't this PR also have doc changes since we are adding a new resource?

Regarding the singleton vs non-singleton use case, I think both @annakhm  and @ksamoray have good points.
I'd tend to slightly prefer user experience over scalability, especially considering we should not see hundreds or thousands of groups in the exclusion list, so I'd say maybe concurrency is not a big concern.

ksamoray:
@salv-orlando yeah we need documentation but no point in adding that before we concluded the implementation.

annakhm (approved):

Issue #845, opened 2023-03-01
nsxt_policy_lb_pool to support a list for active_monitor_path 


declan-morris:
### Is your feature request related to a problem? Please describe.

nsxt_policy_lb_pool can have an active monitor attached to it. NSXT allows you to add multiple active monitors in the UI but this doesn't appear to be supported via terraform. Adding one via terraform allows management of the first one in the list, more can be added in the NSXT UI and terraform won't see that change.

### Describe the solution you'd like

I think the active_monitor_path should support a list of string for the different paths for a list of active monitors

### Describe alternatives you've considered

Separate out the virtual server to have one for each port and use a check on each virtual server

PR #1039, opened 2023-11-22
Policy LB pool: introduce active_monitor_paths attribute (#845)

salv-orlando:
The active_monitor_path supports a single attribute whereas the NSX API allows for multiple monitor paths to be specified.

This PR introduces a list attribute called active_monitor_paths and deprecates active_monitor_path. The two attributes are mutually exclusive.


nsxt/resource_nsxt_policy_lb_pool_test.go
```diff
@@ -26,6 +27,7 @@ var accTestPolicyLBPoolUpdateAttributes = map[string]string{
 	"algorithm":               "WEIGHTED_ROUND_ROBIN",
 	"min_active_members":      "5",
 	"tcp_multiplexing_number": "5",
+	"active_monitor_path":     "/infra/lb-monitor-profiles/default-http-lb-monitor",
```
annakhm, line 30: Lets also validate that this attribute is assigned within the test (validating that list size `active_monitor_paths.#` equals to 1)

nsxt/resource_nsxt_policy_lb_pool.go
```diff
@@ -491,17 +501,22 @@ func resourceNsxtPolicyLBPoolRead(d *schema.ResourceData, m interface{}) error {
 		return handleReadError(d, "LBPool", id, err)
 	}
 
+	curState := d.State()
+	if curState.Attributes["active_monitor_path"] != "" {
```
annakhm, line 505: We don't expect this attribute to be set in state during the first read (which is the case with create and import).
Deprecation is a bit tricky within the provider since two attributes affectively control the same setting.
I would suggest to check if the user specified the new attribute in intent, and if so, nullify the deprecated one, and vice versa.
If none is specified (potential import scenario), set the new attribute. We can add a note in documentation, stressing out that for new resources (which includes the import use case) only new attribute must be used, otherwise inconsistency may occur.

nsxt/resource_nsxt_policy_lb_pool_test.go
salv-orlando: right, I forgot to do that

nsxt/resource_nsxt_policy_lb_pool.go
```diff
@@ -491,17 +501,22 @@ func resourceNsxtPolicyLBPoolRead(d *schema.ResourceData, m interface{}) error {
 		return handleReadError(d, "LBPool", id, err)
 	}
 
+	if _, ok := d.GetOk("active_monitor_path"); ok {
+		// deprecated value was set
+		if obj.ActiveMonitorPaths != nil {
+			d.Set("active_monitor_path", obj.ActiveMonitorPaths[0])
+		}
+	} else {
+		// resource uses lists attribute (non-deprecated) or attribute
```
annakhm, line 510: if `active_monitor_path` was set before, and was deleted now, would it be nullified in the state?

salv-orlando: I hope I understand correctly your question.
I think the answer is yes, but I'm not sure if you want the attribute to be cleared or removed from the state.
The attribute is cleared, and I've also added a check in acceptance tests to verify that.

Resource change: 

```
  # nsxt_policy_lb_pool.test will be updated in-place
  ~ resource "nsxt_policy_lb_pool" "test" {
      - active_monitor_path      = "/infra/lb-monitor-profiles/default-http-lb-monitor" -> null
      + active_monitor_paths     = [
          + "/infra/lb-monitor-profiles/default-http-lb-monitor",
          + "/infra/lb-monitor-profiles/default-icmp-lb-monitor",
        ]
        id                       = "c756df01-a22f-48e9-95a3-4dd0d9e7cfd9"
        # (9 unchanged attributes hidden)

        # (4 unchanged blocks hidden)
    }
```

State after change:

```     
"instances": [
        {
          "schema_version": 0,
          "attributes": {
            "active_monitor_path": "",
            "active_monitor_paths": [
              "/infra/lb-monitor-profiles/default-http-lb-monitor",
              "/infra/lb-monitor-profiles/default-icmp-lb-monitor"
            ],
            "algorithm": "IP_HASH",
            "description": "yyy",
            "display_name": "xxx",
[...]
```

annakhm (approved):

PR #1046, opened 2023-11-30
Change password for BGP neighbor only on demand

annakhm:
Since password is not sent back from NSX for security reasons, it will not be present in state after import.
In this scenario, we need to avoid the password from being cleared next time an update is triggered on the imported resource. Hence we only update the password if a change in this field s detected. Note that user can leave the password blank in configuration of the imported resource.


salv-orlando (approved):
LGTM!

PR #1056, opened 2023-12-06
Separate security policy and rule

### LINKED ISSUE #777: Splitting nsxt_policy_security_policy and security_rule in two disctincs resources

Current resource structure is not flexible to iterate on firewall rule in a specific security policy. 

Current configuration is : 

```
resource "nsxt_policy_security_policy" "policy1" {
  display_name = "policy1"
  description  = "Terraform provisioned Security Policy"
  category     = "Application"
  locked       = false
  stateful     = true
  tcp_strict   = false
  scope        = [nsxt_policy_group.pets.path]

  rule {
    display_name       = "block_icmp"
    destination_groups = [nsxt_policy_group.cats.path, nsxt_policy_group.dogs.path]
    action             = "DROP"
    services           = [nsxt_policy_service.icmp.path]
    logged             = true
  }
}
```

A dedicated firewall_rule resource could permit us to use Terraform Meta-argument for_each or count.

Possible configuration : 

```
resource "nsxt_policy_security_policy" "policy1" {
  display_name = "policy1"
  description  = "Terraform provisioned Security Policy"
  category     = "Application"
  locked       = false
  stateful     = true
  tcp_strict   = false
  scope        = [nsxt_policy_group.pets.path]
}

resource "nsxt_policy_security_rule" "rule1" {
    **security_policy = nsxt_policy_security_policy.policy1.id**
    display_name       = "block_icmp"
    destination_groups = [nsxt_policy_group.cats.path, nsxt_policy_group.dogs.path]
    action             = "DROP"
    services           = [nsxt_policy_service.icmp.path]
    logged             = true
}
```

Dynamic object can be used to achieve dynamic rule creation. However, passing argument to such module cloud be tricky and need to use experimental terraform feature.

### Additional context

_No response_

### LINKED ISSUE #1043: Importing T1 information failing on VMC-NSX-T
### Describe the bug

Hi There, 
We have pre-created T1’s, the CGW and the MGW. They are pre-created with some default rules to allow outbound internet access. These rules need to be imported into our Terraform code.  We are getting import errors and appreciate some help on resolving those.

### Reproduction steps

1. created the resource
2. MGW definition 
resource "nsxt_policy_gateway_policy" "mgw_policy" {
  category     = "LocalGatewayRules"
  display_name = "default"
  domain       = "mgw"
}

CGW definition
resource "nsxt_policy_gateway_policy" "cgw_policy" {
  category     = "LocalGatewayRules"
  display_name = "default"
  domain       = "cgw"
}

2) Issue the command bellow 
 terraform import nsxt_policy_gateway_policy.cgw_policy cgw/default

terraform import nsxt_policy_gateway_policy.mgw_policy mgw/default

3.Getting the error bellow due to the resource we created above. Can we get some help here ? what am I missing here ? 
sxt_policy_gateway_policy.cgw_policy: Import prepared!
  Prepared nsxt_policy_gateway_policy for import
data.nsxt_policy_security_policy.predefined: Read complete after 0s [id=default-layer3-section]
╷
│ Error: Resource already managed by Terraform
│ 
│ Terraform is already managing a remote object for nsxt_policy_gateway_policy.cgw_policy. To import to this address you must first remove the existing object from the state.

...

### Expected behavior

Import to be ok and I can get those into my tf state file

### Additional context

_No response_

GraysonWu:
Fixes #777 
Fixes #1043


GraysonWu:
I'm wondering if there is a client for Rule. If yes, then some codes could be optimized.

annakhm:
Hi @GraysonWu, I don't see a use case for rule data source - can you suggest where it might be needed?

nsxt/policy_common.go
```diff
@@ -282,19 +288,28 @@ func getSecurityPolicyAndGatewayRulesSchema(scopeRequired bool, isIds bool, nsxI
...
-		},
+	if separated {
+		ruleSchema["policy_path"] = getPolicyPathSchema(true, false, "Security Policy path")
+		ruleSchema["sequence_number"] = &schema.Schema{
+			Type:        schema.TypeInt,
+			Description: "Sequence number of the this rule",
+			Required:    true,
+		}
+		ruleSchema["context"] = getContextSchema()
+		ruleSchema["path"] = getPathSchema()
```
annakhm, line 299: I think `path` is useful for both cases

nsxt/policy_common.go
```diff
@@ -282,19 +288,28 @@ func getSecurityPolicyAndGatewayRulesSchema(scopeRequired bool, isIds bool, nsxI
...
-			Schema: ruleSchema,
-		},
+	if separated {
+		ruleSchema["policy_path"] = getPolicyPathSchema(true, false, "Security Policy path")
+		ruleSchema["sequence_number"] = &schema.Schema{
+			Type:        schema.TypeInt,
+			Description: "Sequence number of the this rule",
+			Required:    true,
+		}
+		ruleSchema["context"] = getContextSchema()
```
annakhm, line 298: It makes sense to specify the context for sake of consistency, but its also true that context can be derived from the policy object. Not sure what is the best approach here - specifying the context for hundreds of rules can be an overhead for the user

2ez4szliu:
> Hi @GraysonWu, I don't see a use case for rule data source - can you suggest where it might be needed?

I think it might be related to Issue 1043 https://github.com/vmware/terraform-provider-nsxt/issues/1043
But seems a resource for rules can already satisfy the user's need.

nsxt/resource_nsxt_policy_security_policy_no_rule.go
```diff
@@ -0,0 +1,149 @@
...
+		Delete: resourceNsxtPolicySecurityPolicyNoRuleDelete,
+		Importer: &schema.ResourceImporter{
+			State: nsxtDomainResourceImporter,
+		},
+		Schema: getPolicySecurityPolicySchema(false, true, false),
+	}
+}
+
+func resourceNsxtPolicySecurityPolicyNoRuleCreate(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
```
annakhm, line 30: would it make sense to parametrize the CRUD to functions to either include or exclude the rule, in order to reuse the common code with the other policy resource?

nsxt/provider.go
```diff
@@ -437,6 +438,8 @@ func Provider() *schema.Provider {
 			"nsxt_edge_high_availability_profile":          resourceNsxtEdgeHighAvailabilityProfile(),
 			"nsxt_policy_host_transport_node_collection":   resourceNsxtPolicyHostTransportNodeCollection(),
 			"nsxt_policy_lb_client_ssl_profile":            resourceNsxtPolicyLBClientSslProfile(),
+			"nsxt_policy_security_policy_rule":             resourceNsxtPolicySecurityPolicyRule(),
+			"nsxt_policy_security_policy_no_rule":          resourceNsxtPolicySecurityPolicyNoRule(),
```
annakhm, line 442: I feel that `nsxt_policy_security_policy_no_rule` is confusion, especially next to `nsxt_policy_security_policy_rule`.
I don't have a much better option, but lets not give up on better naming yet :)
`nsxt_policy_empty_security_policy`? `nsxt_policy_parent_security_policy`?
There is also an option to remove the first `policy` part, since non-policy resources are deprecated by now.
This would result in `nsxt_security_policy` and `nsxt_security_policy_rule`.

nsxt/policy_common.go
GraysonWu: Done.

nsxt/provider.go
GraysonWu: Changed to nsxt_policy_parent_security_policy. Ideas are welcome!

nsxt/policy_common.go
GraysonWu: Changed. If users didn't provide the context, then we will extract the `project_id` from `policy_path` and fill the `context`.

GraysonWu:
> Hi @GraysonWu, I don't see a use case for rule data source - can you suggest where it might be needed?

Not for now. As @ 2ez4szliu said, a data source might be needed for issue #1043. Considering a resource should be enough, let me remove it.

nsxt/resource_nsxt_policy_security_policy_no_rule.go
GraysonWu: It could make sense but I think it might deserve a separate PR to do so, including switch from using H-API to using  security policy API for `nsxt_policy_security_policy`

GraysonWu: I just noticed that securityPolicy's API couldn't delete rules. So to let `parent_security_policy` and `security_policy` share the common code I changed to use H-API for the `parent_security_policy` with the latest commit. It will make more sense to conduct in this PR instead of another one.
And, as for the other policy resource, did you mean `gateway_policy` and `predefined_security_poilcy` as well? From my understanding, `predefined_security_policy` has a different schema than `security_policy`. And `gateway_policy`, although sharing most of the schema, is a different model. I'm not sure merging those two resources could share much common code. Seems making them self-contained is a better option.
Glad to hear your thoughts!

annakhm: Could you please clarify what you mean by lack of support for deleting rules from the API - I see [this](https://github.com/vmware/vsphere-automation-sdk-go/blob/master/services/nsxt/infra/domains/security_policies/RulesClient.go#L34) in the SDK. 
Regardless, follow up PR for optimizing's the code is fine by me!

nsxt/policy_common.go
```diff
@@ -291,19 +298,27 @@ func getSecurityPolicyAndGatewayRulesSchema(scopeRequired bool, isIds bool, nsxI
...
-			Schema: ruleSchema,
-		},
+	if separated {
+		ruleSchema["policy_path"] = getPolicyPathSchema(true, false, "Security Policy path")
+		ruleSchema["sequence_number"] = &schema.Schema{
+			Type:        schema.TypeInt,
+			Description: "Sequence number of the this rule",
+			Required:    true,
+		}
+		ruleSchema["context"] = getComputedContextSchema()
```
annakhm, line 308: Could you please add a comment here explaining why we are handling `context` here differently (context is required for consistency and can be derived from parent object)

nsxt/resource_nsxt_policy_security_policy_no_rule.go
GraysonWu: Sorry for the confusion. 
I mean I couldn't use the security policy [client](https://github.com/vmware/vsphere-automation-sdk-go/blob/master/services/nsxt/infra/domains/SecurityPoliciesClient.go) to delete the rule belonging to it. If we use the SDK for `Rule` to manage `security_policy`, not the parent one, we have to make several additional API calls for the rules, instead of one API call as H-API does. I'm a little bit concerned about the performance.

nsxt/resource_nsxt_policy_parent_security_policy_test.go
```diff
@@ -0,0 +1,167 @@
...
+	return testAccNsxtPolicySecurityPolicyDeps() + fmt.Sprintf(`
+resource "nsxt_policy_parent_security_policy" "test" {
+%s
+  display_name    = "%s"
+  description     = "Acceptance Test"
+  domain = "default"
+  category        = "Application"
+  locked          = %s
+  sequence_number = %s
+  stateful        = "true"
```
annakhm, line 156: boolean values don't require quotation marks

nsxt/resource_nsxt_policy_parent_security_policy_test.go
```diff
@@ -0,0 +1,167 @@
...
+  tcp_strict      = %s
+  scope           = [nsxt_policy_group.group1.path]
+
+  tag {
+    scope = "color"
+    tag   = "orange"
+  }
+
+  depends_on = [nsxt_policy_group.group1]
```
annakhm, line 165: there is an implicit dependency for the group, so explicit one shouldn't be required?

nsxt/resource_nsxt_policy_security_policy.go
```diff
@@ -96,28 +70,58 @@ func policySecurityPolicyBuildAndPatch(d *schema.ResourceData, m interface{}, co
 		obj.Revision = &revision
```
annakhm, line 70: we shouldn't set revision for create flow even if rules are present

website/docs/r/policy_security_policy_rule.html.markdown
```diff
@@ -0,0 +1,98 @@
...
+  display_name       = "rule1"
+  description        = "Terraform provisioned Security Policy Rule"
+  policy_path        = nsxt_policy_parent_security_policy.policy1.path
+  sequence_number    = 1
+  destination_groups = [nsxt_policy_group.cats.path, nsxt_policy_group.dogs.path]
+  action             = "DROP"
+  services           = [nsxt_policy_service.icmp.path]
+  logged             = true
+
+  depends_on = [nsxt_policy_parent_security_policy.policy1]
```
annakhm, line 31: as there is implicit dependency, explicit one shouldn't be needed

website/docs/r/policy_security_policy_rule.html.markdown
```diff
@@ -0,0 +1,98 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `display_name` - (Required) Display name of the resource.
+* `description` - (Optional) Description of the resource.
+* `tag` - (Optional) A list of scope + tag pairs to associate with this policy.
+* `nsx_id` - (Optional) The NSX ID of this resource. If set, this ID will be used to create the resource.
+* `policy_path` - (Required) The path of the Security Policy which the object belongs to
+* `context` - (Optional) The context which the object belongs to
```
annakhm, line 61: Lets specify here that if context is not provided, it will be derived from parent policy

website/docs/r/policy_security_policy_rule.html.markdown
```diff
@@ -0,0 +1,98 @@
...
+description: A resource to configure a Security Policy Rule.
+---
+
+# nsxt_policy_security_policy_rule
+
+This resource provides a method for the management of Security Policy Rule.
+
+Note: to avoid unexpected behavior, don't use this resource and resource `nsxt_policy_security_policy` to manage rules under a security policy at the same time. 
+Instead, please use this resource with resource `nsxt_policy_parent_security_policy` to manage a security policy and its rules separately, and use `nsxt_policy_security_policy` to manage a security policy and its rules in one single resource.
+Highly recommend to add `depends_on` to ensure its reference `nsxt_policy_parent_security_policy` could be consumed first.
```
annakhm, line 14: Can you please clarify why explicit dependency is needed?

website/docs/d/policy_security_policy_rule.html.markdown
```diff
@@ -0,0 +1,51 @@
+---
+subcategory: "Firewall"
```
annakhm, line 2: this document is not relevant anymore, since data source is removed?

website/docs/r/policy_security_policy_rule.html.markdown
GraysonWu: My bad. Before I misunderstood implicit dependency only works while creating resources not updating. Removed it.

nsxt/policy_common.go
GraysonWu: Done.

nsxt/resource_nsxt_policy_parent_security_policy_test.go
GraysonWu: Done.

nsxt/resource_nsxt_policy_parent_security_policy_test.go
GraysonWu: Done.

nsxt/resource_nsxt_policy_security_policy.go
GraysonWu: Good catch. Done.

website/docs/r/policy_security_policy_rule.html.markdown
GraysonWu: Done.

website/docs/r/policy_security_policy_rule.html.markdown
GraysonWu: Done.

website/docs/d/policy_security_policy_rule.html.markdown
GraysonWu: Done.

nsxt/resource_nsxt_policy_security_policy_no_rule.go
annakhm: I think when you fire the DELETE api call for the policy, all of its rules will be deleted - no need to delete them beforehand. What that your concern?

GraysonWu: I mean for example, there is a policy that has 200 rules and we want to delete 100 rules.

nsxt/resource_nsxt_policy_security_policy_rule.go
```diff
@@ -0,0 +1,246 @@
...
+	domain := getDomainFromResourcePath(policyPath)
+	policyID := getPolicyIDFromPath(policyPath)
+
+	if len(d.Get("context").([]interface{})) == 0 {
+		contexts := make([]interface{}, 1)
+		ctxMap := make(map[string]interface{})
+		ctxMap["project_id"] = projectID
+		contexts[0] = ctxMap
+		d.Set("context", contexts)
+	}
```
annakhm, line 46: I suggest a validation here that would fail if context for the rule is different than context for the parent policy?

nsxt/resource_nsxt_policy_security_policy_rule.go
```diff
@@ -0,0 +1,246 @@
...
+
+	// Initialize resource Id and verify this ID is not yet used
+	id, err := getOrGenerateID2(d, m, resourceNsxtPolicySecurityPolicyRuleExistsPartial(d.Get("policy_path").(string)))
```
annakhm, line 49: lets use `policyPath` here for readibility

GraysonWu: Done.

nsxt/resource_nsxt_policy_security_policy_rule.go
GraysonWu: Make sense. Done.

nsxt/policy_common.go
```diff
@@ -291,19 +298,29 @@ func getSecurityPolicyAndGatewayRulesSchema(scopeRequired bool, isIds bool, nsxI
...
-	return &schema.Schema{
-		Type:        schema.TypeList,
-		Description: "List of rules in the section",
-		Optional:    true,
-		MaxItems:    1000,
-		Elem: &schema.Resource{
-			Schema: ruleSchema,
-		},
+	if separated {
+		ruleSchema["policy_path"] = getPolicyPathSchema(true, false, "Security Policy path")
```
annakhm, line 302: The second parameter here denotes `forceNew`, which should be true - if policy path changes, rule should be recreated.

GraysonWu: Good catch. Done.

website/docs/r/policy_parent_security_policy.html.markdown
```diff
@@ -0,0 +1,160 @@
...
+  display_name       = "rule1"
+  description        = "Terraform provisioned Security Policy Rule"
+  policy_path        = nsxt_policy_parent_security_policy.policy1.path
+  sequence_number    = 1
+  destination_groups = [nsxt_policy_group.cats.path, nsxt_policy_group.dogs.path]
+  action             = "DROP"
+  services           = [nsxt_policy_service.icmp.path]
+  logged             = true
+
+  depends_on = [nsxt_policy_parent_security_policy.policy1]
```
annakhm, line 44: cosmetic - still some leftover `depends_on` in this doc

nsxt/resource_nsxt_policy_security_policy_rule.go
```diff
@@ -0,0 +1,252 @@
...
+	return client.Delete(domain, policyID, id)
+}
+
+func nsxtSecurityPolicyRuleImporter(d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
+	importID := d.Id()
+	rd, err := nsxtPolicyPathResourceImporterHelper(d, m)
+	if err != nil {
+		return rd, err
+	}
+	d.Set("policy_path", importID[:strings.Index(importID, "rule")-1])
```
annakhm, line 250: This requires validation, since we can't trust the user to always provide correct path. `rule` might not be present there at all, in which case provider will crash

website/docs/r/policy_parent_security_policy.html.markdown
GraysonWu: Done.

nsxt/resource_nsxt_policy_security_policy_rule.go
GraysonWu: Good one. Done.

annakhm (approved):
Thanks! LGTM

Nacymus:
Hello @GraysonWu ,

Why is  **nsxt_policy_security_policy_rule** still a beta source in the latest release ? I tried CRUD operations on it, it works wonders.

annakhm:
Hello @Nacymus, glad to hear it works for you! We also feel good about the quality of this resource, but haven't run the full set of tests on it yet. We expect to promote this from beta on next release.

PR #1063, opened 2023-12-12
Add support of overwriting role binding of local user

wsquan171:
This PR adds`overwrite_local_user` to role binding resource. When set to true, allows `Create` and `Delete` of role bindings of local users. On create, existing role bindings which comes with local user creation will be overwritten per terraform definition. On delete, the role binding will be reverted back to auditor only.

Also refactors `roles_for_path` as a set instead of list. This affects both principle identity and role binding resource, but no further doc change is needed.

Added build-in roles on NSX to doc for easier referenced. Linked NSX doc for actual permission matrix.

### DIFF
--- a/nsxt/resource_nsxt_policy_role_binding.go
+++ b/nsxt/resource_nsxt_policy_role_binding.go
@@ -6,6 +6,7 @@ package nsxt
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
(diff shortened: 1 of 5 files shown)


ksamoray:
Hi, I've tested the role part, it works :)
Can you also while you're at this, change `password_reset_required` to `Optional: true`?
[API](https://vdc-download.vmware.com/vmwb-repository/dcr-public/8e2a4b1a-659f-4be6-a4f4-7ca42b04cb84/9a2c1372-7174-40f0-89d2-27615bfd5d08/api_includes/types_NodeUserProperties.html) allows updating of this attribute but the resource defines it as computed only.

wsquan171:
> Hi, I've tested the role part, it works :) Can you also while you're at this, change `password_reset_required` to `Optional: true`? [API](https://vdc-download.vmware.com/vmwb-repository/dcr-public/8e2a4b1a-659f-4be6-a4f4-7ca42b04cb84/9a2c1372-7174-40f0-89d2-27615bfd5d08/api_includes/types_NodeUserProperties.html) allows updating of this attribute but the resource defines it as computed only.

`password_reset_required` is actually a read only attr, determined based on last password change and password policy. You can try PUT updating the field, it won't take any effect.

ksamoray:
Indeed it seems like my problem is different: API doc states that for `password_change_frequency`:
"This can be set to 0 to indicate no password change is required"
But instead it requires a password change for every login. I'll open a bug for this.
As for `password_reset_required` - doc says it's writable (and it didn't error when I set it with `curl`). However I don't see any impact, so let's leave it as is for now.

ksamoray (approved):

PR #1067, opened 2023-12-14
Revise transport_node and host_transport_node

ksamoray:
As nsxt_policy_transport_node covers host-node, the resource above is required to cover only edge-nodes.
Therefore the resource is renamed and non-edge attributes are removed.

Also remove redundant attributes from policy_host_transport_node resource.


annakhm:
Do we want to change `transport_zone_profile` to `transport_zone_profiles` as part of this PR as well, to avoid confusion on formatting this attribute (as a list rather than string)?

annakhm (approved):

martinrohrbach:
@ksamoray @annakhm 
We just started testing deployment of edges using the nsxt_transport_node resource and stumbled over the fact that the currently released provider doesn't allow for multiple "standard_host_switch" entries. This PR should fix this ... so ... would you by any chance know when the next release of the provider is planned? :)

(I thought it's easier to comment here rather than openening an issue, after all it's a beta resource as well, hope that is ok)

ksamoray:
> @ksamoray @annakhm We just started testing deployment of edges using the nsxt_transport_node resource and stumbled over the fact that the currently released provider doesn't allow for multiple "standard_host_switch" entries. This PR should fix this ... so ... would you by any chance know when the next release of the provider is planned? :)
> 
> (I thought it's easier to comment here rather than openening an issue, after all it's a beta resource as well, hope that is ok)

Hi @martinrohrbach, the release schedule is published [here](https://github.com/vmware/terraform-provider-nsxt/milestones):

PR #1068, opened 2023-12-15
[CI Fix] Specify display_name of LBAppProfile data source in test case

2ez4szliu:
In some test cases, we only specify the type for LbAppProfile data source when retrieving the default NSX LB app profile,
for example:
```
data "nsxt_policy_lb_app_profile" "default_http"{
  type = "HTTP"
}
```
this will fail if there are other test cases that run in parallel and create LB app profile of the same type.
We need to also specify display name for the default lb app profile when creating such data source
```
data "nsxt_policy_lb_app_profile" "default_http"{
  type = "HTTP"
  display_name = "default-http-lb-app-profile"
}
```


ksamoray (approved):
Adding some random element into the display name will reduce the risk of collision when there are some leftovers from tests which failed before.
But for now I think it's good as it is.

PR #1069, opened 2023-12-19
Resource to prepare for NSXT cluster upgrade

2ez4szliu:
Implement resources to support the upgrade for NSXT cluster
New resources and data source include:

1. nsxt_upgrade_prepare - upload precheck bundle and run precheck
2. nsxt_upgrade_precheck_acknowledge - acknowledge the warnings from precheck
3. data source nsxt_upgrade_prepare_status - Check if NSX Manager is ready for upgrade by checking if there are errors in precheck.

### DIFF
--- a/nsxt/provider.go
+++ b/nsxt/provider.go
@@ -318,6 +318,7 @@ func Provider() *schema.Provider {
 			"nsxt_host_upgrade_group":                        dataSourceNsxtHostUpgradeGroup(),
 			"nsxt_policy_gateway_interface_realization_info": dataSourceNsxtPolicyGatewayInterfaceRealizationInfo(),
 			"nsxt_upgrade_postcheck":                         dataSourceNsxtUpgradePostCheck(),
+			"nsxt_upgrade_prepare_ready":                     dataSourceNsxtUpgradePrepareReady(),
 		},
 
 		ResourcesMap: map[string]*schema.Resource{
@@ -473,6 +474,8 @@ func Provider() *schema.Provider {
 			"nsxt_policy_lb_udp_monitor_profile":           resourceNsxtPolicyLBUdpMonitorProfile(),
 			"nsxt_policy_tier0_gateway_gre_tunnel":         resourceNsxtPolicyTier0GatewayGRETunnel(),
 			"nsxt_upgrade_run":                             resourceNsxtUpgradeRun(),
+			"nsxt_upgrade_prepare":                         resourceNsxtUpgradePrepare(),
+			"nsxt_upgrade_precheck_acknowledge":            resourceNsxtUpgradePrecheckAcknowledge(),
 		},
 
 		ConfigureFunc: providerConfigure,
--- a/nsxt/resource_nsxt_upgrade_precheck_acknowledge.go
+++ b/nsxt/resource_nsxt_upgrade_precheck_acknowledge.go
@@ -1,0 +1,139 @@
+/* Copyright © 2023 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	nsxModel "github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/model"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/upgrade"
+)
+
+func resourceNsxtUpgradePrecheckAcknowledge() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNsxtUpgradePrecheckAcknowledgeCreate,
+		Read:   resourceNsxtUpgradePrecheckAcknowledgeRead,
+		Update: resourceNsxtUpgradePrecheckAcknowledgeUpdate,
+		Delete: resourceNsxtUpgradePrecheckAcknowledgeDelete,
+
+		Schema: map[string]*schema.Schema{
+			"precheck_ids": {
+				Type:        schema.TypeList,
+				Description: "IDs of precheck warnings that need to be acknowledged",
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				Required: true,
+			},
+			"precheck_warnings": {
+				Type:        schema.TypeList,
+				Description: "List of warnings from precheck",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:        schema.TypeString,
+							Description: "ID of the precheck warning",
+							Computed:    true,
+						},
+						"message": {
+							Type:        schema.TypeString,
+							Description: "Message of the precheck warning",
+							Computed:    true,
+						},
+						"is_acknowledged": {
+							Type:        schema.TypeBool,
+							Description: "Boolean value which identifies if the precheck warning has been acknowledged.",
+							Computed:    true,
+						},
+					},
+				},
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceNsxtUpgradePrecheckAcknowledgeCreate(d *schema.ResourceData, m interface{}) error {
+	id := d.Id()
+	if id == "" {
+		id = newUUID()
+	}
+	d.SetId(id)
+	precheckIDs := interface2StringList(d.Get("precheck_ids").([]interface{}))
+	err := acknowledgePrecheckWarnings(m, precheckIDs)
+	if err != nil {
+		return handleCreateError("NsxtPrecheckAcknowledge", id, err)
+	}
+	return resourceNsxtUpgradePrecheckAcknowledgeRead(d, m)
+}
+
+func acknowledgePrecheckWarnings(m interface{}, precheckIDs []string) error {
+	connector := getPolicyConnector(m)
+	client := upgrade.NewPreUpgradeChecksClient(connector)
+	for _, precheckID := range precheckIDs {
+		err := client.Acknowledge(precheckID)
+		if err != nil {
+			msg := fmt.Sprintf("Failed to acknowledge precheck warning with ID %s", precheckID)
+			return logAPIError(msg, err)
+		}
+	}
+	return nil
+}
+
+func resourceNsxtUpgradePrecheckAcknowledgeRead(d *schema.ResourceData, m interface{}) error {
+	id := d.Id()
+	typeParam := nsxModel.UpgradeCheckFailure_TYPE_WARNING
+	precheckWarnings, err := getPrecheckErrors(m, &typeParam)
+	if err != nil {
+		return handleReadError(d, "NsxtUpgradePrecheckAcknowledge", id, err)
+	}
+	err = setPrecheckWarningsInSchema(d, precheckWarnings)
+	if err != nil {
+		return handleReadError(d, "NsxtUpgradePrecheckAcknowledge", id, err)
+	}
+	return nil
+}
+
+func setPrecheckWarningsInSchema(d *schema.ResourceData, precheckWarnings []nsxModel.UpgradeCheckFailure) error {
+	var precheckWarningList []map[string]interface{}
+	for _, precheckWarning := range precheckWarnings {
+		elem := make(map[string]interface{})
+		elem["id"] = precheckWarning.Id
+		elem["message"] = precheckWarning.Message.Message
+		elem["is_acknowledged"] = precheckWarning.Acked
+		precheckWarningList = append(precheckWarningList, elem)
+	}
+	return d.Set("precheck_warnings", precheckWarningList)
+}
+
+func resourceNsxtUpgradePrecheckAcknowledgeUpdate(d *schema.ResourceData, m interface{}) error {
+	id := d.Id()
+	precheckIDs := interface2StringList(d.Get("precheck_ids").([]interface{}))
+	err := acknowledgePrecheckWarnings(m, precheckIDs)
+	if err != nil {
+		return handleUpdateError("NsxtPrecheckAcknowledge", id, err)
+	}
+	return resourceNsxtUpgradePrecheckAcknowledgeRead(d, m)
+}
+
+func resourceNsxtUpgradePrecheckAcknowledgeDelete(d *schema.ResourceData, m interface{}) error {
+	return nil
+}
+
+func getAcknowledgedPrecheckIDs(m interface{}) ([]string, error) {
+	var result []string
+	typeParam := nsxModel.UpgradeCheckFailure_TYPE_WARNING
+	precheckWarnings, err := getPrecheckErrors(m, &typeParam)
+	if err != nil {
+		return result, err
+	}
+	for _, warning := range precheckWarnings {
+		acked := *warning.Acked
+		if acked {
+			result = append(result, *warning.Id)
+		}
+	}
+	return result, nil
+}
--- a/nsxt/resource_nsxt_upgrade_prepare.go
+++ b/nsxt/resource_nsxt_upgrade_prepare.go
@@ -1,0 +1,493 @@
+/* Copyright © 2023 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"log"
+	"strings"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/vmware/vsphere-automation-sdk-go/lib/vapi/std/errors"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx"
+	nsxModel "github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/model"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/upgrade"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/upgrade/bundles"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/upgrade/eula"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/upgrade/pre_upgrade_checks"
+)
+
+var precheckComponentTypes = []string{"EDGE", "HOST", "MP"}
+
+const bundleUploadTimeout int = 3600
+const ucUpgradeTimeout int = 3600
+const precheckTimeout int = 3600
+
+func resourceNsxtUpgradePrepare() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNsxtUpgradePrepareCreate,
+		Read:   resourceNsxtUpgradePrepareRead,
+		Update: resourceNsxtUpgradePrepareUpdate,
+		Delete: resourceNsxtUpgradePrepareDelete,
+
+		Schema: map[string]*schema.Schema{
+			"version": {
+				Type:        schema.TypeString,
+				Description: "Target upgrade version for NSX, format is x.x.x..., should include at least 3 digits, example: 4.1.2",
+				Optional:    true,
+			},
+			"upgrade_bundle_url": {
+				Type:        schema.TypeString,
+				Description: "URL of the NSXT Upgrade bundle",
+				Required:    true,
+				ForceNew:    true,
+			},
+			"precheck_bundle_url": {
+				Type:        schema.TypeString,
+				Description: "URL of the NSXT Upgrade precheck bundle (Only applied to NSXT version >= 4.1.1)",
+				Optional:    true,
+			},
+			"accept_user_agreement": {
+				Type:        schema.TypeBool,
+				Description: "Whether to accept the user agreement",
+				Required:    true,
+			},
+			"bundle_upload_timeout": {
+				Type:        schema.TypeInt,
+				Description: "Timeout for uploading bundle in seconds",
+				Optional:    true,
+				Default:     bundleUploadTimeout,
+			},
+			"uc_upgrade_timeout": {
+				Type:        schema.TypeInt,
+				Description: "Timeout for upgrading upgrade coordinator in seconds",
+				Optional:    true,
+				Default:     ucUpgradeTimeout,
+			},
+			"precheck_timeout": {
+				Type:        schema.TypeInt,
+				Description: "Timeout for executing pre-upgrade checks in seconds",
+				Optional:    true,
+				Default:     precheckTimeout,
+			},
+			"failed_prechecks": {
+				Type:        schema.TypeList,
+				Description: "List of failed prechecks for the upgrade, including both warnings and errors",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"type": {
+							Type:        schema.TypeString,
+							Description: "Type of the precheck failure (warning or error)",
+							Computed:    true,
+						},
+						"id": {
+							Type:        schema.TypeString,
+							Description: "ID of failed precheck",
+							Computed:    true,
+						},
+						"message": {
+							Type:        schema.TypeString,
+							Description: "Message of the failed precheck",
+							Computed:    true,
+						},
+						"needs_ack": {
+							Type:        schema.TypeBool,
+							Description: "Boolean value which identifies if acknowledgement is required for the precheck",
+							Computed:    true,
+						},
+						"needs_resolve": {
+							Type:        schema.TypeBool,
+							Description: "Boolean value identifies if resolution is required for the precheck",
+							Computed:    true,
+						},
+						"acked": {
+							Type:        schema.TypeBool,
+							Description: "Boolean value which identifies if precheck has been acknowledged",
+							Computed:    true,
+						},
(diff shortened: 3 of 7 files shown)


nsxt/resource_nsxt_upgrade_precheck_acknowledge.go
```diff
@@ -0,0 +1,88 @@
...
+
+func resourceNsxtUpgradePrecheckAcknowledge() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNsxtUpgradePrecheckAcknowledgeCreate,
+		Read:   resourceNsxtUpgradePrecheckAcknowledgeRead,
+		Update: resourceNsxtUpgradePrecheckAcknowledgeUpdate,
+		Delete: resourceNsxtUpgradePrecheckAcknowledgeDelete,
+
+		Schema: map[string]*schema.Schema{
+			"precheck_id": {
```
annakhm, line 20: Can this be just a list of strings?

nsxt/resource_nsxt_upgrade_prepare.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+	err = setFailedPrechecksInSchema(d, precheckWarnings)
+	if err != nil {
+		return handleReadError(d, "NsxtUpgradePrepare", id, err)
+	}
+	return nil
+}
+
+func resourceNsxtUpgradePrepareUpdate(d *schema.ResourceData, m interface{}) error {
+	return nil
```
annakhm, line 195: Is this empty because the PR is work in progress? We need to implement update for repeated applies.

nsxt/resource_nsxt_upgrade_prepare.go
```diff
@@ -0,0 +1,227 @@
...
+	// 2. Accept user agreement
+	err = acceptUserAgreement(d, m)
+	if err != nil {
+		return handleCreateError("NsxtUpgradePrepare", id, err)
+	}
+	// 3. Execute pre-upgrade checks
+	err = executePreupgradeChecks(m)
+	if err != nil {
+		return handleCreateError("NsxtUpgradePrepare", id, err)
+	}
```
annakhm, line 169: I believe this code also needs to upgrade the upgrade coordinator

nsxt/resource_nsxt_upgrade_prepare.go
```diff
@@ -0,0 +1,227 @@
...
+	if err != nil {
+		return handleCreateError("NsxtUpgradePrepare", "", err)
+	}
+	return nil
+}
+
+func executePreupgradeChecks(m interface{}) error {
+	connector := getPolicyConnector(m)
+	client := nsx.NewUpgradeClient(connector)
+	return client.Executepreupgradechecks(nil, nil, nil, nil, nil, nil)
```
annakhm, line 122: I believe we should use a separate polling API to wait for the checks to complete

nsxt/resource_nsxt_upgrade_prepare.go
2ez4szliu: Hi Anna, yes you are right

nsxt/resource_nsxt_upgrade_prepare.go
2ez4szliu: I think In step 1 after we call API to upload the upgrade bundle, a API call to upgrade UC is automatically invoked.

GraysonWu:
I noticed that MP SDK is used here. Just curious if it still can work for the [latestVersion-1 ] -> [latestVersion] ?

2ez4szliu:
> I noticed that MP SDK is used here. Just curious if it still can work for the [latestVersion-1 ] -> [latestVersion] ? And is MP API being deprecated?
I think that at least for the APIs related cluster fabric and upgrade, they are not deprecated in MP and we still need to use them from nsxt-mp, it seems only the client function for accept user agreement are implemented in nsxt policy API, others are still in MP

nsxt/resource_nsxt_upgrade_prepare.go
```diff
@@ -0,0 +1,227 @@
...
+
+func waitForBundleUpload(m interface{}, bundleID string) error {
+	connector := getPolicyConnector(m)
+	client := bundles.NewUploadStatusClient(connector)
+	// c := m.(nsxtClients)
+	// Todo:
+	// 1. utilize Min/Max RetryInterval in provider config
+	interval := 20000
+	maxRetry := 50
+	for i := 0; i < 50; i++ {
```
annakhm, line 210: Lets consider using `stateConf.WaitForState()` tool from terraform sdk for consistency

nsxt/resource_nsxt_upgrade_precheck_acknowledge.go
2ez4szliu: Done

nsxt/resource_nsxt_upgrade_prepare.go
2ez4szliu: An update on this: yes we do need a separate polling API, otherwise the failed prechecks that we retrieved from FailureClient will only be the ones that are already executed.

nsxt/resource_nsxt_upgrade_precheck_acknowledge.go
```diff
@@ -0,0 +1,96 @@
...
+	precheckIDs := interface2StringList(d.Get("precheck_ids").([]interface{}))
+	err := acknowledgePrecheckWarnings(m, precheckIDs)
+	if err != nil {
+		return handleUpdateError("NsxtPrecheckAcknowledge", id, err)
+	}
+	return resourceNsxtUpgradePrecheckAcknowledgeRead(d, m)
+}
+
+func resourceNsxtUpgradePrecheckAcknowledgeDelete(d *schema.ResourceData, m interface{}) error {
+	return nil
```
annakhm, line 95: there is no mechanism to un-acknowledge a precheck? Not sure if it makes sense :)

nsxt/resource_nsxt_upgrade_prepare.go
```diff
@@ -0,0 +1,378 @@
...
+	if err != nil {
+		return handleCreateError("NsxtUpgradePrepare", id, err)
+	}
+	d.SetId(id)
+	return resourceNsxtUpgradePrepareRead(d, m)
+}
+
+func prepareForUpgrade(d *schema.ResourceData, m interface{}) error {
+	// 1. Upload upgrade bundle and wait for upload to complete
+	err := uploadPrecheckAndUpgradeBundle(d, m)
```
annakhm, line 89: I assumed we would skip this step if bundles are already uploaded (and haven't changed since), but I don't see this logic implemented. Is there a reason to upload again if bundle status is already ok?

nsxt/resource_nsxt_upgrade_prepare.go
```diff
@@ -0,0 +1,378 @@
...
+	err := client.Create()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func upgradeUc(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+	client := nsx.NewUpgradeClient(connector)
```
annakhm, line 220: Same comment as above - should we upgrade UC if this step was already executed succesfully?

nsxt/resource_nsxt_upgrade_prepare.go
```diff
@@ -0,0 +1,378 @@
...
+	client := upgrade.NewUcUpgradeStatusClient(connector)
+	pendingStates := []string{
+		nsxModel.UcUpgradeStatus_STATE_NOT_STARTED,
+		nsxModel.UcUpgradeStatus_STATE_IN_PROGRESS,
+	}
+	targetStates := []string{
+		nsxModel.UcUpgradeStatus_STATE_SUCCESS,
+		nsxModel.UcUpgradeStatus_STATE_FAILED,
+	}
+	timeout := 1200
```
annakhm, line 317: Should we make the timeout configurable on the resource?

nsxt/resource_nsxt_upgrade_prepare.go
```diff
@@ -0,0 +1,378 @@
...
+		return nil, err
+	}
+	return resultList.Results, nil
+}
+
+func setFailedPrechecksInSchema(d *schema.ResourceData, precheckWarnings []nsxModel.UpgradeCheckFailure) error {
+	var failedPrechecksList []map[string]interface{}
+	for _, warning := range precheckWarnings {
+		if *warning.NeedsAck {
+			id := *warning.Id
```
annakhm, line 258: I think we should list all warnings, including those that don't require ack. Maybe we can list them in separate arrays for clarity.

nsxt/resource_nsxt_upgrade_prepare.go
```diff
@@ -0,0 +1,378 @@
...
+			failureIDs = append(failureIDs, failureID)
+		}
+		return fmt.Errorf("Encounter errors in precheck %s, please address these error on NSX UI", failureIDs)
+	}
+	return nil
+}
+
+func resourceNsxtUpgradePrepareRead(d *schema.ResourceData, m interface{}) error {
+	id := d.Id()
+	precheckWarnings, err := getPrecheckErrors(m, nsxModel.UpgradeCheckFailure_TYPE_WARNING)
```
annakhm, line 126: Are we presenting errors as well somewhere?

annakhm: Also, I wonder if acknowledged prechecks would disappear from the list once acknowledged?

nsxt/resource_nsxt_upgrade_prepare.go
```diff
@@ -0,0 +1,378 @@
...
+}
+
+func uploadPrecheckAndUpgradeBundle(d *schema.ResourceData, m interface{}) error {
+	upgradeBundleURL := d.Get("upgrade_bundle_url").(string)
+	precheckBundleURL := d.Get("precheck_bundle_url").(string)
+	upgradeBundleType := nsxModel.UpgradeBundleFetchRequest_BUNDLE_TYPE_UPGRADE
+	precheckBundleType := nsxModel.UpgradeBundleFetchRequest_BUNDLE_TYPE_PRE_UPGRADE
+	version := d.Get("version").(string)
+	c := m.(nsxtClients)
+	userName := c.NsxtClientConfig.UserName
```
annakhm, line 157: other types of auth are not supported?
We'll need to clarify the docs about limitations of upgrade config.

2ez4szliu: I'm afraid that's the case, will add this to the doc.
Since nsxt-mp/v0.4.0, password, username, version and bundle type are added to the bundleFetchRequest schema, before v0.4.0 only URL is required. (I think nsxt-mp/v0.4.0 corresponds to nsx version 4.1.1 or higher because that's when NSX starts to support uploading pub type bundle)

nsxt/resource_nsxt_upgrade_precheck_acknowledge.go
2ez4szliu: No there is no un-acknowledge operation for precheck warnings(from either API or UI)

nsxt/resource_nsxt_upgrade_prepare.go
2ez4szliu: If there are errors then the creation of the resource will fail, and id of these errors will be displayed in error message

nsxt/resource_nsxt_upgrade_prepare.go
2ez4szliu: Added a step to check upgrade summary, it will compare the existing bundle file name to the bundle file name from url, if they are equal then skip the upload process.

nsxt/resource_nsxt_upgrade_prepare.go
2ez4szliu: Done

nsxt/resource_nsxt_upgrade_prepare.go
2ez4szliu: done

nsxt/resource_nsxt_upgrade_prepare.go
2ez4szliu: Done. now the timeout for upgrading UC, execute precheck and upload bundle configurable to user

nsxt/resource_nsxt_upgrade_precheck_acknowledge.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return handleReadError(d, "NsxtUpgradePrecheckAcknowledge", id, err)
+	}
+	return nil
+}
+
+func setAcknowledgedPrecheckIDsInSchema(d *schema.ResourceData, precheckWarnings []nsxModel.UpgradeCheckFailure) error {
+	var precheckWarningIDs []string
+	for _, precheckWarning := range precheckWarnings {
+		if !(*precheckWarning.NeedsAck) {
```
annakhm, line 76: Does `NeedsAck` reflect ack state? I understood that this flag indicates whether ack is needed to proceed, not whether it has been acknowledged by the ack API. If my understanding is correct, this would prevent the check from being assigned to state and will cause non-empty state.

2ez4szliu: I am still thinking about what would be the final state of this resource, I think we should have a computed attribute for this resource to show users all prechecks that were acknowledged, or maybe just all precheck failures like the upgrade_prepare resource?

nsxt/resource_nsxt_upgrade_prepare.go
```diff
@@ -0,0 +1,451 @@
...
+				Type:        schema.TypeString,
+				Description: "URL of the NSXT Upgrade precheck bundle (Only applied to NSXT version >= 4.1.1)",
+				Optional:    true,
+			},
+			"accept_user_agreement": {
+				Type:        schema.TypeBool,
+				Description: "Whether to accept the user agreement",
+				Required:    true,
+			},
+			"bundle_upload_timeout": {
```
annakhm, line 52: I would suggest to define timeouts under a sub-clause

nsxt/resource_nsxt_upgrade_prepare.go
```diff
@@ -0,0 +1,451 @@
...
+	} else {
+		// Check if same precheck bundle has already been uploaded, if so then skip uploading.
+		// example bundle version from NSX API: 5.0.0.0.0.23236379
+		// example precheck bundle name: VMware-NSX-upgrade-bundle-5.0.0.0.0.23236379-pre-check.pub
+		if summary.PreUpgradeBundleVersion != nil {
+			existingBundleVersion := *summary.PreUpgradeBundleVersion
+			arr := strings.Split(url, "/")
+			bundleName := arr[len(arr)-1]
+			bundleVersionArr := strings.Split(bundleName, "-")
+			if len(bundleVersionArr) != 7 {
```
annakhm, line 234: I'm not sure we should rely on the name from url, since it could be anything..

2ez4szliu: Unfortunately, it seems the only way to get uploaded bundle information is through the upgrade/summary API, which will return the name of uploaded mub bundle and the version of uploaded pub bundle... If user decides to rename the bundle to a random format I'm afraid we cannot check if the same bundle has been uploaded

nsxt/resource_nsxt_upgrade_prepare.go
2ez4szliu: Resolve this conversation as we are displaying both error and warnings in final state of the resource

nsxt/resource_nsxt_upgrade_prepare.go
2ez4szliu: Added `version` back to the schema of the resource, it will be used to determine if precheck bundle needs re-upload. For upgrade bundle, we use d.HasChange(bundleURL) to decide if reload is needed.

nsxt/provider.go
```diff
@@ -317,6 +317,7 @@ func Provider() *schema.Provider {
 			"nsxt_edge_upgrade_group":                        dataSourceNsxtEdgeUpgradeGroup(),
 			"nsxt_host_upgrade_group":                        dataSourceNsxtHostUpgradeGroup(),
 			"nsxt_policy_gateway_interface_realization_info": dataSourceNsxtPolicyGatewayInterfaceRealizationInfo(),
+			"nsxt_upgrade_prepare_ready":                     dataSourceNsxtUpgradePrepareReady(),
```
annakhm, line 320: Just a naming comment - `prepare_ready` sounds non-optimal to me, maybe `prepare_done`? Other suggestions? @GraysonWu @salv-orlando

website/docs/d/upgrade_prepare_ready.html.markdown
```diff
@@ -0,0 +1,25 @@
...
+
+This data source provides information about the upgrade readiness of NSX Manager.
+The resource creation will succeed only if there are no errors in precheck and
+all warnings have been acknowledged.
+
+## Example Usage
+
+```hcl
+data "nsxt_upgrade_prepare_ready" "test" {
+  upgrade_prepare_id = nsxt_upgrade_prepare.test.id
```
annakhm, line 18: I think we should explain that this data source might depend on  `nsxt_upgrade_precheck_acknowledge` if those exist. I would add this to the example.

nsxt/resource_nsxt_upgrade_prepare.go
```diff
@@ -0,0 +1,490 @@
...
+	if err != nil {
+		return err
+	}
+	timeout := d.Get("uc_upgrade_timeout").(int)
+	return waitForUcUpgrade(m, timeout)
+}
+
+func executePreupgradeChecks(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+	client := nsx.NewUpgradeClient(connector)
```
annakhm, line 335: Lets add some logging here, and also in `precheckNeeded` routine so that we know why precheck was triggered or not triggered.

nsxt/provider.go
2ez4szliu: prepare_complete? I'm ok with both

nsxt/resource_nsxt_upgrade_prepare.go
2ez4szliu: done

website/docs/d/upgrade_prepare_ready.html.markdown
2ez4szliu: done

website/docs/r/upgrade_prepare.html.markdown
```diff
@@ -0,0 +1,58 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `upgrade_bundle_url` - (Required) The url to download the Manager Upgrade bundle.
+* `version` - (Optional) Target upgrade version for NSX, format is x.x.x..., should include at least 3 digits, example: 4.1.2
+* `precheck_bundle_url` - (Optional) The url to download the Precheck bundle. This field is required and only applied to NSX version >= 4.1.1
```
salv-orlando, line 39: nit: This field is required if NSX >= 4.1.1

The current statement creates a bit of confusion on whether the parameter is required or optional

website/docs/r/upgrade_prepare.html.markdown
```diff
@@ -0,0 +1,58 @@
...
+* `accept_user_agreement` - (Required) User agreement must be accepted before upgrade.
```
salv-orlando, line 40: I think it's too trivial to add that it must be set to True otherwise the upgrade won't proceed?

website/docs/r/upgrade_precheck_acknowledge.html.markdown
```diff
@@ -0,0 +1,33 @@
...
+resource "nsxt_upgrade_precheck_acknowledge" "test" {
+  precheck_ids = ["backupOperationCheck", "pUBCheck"]
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `precheck_ids` - (Required) List of ids of failed prechecks user want to acknowledge.
```
salv-orlando, line 25: nit: either "users want" or "user wants"

website/docs/d/upgrade_prepare_ready.html.markdown
```diff
@@ -0,0 +1,28 @@
...
+subcategory: "Beta"
+layout: "nsxt"
+page_title: "NSXT: upgrade_prepare_ready"
+description: A data source to check if NSX Manager is ready for upgrade.
+---
+
+# nsxt_upgrade_prepare_ready
+
+This data source provides information about the upgrade readiness of NSX Manager.
+The resource creation will succeed only if there are no errors in precheck and
```
salv-orlando, line 11: Question: is it correct to talk about "resource creation" for a datasource?

website/docs/d/upgrade_prepare_ready.html.markdown
```diff
@@ -0,0 +1,28 @@
...
+all warnings have been acknowledged.
+If there are nsxt_upgrade_precheck_acknowledge resource, then nsxt_upgrade_prepare_ready resource
```
salv-orlando, line 13: nit: if there is any `nsxt_upgrade_precheck_acknowledge` resource, ....

website/docs/r/upgrade_prepare.html.markdown
2ez4szliu: done

website/docs/r/upgrade_prepare.html.markdown
2ez4szliu: I have added to doc that this field must be set to true otherwise upgrade will not proceed

nsxt/provider.go
annakhm: Maybe just `nsxt_upgrade_ready`?

nsxt/resource_nsxt_upgrade_precheck_acknowledge.go
```diff
@@ -0,0 +1,141 @@
...
+		return handleReadError(d, "NsxtUpgradePrecheckAcknowledge", id, err)
+	}
+	return nil
+}
+
+func setPrecheckWarningsInSchema(d *schema.ResourceData, precheckWarnings []nsxModel.UpgradeCheckFailure) error {
+	var precheckWarningList []map[string]interface{}
+	for _, precheckWarning := range precheckWarnings {
+		id := *precheckWarning.Id
+		message := *(precheckWarning.Message.Message)
```
annakhm, line 102: Can we assign the values directly, as we usually do?
```
elem["message"] = precheckWarning.Message.Message
```

This way terraform `Set` would check for nil pointer when dereferencing the value

nsxt/resource_nsxt_upgrade_prepare.go
```diff
@@ -0,0 +1,502 @@
...
+	connector := getPolicyConnector(m)
+	summaryClient := upgrade.NewSummaryClient(connector)
+	summary, err := summaryClient.Get()
+	if err != nil {
+		return false, err
+	}
+	var bundleURL string
+	var ucUpgraded bool
+	var upgradeStatus string
+	if summary.UpgradeBundleFileName != nil {
```
annakhm, line 156: If `UpgradeBundleFileName` is nil, seems like we need to skip precheck as well?

2ez4szliu: Adjusted the if condition to `if summary.UpgradeBundleFileName == nil || len(*summary.UpgradeBundleFileName) == 0`
same for the other 2 conditions

nsxt/resource_nsxt_upgrade_precheck_acknowledge.go
2ez4szliu: done

nsxt/resource_nsxt_upgrade_precheck_acknowledge.go
```diff
@@ -0,0 +1,138 @@
...
+	return resourceNsxtUpgradePrecheckAcknowledgeRead(d, m)
+}
+
+func acknowledgePrecheckWarnings(m interface{}, precheckIDs []string) error {
+	connector := getPolicyConnector(m)
+	client := upgrade.NewPreUpgradeChecksClient(connector)
+	for _, precheckID := range precheckIDs {
+		err := client.Acknowledge(precheckID)
+		if err != nil {
+			return fmt.Errorf("Failed to acknowledge precheck warning with ID %s: %s", precheckID, err)
```
annakhm, line 78: If formatted error is returned here, `handleCreateError` will fail to parse it as NSX error, and the user will see something like ` Failed to acknowledge precheck warning with ID backupOperationCheck: com.vmware.vapi.std.errors.invalid_request`
I would suggest to either return `err` here, or use `logApiError` to convey NSX message to the user

2ez4szliu: Done

annakhm:
When dependency on `nsxt_upgrade_precheck_acknowledge` resource is missing in `nsxt_upgrade_prepare_ready` data source, we get into never-ending cycle of prechecks and acks that never gets resolved:
1. `prepare` resource initiates fresh precheck cycle that would require re-acknowledging
2.  `ready` data source gets evaluated right away, and acks are missing
3. `ack` resource does its job, but too late
For now lets document this, but in a follow up patch I suggest to solve this with amending `prepare` resource to only re-run prechecks if needed, which is one of the following cases:
1. system is ready for its first run of prechecks on this upgrade bundle (i.e, UC is upgraded etc, but we never run prechecks before)
2. prechecks were executed before and resulted in error or warning that requires ack

we need to be careful with this change since there might be scenario when prechecks were run before, but user decided to change upgrade bundle.

annakhm (approved):

PR #1071, opened 2023-12-20
Node user tests should run for v4.1 and above

ksamoray:
As the API used isn't implemented before this version.


annakhm (approved):

wsquan171:
Thanks Kobi. PI and role bindings should be available on 3.2 though. I'm not sure if the `roles_for_path` is supported though, as api doc shows conflicting results on 3.2.1. Pretty sure both should be OK from 4.0 at least.

wsquan171 (approved):

ksamoray:
Thanks @wsquan171

PR #1074, opened 2024-01-03
BGP config: support local_as_num and tags for VRFs.

salv-orlando:
These two attributes can also be specified for BGP configuration on VRFs. The terraform provider currently only sets them for BGP config for Tier-0 routers.


annakhm:
Looks like documentation also needs to be updated (it states ` BGP AS number in ASPLAIN/ASDOT Format. This setting is not applicable to VRF-Lite Gateway, and is required otherwise.`)

salv-orlando:
@annakhm your note about documentation also point to the fact that for a complete fix we also need to update the Tier0 Gateway resource as it allows for directly setting bgpconfigs

annakhm (approved):

PR #1075, opened 2024-01-03
Add pagination to certificate data source

### LINKED ISSUE #849: data resource "nsxt_certificate" does not support pagination
### Describe the bug

We've been using the data resource "nsxt_certificate" for years to import certificates using the display_name in order to use them for virtual servers.

However, we just recently updated NSX-T from 3.1.3.7 to 3.2.1.2 and suddenly we cannot import certain certificates by name anymore although we can confirm that they are present in NSX-T.

Digging into this, it seems to be a pagination issue. I can see that the API to retrieve certificates provides about 50 or so at a time and we have > 200 certificates installed. I am not sure if the number of returned certificated per page has changed between versions or we were lucky in the past that all relevant certificates were on the first page, but it was working before and isn't working anymore after the upgrade. 

The relevant code also lists a "todo" to add pagination so I think the culprit is quite clear:
https://github.com/vmware/terraform-provider-nsxt/blob/2842c1441d130b52edf56351955f377b803d0896/nsxt/data_source_nsxt_certificate.go#L66

(We've implemented a workaround using the "external" provider and a python script that's simply reading all certificates WITH pagination so we can still reference the IDs from within terraform, but still I would consider this a bug.)

### Reproduction steps

1. Have a large number of certificates installed in NSX-T
2. Check for a certificate that appears on 2nd+ page of the /api/v1/trust-management/certificates API call
3. Import said certificate usind the display_name and "nsxt_certificate" data resource
4. Terraform reports that the certificate cannot be found

### Expected behavior

Certificate is found, i.e. pagination is honoured.

### Additional context

_No response_

wsquan171:
This PR adds pagination support to certificate data source when getting a cert by name.

Fixes #849 

Testing done:
Verified importing cert on NSX-T by name with overriding `info.LocalVarOptionals["pageSize"] = int64(20)`
while total cert count is 100. 

Confirmed with debug log with proper pagination, for both found, not found and multiple found casaes.

### DIFF
--- a/nsxt/data_source_nsxt_certificate.go
+++ b/nsxt/data_source_nsxt_certificate.go
@@ -5,7 +5,6 @@ package nsxt
 
 import (
 	"fmt"
-
 	"net/http"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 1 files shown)


annakhm:
Hi Shawn, I suggest to use `handlePagination` routine that is used by other MP data sources. It has printouts and additional safeguards.

wsquan171:
> Hi Shawn, I suggest to use `handlePagination` routine that is used by other MP data sources. It has printouts and additional safeguards.

Thanks. Fixed

annakhm (approved):

PR #1076, opened 2024-01-04
Add upgradeGroup data source

GraysonWu:


annakhm:
Do we know when the upgrade groups are created in NSX config? If they are not yet ready by the time data sources are evaluated, we need a special handing (at least not to fail config)

GraysonWu:
> Do we know when the upgrade groups are created in NSX config? If they are not yet ready by the time data sources are evaluated, we need a special handing (at least not to fail config)

Thanks. Good one. I didn't find any places mentioning this. I have added a polling func to avoid this issue.

annakhm:
> > Do we know when the upgrade groups are created in NSX config? If they are not yet ready by the time data sources are evaluated, we need a special handing (at least not to fail config)
> 
> Thanks. Good one. I didn't find any places mentioning this. I have added a polling func to avoid this issue.

If the groups are not yet populated by NSX at the time config is evaluated, perhaps we better fail fast rather than wait (and also recommend making groups dependant on `prepare` resource, if that resources triggers group creation. 
if we have established that `prepare` resource triggers groups creation, we may also consider making this dependency explicit by adding a fake attribute `prepare_upgrade_id` or something similar.

GraysonWu:
> > > Do we know when the upgrade groups are created in NSX config? If they are not yet ready by the time data sources are evaluated, we need a special handing (at least not to fail config)
> > 
> > 
> > Thanks. Good one. I didn't find any places mentioning this. I have added a polling func to avoid this issue.
> 
> If the groups are not yet populated by NSX at the time config is evaluated, perhaps we better fail fast rather than wait (and also recommend making groups dependant on `prepare` resource, if that resources triggers group creation. if we have established that `prepare` resource triggers groups creation, we may also consider making this dependency explicit by adding a fake attribute `prepare_upgrade_id` or something similar.

As I tested, upgrade groups will be created after a few seconds after the upgrade bundle upload starts.
Added the attribute for the sake of dependency and change the timeout to 1 min.

nsxt/data_source_nsxt_edge_upgrade_group.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	d.SetId(*obj.Id)
+	d.Set("display_name", obj.DisplayName)
+	d.Set("description", obj.Description)
+
+	return nil
+}
+
+func waitUpgradeGroupCreate(client upgrade.UpgradeUnitGroupsClient, groupType string) error {
```
annakhm, line 115: Can we use `waitForState` for consistency here?

GraysonWu: Done.

nsxt/data_source_nsxt_edge_upgrade_group.go
```diff
@@ -0,0 +1,144 @@
...
+
+func dataSourceNsxtEdgeUpgradeGroupRead(d *schema.ResourceData, m interface{}) error {
+	return upgradeGroupRead(d, m, edgeUpgradeGroup)
+}
+
+func upgradeGroupRead(d *schema.ResourceData, m interface{}, groupType string) error {
+	connector := getPolicyConnector(m)
+	client := upgrade.NewUpgradeUnitGroupsClient(connector)
+
+	err := waitUpgradeGroupCreate(client, groupType)
```
annakhm, line 50: This would be an extra roundtrip for each group.. I don't think it would be very noticeable, but I'm also not sure this is needed one we have the `upgrade_prepare_id` being forced on the user. The prepare resource waits for quite a few async processes, and I would expect the groups to be created way before it finishes - do you think further wait would be required here?
I would suggest to remove this wait for now, and add it only if we experience the actual need for it. What do you think?

GraysonWu: Make sense. And based on my test those groups will be created in the early stage of uploading the upgrade bundle. We could add it back later if necessary.

annakhm (approved):

PR #1078, opened 2024-01-08
Fix connection initialization flow for joined node

annakhm:
When a connector is initialized for the sake of joining the manager cluster, the endpoint is different from main provider endpoint, thus global connection init flows such as NSX Version retrieval and licenses should not be triggered by such connection.

### DIFF
--- a/nsxt/resource_nsxt_manager_cluster.go
+++ b/nsxt/resource_nsxt_manager_cluster.go
@@ -168,7 +168,7 @@ func waitForNodeStatus(d *schema.ResourceData, m interface{}, nodes []NsxCluster
 		log.Printf("[DEBUG]: API probing for NSX is disabled")
 		return nil
 	}
-	connector := getPolicyConnectorForInit(m, false)
+	connector := getStandalonePolicyConnector(m, false)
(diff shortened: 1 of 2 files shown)


nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -168,7 +168,7 @@ func waitForNodeStatus(d *schema.ResourceData, m interface{}, nodes []NsxCluster
 		log.Printf("[DEBUG]: API probing for NSX is disabled")
 		return nil
 	}
-	connector := getPolicyConnectorForInit(m, false)
+	connector := getStandalonePolicyConnector(m, false)
```
ksamoray, line 171: I still see the same failure with the PR applied.
I've tried to change this line to 
```
```
Which worked - I think that as with `getStandalonePolicyConnector` we skip license assignment, this is where we want to assign the license and therefore use the regular `getPolicyConnector`

annakhm: This connection just waits for node status, it should not require license, but we don't want it to initialize the version for the main NSX connection and thus skip license assignment for it..

ksamoray (approved):
LGTM!

PR #1080, opened 2024-01-09
Add VMC Oauth app support

### LINKED ISSUE #944: Oauth App Support for VMConAWS authentication

NSXT Terraform already supports using the CSP refresh token to obtain an access token from CSP, which is then used for authenticating to VMC NSX. Similarly, OAuth2 is another method of obtaining an access token.

https://blogs.vmware.com/cloud/2023/03/28/announcing-oauth-app-support-for-vmware-cloud-automation-tools/

This issue is to add this additional method

Implementation of OAuth App Support for the NSX-T Terraform Provider for VMConAWS environmnents:
https://blogs.vmware.com/cloud/2023/03/28/announcing-oauth-app-support-for-vmware-cloud-automation-tools/

_No response_

### Additional context

_No response_

wsquan171:
This change adds support of NSX token exchange on VMC using Oauth app, which is a newer method compared to using refresh token.

Implementation mainly referenced [python client for vmc on aws](https://github.com/vmware/python-client-for-vmware-cloud-on-aws/blob/development/pyvmc_fxns.py#L166) and terraform vmc provider.

Testing done:
With Oauth app created in server-to-server mode, assigned with all roles under "VMware Cloud on AWS" category, provider configured as:

```
provider "nsxt" {
  host                 = "<nsx-private-ip>.rp.vmwarevmc.com/vmc/reverse-proxy/api/orgs/<org-id>/sddcs/<ssdc-id>/sks-nsxt-manager"
  allow_unverified_ssl = true
  max_retries          = 2
 
  enforcement_point    = "vmc-enforcementpoint"
  vmc_client_id        = "<oauth-app-id>"
  vmc_client_secret    = "<oauth-app-secret>"
}
``` 

Able to create and delete resources like:
```
resource "nsxt_policy_gateway_qos_profile" "test" {
  display_name        = "test"
  description         = "Terraform provisioned profile"
  burst_size          = 10
  committed_bandwidth = 20
}
```

For regression, verified using access token to create and delete the same resource. The provider is configured as follows:
```
provider "nsxt" {
  host                 = "<nsx-private-ip>.rp.vmwarevmc.com/vmc/reverse-proxy/api/orgs/<org-id>/sddcs/<ssdc-id>/sks-nsxt-manager"
  allow_unverified_ssl = true
  max_retries          = 2
 
  enforcement_point    = "vmc-enforcementpoint"
  vmc_token            = "<access-token>"
}
```

e2e test covers non-vmc cases.

Closes #944

### DIFF
--- a/nsxt/provider.go
+++ b/nsxt/provider.go
@@ -158,13 +158,30 @@ func Provider() *schema.Provider {
 				Type:        schema.TypeString,
 				Optional:    true,
 				Description: "URL for VMC authorization service (CSP)",
-				DefaultFunc: schema.EnvDefaultFunc("NSXT_VMC_AUTH_HOST", "console.cloud.vmware.com/csp/gateway/am/api/auth/api-tokens/authorize"),
+				DefaultFunc: schema.EnvDefaultFunc("NSXT_VMC_AUTH_HOST", nil),
 			},
 			"vmc_token": {
-				Type:        schema.TypeString,
-				Optional:    true,
-				Description: "Long-living API token for VMC authorization",
-				DefaultFunc: schema.EnvDefaultFunc("NSXT_VMC_TOKEN", nil),
+				Type:          schema.TypeString,
+				Optional:      true,
+				Description:   "Long-living API token for VMC authorization",
+				DefaultFunc:   schema.EnvDefaultFunc("NSXT_VMC_TOKEN", nil),
+				ConflictsWith: []string{"vmc_client_id", "vmc_client_secret"},
+			},
+			"vmc_client_id": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Description:   "ID of OAuth App associated with the VMC organization",
+				DefaultFunc:   schema.EnvDefaultFunc("NSXT_VMC_CLIENT_ID", nil),
+				ConflictsWith: []string{"vmc_token"},
+				RequiredWith:  []string{"vmc_client_secret"},
+			},
+			"vmc_client_secret": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Description:   "Secret of OAuth App associated with the VMC organization",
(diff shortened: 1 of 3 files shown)


nsxt/provider.go
```diff
@@ -158,13 +158,30 @@ func Provider() *schema.Provider {
 				Type:        schema.TypeString,
 				Optional:    true,
 				Description: "URL for VMC authorization service (CSP)",
-				DefaultFunc: schema.EnvDefaultFunc("NSXT_VMC_AUTH_HOST", "console.cloud.vmware.com/csp/gateway/am/api/auth/api-tokens/authorize"),
```
annakhm, line 161: I'm afraid we're not allowed to change defaults in order not to break backward compatibility. Is this change required?

wsquan171: Technically the backward compatibility is not broken in that the same config (leaving out auth host in both config and env) will still work. The same default value is set in another place.

The reason I need to change this, is because oauth and access token need to use different CSP API URL to exchange for api token. The default value now needs to be adjusted based on auth method used (access token or oauth app). I checked that currently from schema this field is consumed in `getAPIToken`. I think it should be safe.

annakhm (approved):

website/docs/index.html.markdown
```diff
@@ -221,7 +221,15 @@ The following arguments are used to configure the VMware NSX-T Provider:
   partially successful realization as valid state and not fail apply.
 * `vmc_token` - (Optional) Long-lived API token for authenticating with VMware
   Cloud Services APIs. This token will be used to short-lived token that is
-  needed to communicate with NSX Manager in VMC environment.
+  needed to communicate with NSX Manager in VMC environment. In conflict with
```
annakhm, line 224: Just cosmetic comment - I'm not sure `in conflict` is a term the user will always understand. Perhaps `can not be specified together with XXX`?

PR #1081, opened 2024-01-12
Clarify and enhance docs for IDS Profile

annakhm:
Different representation of values in the UI causes confusion. This PR adds examples to the docs and adds a note about potential confusion.


wsquan171 (approved):

ksamoray:
Change LGTM, but BTW where would a user get the values for attack types, targets etc?
Are those documented anywhere?

annakhm:
> Change LGTM, but BTW where would a user get the values for attack types, targets etc? Are those documented anywhere?

As I understand, NSX teams are working on giving access to these lists of values, and also possibly reverting the UI to show the actual value rather than the prettified one.

PR #1084, opened 2024-01-18
Add discover node data source

GraysonWu:
1. Add Discover Node data source.

2. Remove node_deployment_info from Host Transport Node. Considering we only support Host Transport Node create from vCenter server managing the ESXi type HostNode and in this case, discovered_node_id is required, so no need to provide node_deployment_info.
Additionally, this could fix the issue that using node_deployment_info with VDS type host switch, which is the only host switch type supported, will receive error from NSX:
```
Failed to create HostTransportNode 9756979d-0165-4cc6-93dc-0c30bb6da065:
VDS Configuration is specified for host 9756979d-0165-4cc6-93dc-0c30bb6da065
and its not managed by a vCenter. Please correct TransportNode configuration
or connect the host to a VCenter. (code 9549)
```

### DIFF
--- a/nsxt/data_source_discover_node.go
+++ b/nsxt/data_source_discover_node.go
@@ -1,0 +1,70 @@
+/* Copyright © 2024 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/fabric"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/model"
+)
+
+func dataSourceNsxtDiscoverNode() *schema.Resource {
+	return &schema.Resource{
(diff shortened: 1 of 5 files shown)


nsxt/data_source_discover_node.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+		if err != nil {
+			return fmt.Errorf("Error while reading Discover Node %s: %v", objID, err)
+		}
+		obj = objGet
+	} else if ipAddress == "" {
+		return fmt.Errorf("Error obtaining Discover Node external ID or IP address during read")
+	} else {
+		// Get by IP address
+		objList, err := discoverNodeClient.List(nil, nil, nil, nil, nil, nil, &ipAddress, nil, nil, nil, nil, nil, nil, nil)
```
annakhm, line 57: I feel pagination might be required here for big deployments?

GraysonWu: Maybe I misunderstood something. Can multiple Nodes share the same IP address? I expect only one or none(NotFound) Node will be returned.

annakhm: My bad, I didn't realize the List has IP filter already in the call

annakhm (approved):

PR #1085, opened 2024-01-20
Add Upgrade Run Resource

GraysonWu:
Rebased on PR #1076

### DIFF
--- a/nsxt/data_source_nsxt_upgrade_postcheck.go
+++ b/nsxt/data_source_nsxt_upgrade_postcheck.go
@@ -1,0 +1,159 @@
+/* Copyright © 2024 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/model"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/upgrade/upgrade_unit_groups"
+)
+
+var (
+	// Default waiting setup in seconds
+	defaultUpgradePostcheckInterval = 30
+	defaultUpgradePostcheckTimeout  = 1800
+	defaultUpgradePostcheckDelay    = 30
+)
+
+func dataSourceNsxtUpgradePostCheck() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNsxtUpgradePostCheckRead,
+
+		Schema: map[string]*schema.Schema{
+			"upgrade_run_id": {
+				Type:        schema.TypeString,
+				Description: "ID of corresponding nsxt_upgrade_run resource",
+				Required:    true,
+			},
+			"type": {
+				Type:         schema.TypeString,
+				Description:  "Component Type",
+				Required:     true,
+				ValidateFunc: validation.StringInSlice([]string{edgeUpgradeGroup, hostUpgradeGroup}, false),
+			},
+			"timeout": {
+				Type:         schema.TypeInt,
+				Description:  "Post-check status checks timeout in seconds",
+				Optional:     true,
+				Default:      defaultUpgradePostcheckTimeout,
+				ValidateFunc: validation.IntAtLeast(1),
+			},
+			"interval": {
+				Type:         schema.TypeInt,
+				Description:  "Interval to check Post-check status in seconds",
+				Optional:     true,
+				Default:      defaultUpgradePostcheckInterval,
+				ValidateFunc: validation.IntAtLeast(1),
+			},
+			"delay": {
+				Type:         schema.TypeInt,
+				Description:  "Initial delay to start Post-check status checks in seconds",
+				Optional:     true,
+				Default:      defaultUpgradePostcheckDelay,
+				ValidateFunc: validation.IntAtLeast(0),
+			},
+			"failed_group": {
+				Type:        schema.TypeList,
+				Description: "Upgrade unit groups that failed in upgrade post-checks",
+				Computed:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:        schema.TypeString,
+							Description: "Unique identifier of upgrade unit group",
+							Required:    true,
+						},
+						"status": {
+							Type:        schema.TypeString,
+							Description: "Status of execution of upgrade post-checks",
+							Computed:    true,
+						},
+						"details": {
+							Type:        schema.TypeString,
+							Description: "Details about current execution of upgrade post-checks",
+							Computed:    true,
+						},
+						"failure_count": {
+							Type:        schema.TypeInt,
+							Description: "Total count of generated failures or warnings in last execution of upgrade post-checks",
+							Computed:    true,
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func dataSourceNsxtUpgradePostCheckRead(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+	aggregateInfoClient := upgrade_unit_groups.NewAggregateInfoClient(connector)
+	component := d.Get("type").(string)
+	timeout := d.Get("timeout").(int)
+	interval := d.Get("interval").(int)
+	delay := d.Get("delay").(int)
+
+	stateConf := &resource.StateChangeConf{
+		Pending: []string{model.UpgradeChecksExecutionStatus_STATUS_IN_PROGRESS},
+		Target:  []string{model.UpgradeChecksExecutionStatus_STATUS_COMPLETED},
+		Refresh: func() (interface{}, string, error) {
+			listRes, err := aggregateInfoClient.List(&component, nil, nil, nil, nil, nil, nil, nil)
+			if err != nil {
+				return nil, "", fmt.Errorf("failed to retrieve %s upgrade post-check results: %v", component, err)
+			}
+			for _, res := range listRes.Results {
+				if *res.PostUpgradeStatus.Status == model.UpgradeChecksExecutionStatus_STATUS_IN_PROGRESS {
+					return res.PostUpgradeStatus, model.UpgradeChecksExecutionStatus_STATUS_IN_PROGRESS, nil
+				}
+			}
+			// Return UpgradeChecksExecutionStatus_STATUS_COMPLETED here doesn't mean all UpgradeUnitGroup post-checks are completed.
+			// Just to stop polling the states as none of UpgradeUnitGroup is in in_progress state.
+			return listRes, model.UpgradeChecksExecutionStatus_STATUS_COMPLETED, nil
+		},
+		Timeout:      time.Duration(timeout) * time.Second,
+		PollInterval: time.Duration(interval) * time.Second,
+		Delay:        time.Duration(delay) * time.Second,
+	}
+
+	_, err := stateConf.WaitForState()
+	if err != nil {
+		return err
+	}
+
+	listRes, err := aggregateInfoClient.List(&component, nil, nil, nil, nil, nil, nil, nil)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve %s upgrade post-check result: %v", component, err)
+	}
+	var failedGroup []map[string]interface{}
+	for _, res := range listRes.Results {
+		failureCounts := 0
+		if res.PostUpgradeStatus.FailureCount != nil {
+			failureCounts = int(*res.PostUpgradeStatus.FailureCount)
+		} else if res.PostUpgradeStatus.NodeWithIssuesCount != nil {
+			failureCounts = int(*res.PostUpgradeStatus.NodeWithIssuesCount)
+		}
+		if res.PostUpgradeStatus.Status != nil && *res.PostUpgradeStatus.Status != model.UpgradeChecksExecutionStatus_STATUS_COMPLETED || failureCounts > 0 {
+			elem := make(map[string]interface{}, 4)
+			elem["status"] = *res.PostUpgradeStatus.Status
+			elem["failure_count"] = failureCounts
+			if res.Id != nil {
+				elem["id"] = *res.Id
+			}
+			if res.PostUpgradeStatus.Details != nil {
+				elem["details"] = *res.PostUpgradeStatus.Details
+			}
+			failedGroup = append(failedGroup, elem)
+		}
+	}
+	if len(failedGroup) > 0 {
+		d.Set("failed_group", failedGroup)
+	}
+
+	return nil
+}
--- a/nsxt/resource_nsxt_upgrade_run.go
+++ b/nsxt/resource_nsxt_upgrade_run.go
@@ -1,0 +1,724 @@
+/* Copyright © 2024 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"log"
+	"time"
+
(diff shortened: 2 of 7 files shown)


GraysonWu:
Still a few things to be added:
1. A dummy attribute to make upgrade_run depend on upgrade_prepare
2. Documentation
3. Finalize the schema of output attributes

In the meantime, we could start the review process.

nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -0,0 +1,568 @@
...
+
+var supportedUpgradeGroupExtendedConfigValue = map[string][]string{
+	"upgrade_mode":                                     {"maintenance_mode", "in_place", "stage_in_vlcm"},
+	"maintenance_mode_config_vsan_mode":                {"evacuate_all_data", "ensure_object_accessibility", "no_action"},
+	"maintenance_mode_config_evacuate_powered_off_vms": {"true", "false"},
+	"rebootless_upgrade":                               {"true", "false"},
+}
+
+var (
+	upgradeStatusCheckInterval = 2 * time.Minute
```
annakhm, line 46: should we make this configurable as part of the resource?

nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -0,0 +1,568 @@
...
+		},
+		"maintenance_mode_config_vsan_mode": {
+			Type:         schema.TypeString,
+			Description:  "Maintenance mode config vsan mode",
+			Optional:     true,
+			Computed:     compute,
+			ValidateFunc: validation.StringInSlice(supportedUpgradeGroupExtendedConfigValue["maintenance_mode_config_vsan_mode"], false),
+		},
+		"maintenance_mode_config_evacuate_powered_off_vms": {
+			Type:         schema.TypeString,
```
annakhm, line 153: Why don't we make these boolean? In order to have an "unset" option?

GraysonWu: Yes, it's one of supported options of `extended_configuration`, so it could be "unset".

annakhm: It seems to me that default exists here, and the platform has 2 behaviors rather than 3? If that's the case, I suggest to make those booleans. We can choose to set them in extended config only if different from default.

nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -0,0 +1,568 @@
...
+		settingGet.PauseOnError = &stopOnErr
+	}
+
+	_, err = settingClient.Update(component, settingGet)
+	return err
+}
+
+func runUpgrade(upgradeClientSet upgradeClientSet, d *schema.ResourceData) error {
+	for i := range upgradeComponentList {
+		component := upgradeComponentList[i]
```
annakhm, line 464: lets add log here to indicate we're starting component upgrade?

nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -0,0 +1,568 @@
...
+	err := prepareUpgrade(upgradeClientSet, d, false)
+	if err != nil {
+		return handleUpdateError("NsxtUpgradeRun", id, err)
+	}
+
+	log.Printf("[INFO] Successfully update UpgradeUnitGroup and UpgradePlanSetting. Start Upgrade.")
+
+	err = runUpgrade(upgradeClientSet, d)
+	if err != nil {
+		log.Printf("[WARN] Upgrade didn't success: %v. Refer to the state of resource for more information.", err)
```
annakhm, line 560: We don't return error here?

GraysonWu: I didn't return an error here because for example: if edge upgrade timeout, when users "update" the resource we could know if any group settings or plan settings change. If only the host upgrade setting is changed, there is no need to pause-modify-resume the edge upgrade.
The approach of only creating the resource when all upgrades are successful also sounds good. It is more intuitive and easy to maintain.
With the timeout configurable, maybe returning an error here is a better option.

GraysonWu: Ah my bad. Whether run postcheck should be able to update.

annakhm: I assumed that if, say, an edge was disabled for an upgrade, re-enabling it and applying again should trigger another upgrade cycle? Or do we not support such upgrade paths for now?

nsxt/resource_nsxt_upgrade_run.go
GraysonWu: Done.

nsxt/resource_nsxt_upgrade_run.go
GraysonWu: Make sense. Done.

nsxt/resource_nsxt_upgrade_run.go
GraysonWu: For your case, only the group that is currently in progress needs to be paused and resume. After re-think, I prefer only creating the resource when all upgrades succeed approach and have committed it.

nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -0,0 +1,631 @@
...
+		Refresh: func() (interface{}, string, error) {
+			status, err := getUpgradeStatus(upgradeClientSet.StatusClient, component)
+			if component != nil && *component == mpUpgradeGroup && isServiceUnavailableError(err) {
+				// After MP upgrade is completed, NSXT will restart and service_unavailable error will be received here.
+				return model.ComponentUpgradeStatus_STATUS_SUCCESS, model.ComponentUpgradeStatus_STATUS_SUCCESS, nil
+			}
+			if err != nil {
+				return nil, model.ComponentUpgradeStatus_STATUS_FAILED, err
+			}
+			log.Printf("[DEBUG] Current upgrade status: %s", status)
```
annakhm, line 414: Do we get any error details in case status is Failed?

GraysonWu: Yes. In this case, it will be better to append the status details in the error msg.

nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -0,0 +1,718 @@
...
+		}
+		log.Printf(completeLog)
+	}
+	return nil
+}
+
+func runPostcheck(upgradeClient nsx.UpgradeClient, d *schema.ResourceData) {
+	for i := range upgradeComponentList {
+		component := upgradeComponentList[i]
+		if component != mpUpgradeGroup {
```
annakhm, line 630: there are no post upgrade checks for MP?

website/docs/r/upgrade_run.html.markdown
```diff
@@ -0,0 +1,105 @@
...
+    enabled = false
+  }
+
+  edge_group {
+    id      = data.nsxt_edge_upgrade_group.eg2.id
+    enabled = true
+  }
+
+  host_group {
+    id       = data.nsxt_edge_upgrade_group.eg2.id
```
annakhm, line 39: this should be data.nsxt_host_upgrade_group

nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -0,0 +1,718 @@
...
+			Optional:     true,
+			ValidateFunc: validation.StringInSlice(supportedUpgradeMode, false),
+		}
+		elemSchema["maintenance_mode_config_vsan_mode"] = &schema.Schema{
+			Type:         schema.TypeString,
+			Description:  "Maintenance mode config vsan mode",
+			Optional:     true,
+			ValidateFunc: validation.StringInSlice(supportedMaintenanceModeConfigVsanMode, false),
+		}
+		elemSchema["maintenance_mode_config_evacuate_powered_off_vms"] = &schema.Schema{
```
annakhm, line 271: I think we should have a good reason to replace bool value with `true`, `false`, `unspecified`. In this case it seems that there should be a default..

nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -0,0 +1,718 @@
...
+
+func prepareUpgrade(upgradeClientSet *upgradeClientSet, d *schema.ResourceData, isCreate bool) error {
+	for i := range upgradeComponentList {
+		component := upgradeComponentList[i]
+		// Customize MP upgrade is not allowed
+		if component == mpUpgradeGroup {
+			continue
+		}
+
+		if !isCreate && !d.HasChange(componentToGroupKey[component]) && !d.HasChange(componentToSettingKey[component]) {
```
annakhm, line 370: I think `HasChange` would be true for create use case - if so, looks like this parameter is not needed in this function

website/docs/r/upgrade_run.html.markdown
```diff
@@ -0,0 +1,105 @@
...
+    * `enabled` - (Optional) Flag to indicate whether upgrade of this group is enabled or not. Default: True.
+    * `parallel` - (Optional) Upgrade method to specify whether the upgrade is to be performed in parallel or serially. Default: True.
+    * `pause_after_each_upgrade_unit` - (Optional) Flag to indicate whether upgrade should be paused after upgrade of each upgrade-unit. Default: False.
+    * `upgrade_mode` - (Optional) Upgrade mode. Supported values: `maintenance_mode`, `in_place`, `stage_in_vlcm`.
+    * `maintenance_mode_config_vsan_mode` - (Optional) Maintenance mode config of vsan mode. Supported values: `evacuate_all_data`, `ensure_object_accessibility`, `no_action`.
+    * `maintenance_mode_config_evacuate_powered_off_vms` - (Optional) Maintenance mode config of whether evacuate powered off vms. Supported values: `true`, `false`.
+    * `rebootless_upgrade` - (Optional) Flag to indicate whether to use rebootless upgrade. Default: True.
+* `edge_upgrade_setting` - (Optional) EDGE component upgrade plan setting.
+    * `parallel` - (Optional) Upgrade Method to specify whether the upgrade is to be performed serially or in parallel. Default: True.
```
salv-orlando, line 75: I think we might want to clearly distinguish the upgrade semantics of this parallel setting vs the one in edge_group. This comment applies to host_upgrade_setting as well.

website/docs/r/upgrade_run.html.markdown
GraysonWu: Done.

nsxt/resource_nsxt_upgrade_run.go
GraysonWu: Ah sorry, I forget to change this one while I do the same thing for `rebootless_upgrade`. Done.

nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -0,0 +1,718 @@
...
+		if err != nil {
+			return err
+		}
+
+		if status.Status == model.ComponentUpgradeStatus_STATUS_SUCCESS {
+			log.Printf("[WARN] %s upgrade is already succeed. Any changes on it will be ignored.", component)
+			continue
+		}
+		if status.Status == model.ComponentUpgradeStatus_STATUS_IN_PROGRESS {
+			upgradeClientSet.PlanClient.Pause()
```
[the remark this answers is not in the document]
GraysonWu: If a component upgrade is in progress, to update either UpgradeUnitGroup or UpgradePlanSetting,
we should pause it first. Update an in-flight component will receive an error from API.
Comments added to the code.

nsxt/resource_nsxt_upgrade_run.go
GraysonWu: Good catch. Done.

website/docs/d/upgrade_postcheck.html.markdown
```diff
@@ -0,0 +1,34 @@
...
+data "nsxt_upgrade_postcheck" "pc" {
+  upgrade_run_id = nsxt_upgrade_run.test.id
+  type           = "EDGE"
+}
+```
+
+## Argument Reference
+
+* `upgrade_run_id` - (Required) The ID of corresponding `nsxt_upgrade_run` resource.
+* `type` - (Required) Component type of upgrade post-checks.
```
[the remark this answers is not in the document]
GraysonWu: Done.

website/docs/r/upgrade_run.html.markdown
```diff
@@ -0,0 +1,105 @@
+---
+subcategory: "Beta"
+layout: "nsxt"
+page_title: "NSXT: nsxt_upgrade_run"
+description: A resource to configure and execute upgrade of NSXT cluster.
+---
+
+# nsxt_upgrade_run
+
+This resource provides a method to configure and execute upgrade of NSXT cluster.
```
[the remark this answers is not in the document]
GraysonWu: Sounds good. Done.

website/docs/r/upgrade_run.html.markdown
```diff
@@ -0,0 +1,105 @@
...
+
+It will first configure upgrade unit groups and upgrade plan settings for `EDGE`
+and `HOST` components. Then it will execute the upgrade for each component. If 
+there are either `EDGE` or `HOST` groups disabled from the upgrade, the corresponding 
+component will be paused after enabled groups get upgraded. At the same time, `MP` upgrade
+won't be processed, because that requires `EDGE` and `HOST` to be fully upgraded. Refer to
+the exposed attribute for current upgrade state details.
```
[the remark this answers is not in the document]
GraysonWu: Done.

website/docs/r/upgrade_run.html.markdown
```diff
@@ -0,0 +1,105 @@
...
+    post_upgrade_check = true
+  }
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `upgrade_prepare_ready_id` - (Required) ID of corresponding `nsxt_upgrade_prepare_ready` resource. Update this field will trigger replacement (destroy and create) of this resource.
```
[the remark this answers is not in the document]
GraysonWu: Done.

website/docs/r/upgrade_run.html.markdown
GraysonWu: Done.

website/docs/r/upgrade_run.html.markdown
```diff
@@ -0,0 +1,105 @@
...
+* `edge_group` - (Optional) EDGE component upgrade unit group configurations. Groups will be reordered following the order they present in this field.
+    * `id` - (Required) ID of the upgrade unit group.
+    * `enabled` - (Optional) Flag to indicate whether upgrade of this group is enabled or not. Default: True.
+    * `parallel` - (Optional) Upgrade method to specify whether the upgrade is to be performed in parallel or serially. Default: True.
+    * `pause_after_each_upgrade_unit` - (Optional) Flag to indicate whether upgrade should be paused after upgrade of each upgrade-unit. Default: False.
```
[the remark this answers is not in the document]
GraysonWu: For that component, it will be paused. The MP upgrade won't be processed, which requires EDGE and HOST upgrade to be successful. If a user wants to resume the upgrade he could re-apply this resource to trigger the upgrade.

nsxt/resource_nsxt_upgrade_run.go
GraysonWu: Yes, as you can see there is no post-check btn on UI.

nsxt/data_source_nsxt_upgrade_postcheck.go
```diff
@@ -0,0 +1,159 @@
...
+	for _, res := range listRes.Results {
+		failureCounts := 0
+		if res.PostUpgradeStatus.FailureCount != nil {
+			failureCounts = int(*res.PostUpgradeStatus.FailureCount)
+		} else if res.PostUpgradeStatus.NodeWithIssuesCount != nil {
+			failureCounts = int(*res.PostUpgradeStatus.NodeWithIssuesCount)
+		}
+		if res.PostUpgradeStatus.Status != nil && *res.PostUpgradeStatus.Status != model.UpgradeChecksExecutionStatus_STATUS_COMPLETED || failureCounts > 0 {
+			elem := make(map[string]interface{}, 4)
+			elem["status"] = *res.PostUpgradeStatus.Status
```
annakhm, line 143: sdk would do the pointer dereferencing while also checking for nil, so we usually don't dereference pointers in assignment to schema

annakhm (approved):
Just small comment, otherwise LGTM

PR #1089, opened 2024-01-26
Expose extended attribute in realization info data source

### LINKED ISSUE #793: New Attribute uplink_interface_ips for Ressource/Data-Source nsxt_policy_tier1_gateway

In order to configure VRF Route Leaking the IPv4 / IPv6 Address of the T1 Gateways uplink Interface is required. Here the link to the NSX-T Dokumentation: https://docs.vmware.com/en/VMware-NSX-T-Data-Center/3.2/administration/GUID-6DF4F3CC-17DC-4F44-ACEE-D8771D6675F1.html

At present I have no possibility to find the Uplink IPs of the T1 Gateway with Terraform, hence I have to first create a T1, find the IP in the NSX-T GUI / CLI and then add static routes with the Ressource nsxt_policy_static_route.

It would be very helpful if the Ressource and Data Source: nsxt_policy_tier1_gateway would have an attribute that returns a map with the IPv4 Address and the IPv6 Address of the tier 1 gateways uplink interface.

_No response_

### Additional context

_No response_

GraysonWu:
Fixes #793


ksamoray:
Should we be able to retrieve multiple attributes for the same resource?
Also, why is this a part of the generic realization info, instead of being part of the tier1 datasource?

GraysonWu:
> > Should we be able to retrieve multiple attributes for the same resource? Also, why is this a part of the generic realization info, instead of being part of the tier1 datasource?
> 
> If we choose to expose extended attributes for all objects, I would suggest to return the whole set as a Map type, so that the user would be able to use it as follows: `data.nsxt_policy_realization_info.info.extended_attributes["ip_address"]`

Yes, using a map could be a better option. But there might be multiple entities for a specific `entity_type`. For example, a user wants to know the IP addresses. Multiple `RealizedLogicalRouterPort` entities will be returned, and only one of it have IP addresses. We are not sure which attribute users care to decide which entity we want to return.

Our current logic is: if `entity_type` is not specified, we return the first one. If `entity_type` is specified, we return the first one matching the `entity_type`. So I added the `extended_attribute_key` to let the users specify which key they want. If multiple entities have the key, the first one will be returned.

GraysonWu:
> Should we be able to retrieve multiple attributes for the same resource? Also, why is this a part of the generic realization info, instead of being part of the tier1 datasource?

Because in API, those IP addresses are not exposed in tier1, only in realization info. But we could call realization info API behind the scenes for tier1 data source, which may also require a polling mechanism to wait for that logical router port to be in the REALIZED state. Glad to hear your ideas thx.

ksamoray:
> > Should we be able to retrieve multiple attributes for the same resource? Also, why is this a part of the generic realization info, instead of being part of the tier1 datasource?
> 
> Because in API, those IP addresses are not exposed in tier1, only in realization info. But we could call realization info API behind the scenes for tier1 data source, which may also require a polling mechanism to wait for that logical router port to be in the REALIZED state. Glad to hear your ideas thx.

Or require the user to add a dependency between the T1 data source and the realization resource if the IPs are needed - we can document this if we don't trust the user's intuition 😁

Either way these are just proposals - I'm good with any solution.

annakhm:
LGTM, but lets add a test for this data source? We could piggyback on existing test for gateway interface resource and add this data source.

annakhm (approved):

Issue #1027, opened 2023-11-10
Support GRE tunnels for Tier-0


salv-orlando:
### Is your feature request related to a problem? Please describe.

NSX 4.1.2 added the possibility of creating GRE tunnels terminated at the T0 gateway which support Static Routes,  BGP and BFD.

It would be useful to allow for configuring them via Terraform.

### Describe the solution you'd like

Add resources and datasources for configuring GRE tunnels.
API:  `/infra/tier-0s/<tier-0-id>/locale-services/<locale-svc-id>/tunnels/<tunnel-id>`

PR #1091, opened 2024-01-28
GRE tunnel resource implementation

ksamoray:
Fixes: #1027

### DIFF
--- a/nsxt/resource_nsxt_policy_tier0_gateway_gre_tunnel.go
+++ b/nsxt/resource_nsxt_policy_tier0_gateway_gre_tunnel.go
@@ -1,0 +1,409 @@
+/* Copyright © 2024 Broadcom, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/tier_0s/locale_services"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+)
+
+func resourceNsxtPolicyTier0GatewayGRETunnel() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNsxtPolicyTier0GatewayGRETunnelCreate,
+		Read:   resourceNsxtPolicyTier0GatewayGRETunnelRead,
+		Update: resourceNsxtPolicyTier0GatewayGRETunnelUpdate,
+		Delete: resourceNsxtPolicyTier0GatewayGRETunnelDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceNsxtPolicyTier0GatewayGRETunnelImport,
+		},
+		Schema: map[string]*schema.Schema{
+			"nsx_id":       getNsxIDSchema(),
+			"path":         getPathSchema(),
+			"display_name": getDisplayNameSchema(),
+			"description":  getDescriptionSchema(),
+			"revision":     getRevisionSchema(),
+			"tag":          getTagsSchema(),
+			"locale_service_path": {
+				Type:        schema.TypeString,
+				Description: "Policy path of associated Gateway Locale Service on NSX",
+				Required:    true,
+				ForceNew:    true,
+			},
+			"destination_address": {
+				Type:         schema.TypeString,
+				Description:  "Destination IPv4 address",
+				Required:     true,
+				ValidateFunc: validation.IsIPv4Address,
+			},
+			"enabled": {
+				Type:        schema.TypeBool,
+				Description: "Enable/Disable Tunnel",
+				Optional:    true,
+				Default:     true,
+			},
+			"mtu": {
+				Type:         schema.TypeInt,
+				Description:  "Maximum transmission unit",
+				Optional:     true,
+				Default:      1476,
+				ValidateFunc: validation.IntAtLeast(64),
+			},
+			"tunnel_address": {
+				Type:        schema.TypeList,
+				Description: "Tunnel Address object parameter",
+				Required:    true,
+				MinItems:    1,
+				MaxItems:    8,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"edge_path": {
+							Type:         schema.TypeString,
+							Description:  "Policy edge node path",
+							Required:     true,
+							ValidateFunc: validatePolicyPath(),
+						},
+						"source_address": {
+							Type:         schema.TypeString,
+							Description:  "IPv4 source address",
+							Required:     true,
+							ValidateFunc: validateSingleIP(),
+						},
+						"tunnel_interface_subnet": {
+							Type:        schema.TypeList,
+							Description: "Interface Subnet object parameter",
+							Required:    true,
+							MinItems:    1,
+							MaxItems:    2,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"ip_addresses": {
+										Type:        schema.TypeList,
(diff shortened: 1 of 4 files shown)


nsxt/resource_nsxt_policy_tier0_gateway_gre_tunnel.go
```diff
@@ -0,0 +1,408 @@
...
+			"locale_service_path": {
+				Type:        schema.TypeString,
+				Description: "Policy path of associated Gateway Locale Service on NSX",
+				Required:    true,
+			},
+			"destination_address": {
+				Type:         schema.TypeString,
+				Description:  "Destination IPv4 address",
+				Required:     true,
+				ValidateFunc: validateSingleIP(),
```
annakhm, line 45: If this is IPv4 only, there is `validation.IsIPv4Address` helper. Not sure the description is accurate though

nsxt/resource_nsxt_policy_tier0_gateway_gre_tunnel.go
```diff
@@ -0,0 +1,408 @@
...
+			"nsx_id":       getNsxIDSchema(),
+			"path":         getPathSchema(),
+			"display_name": getDisplayNameSchema(),
+			"description":  getDescriptionSchema(),
+			"revision":     getRevisionSchema(),
+			"tag":          getTagsSchema(),
+			"locale_service_path": {
+				Type:        schema.TypeString,
+				Description: "Policy path of associated Gateway Locale Service on NSX",
+				Required:    true,
```
annakhm, line 39: ForceNew?

website/docs/r/policy_tier0_gateway_gre_tunnel.html.markdown
```diff
@@ -0,0 +1,137 @@
...
+  mtu            = 1500
+  gateway_path   = nsxt_policy_tier0_gateway.test.path
+  segment_path   = nsxt_policy_vlan_segment.test.path
+  edge_node_path = data.nsxt_policy_edge_node.EN.path
+  subnets        = ["192.168.240.10/24"]
+  enable_pim     = "true"
+  urpf_mode      = "STRICT"
+
+
+  tag {
```
annakhm, line 59: since this is a resource for dependency,  I think we should provide minimal working config (tags etc. not needed)

website/docs/r/policy_tier0_gateway_gre_tunnel.html.markdown
```diff
@@ -0,0 +1,130 @@
+---
+subcategory: "Gateways and Routing"
+layout: "nsxt"
+page_title: "NSXT: nsxt_policy_tier0_gateway_gre_tunnel"
+description: A resource to configure a GRE tunnel on Tier-0 gateway on NSX Policy manager.
+---
+
+# nsxt_policy_tier0_gateway_gre_tunnel
+
+This resource provides a method for the management of a Tier-0 gateway GRE tunnel. Note that edge cluster and in interface must be configured on Tier-0 Gateway in order to configure GRE tunnels on it.
```
salv-orlando, line 10: User-level comment: What's the "in" interface? It's not a term I'm familiar with for Tier-0 gateways

website/docs/r/policy_tier0_gateway_gre_tunnel.html.markdown
```diff
@@ -0,0 +1,130 @@
...
+    cidr = "10.2.2.2/24"
+  }
+}
+data "nsxt_policy_edge_node" "EN" {
+  edge_cluster_path = data.nsxt_policy_edge_cluster.EC.path
+  member_index      = 0
+}
+
+resource "nsxt_policy_tier0_gateway" "test" {
+  display_name = "terraform-acctest-resource-74743"
```
salv-orlando, line 39: pedant nit: For documentation I would use simple names. I think you might have extracted those from an acceptance test run?

website/docs/r/policy_tier0_gateway_gre_tunnel.html.markdown
```diff
@@ -0,0 +1,130 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `display_name` - (Required) Display name of the resource.
+* `description` - (Optional) Description of the resource.
+* `tag` - (Optional) A list of scope + tag pairs to associate with this resource.
```
salv-orlando, line 95: Is this "tags" (plural)?

nsxt/resource_nsxt_policy_tier0_gateway_gre_tunnel.go
```diff
@@ -0,0 +1,409 @@
...
+		return handleCreateError("GreTunnel", id, err)
+	}
+	if !isT0 {
+		return fmt.Errorf("locale service path %s is not associated with a tier0 router", localeServicePath)
+	}
+	client := locale_services.NewTunnelsClient(connector)
+
+	obj, err := tier0GatewayGRETunnelFromSchema(d)
+	if err != nil {
+		return fmt.Errorf("failed to create GRE Tunnel: %v", err)
```
salv-orlando, line 260: nit abou the error message: we have not yet tried the creation here, so maybe the error should state "unable to create the GRE tunnel" or something like that

website/docs/r/policy_tier0_gateway_gre_tunnel.html.markdown
ksamoray: Should be "an interface" :)

website/docs/r/policy_tier0_gateway_gre_tunnel.html.markdown
ksamoray: Exactly... Will replace the name.

website/docs/r/policy_tier0_gateway_gre_tunnel.html.markdown
ksamoray: It's always `tag` in all our resources.
And as a general rule - we use singular naming for complex objects as they appear in HCL as a sequence of individual objects, e.g [here](https://github.com/vmware/terraform-provider-nsxt/blob/master/website/docs/r/policy_ip_pool.html.markdown?plain=1#L20) or [here with rule](https://github.com/vmware/terraform-provider-nsxt/blob/master/website/docs/r/policy_predefined_security_policy.html.markdown?plain=1#L34).
For list of simple types (e.g string) we use plural naming, e.g [here](https://github.com/vmware/terraform-provider-nsxt/blob/master/website/docs/r/policy_group.html.markdown?plain=1#L42) with ip_addresses.

website/docs/r/policy_tier0_gateway_gre_tunnel.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+  subnets        = ["192.168.240.10/24"]
+  enable_pim     = "true"
+}
+
+data "nsxt_policy_realization_info" "realization_info" {
+  path = nsxt_policy_tier0_gateway_interface.test.path
+}
+data "nsxt_policy_gateway_locale_service" "test" {
+  gateway_path = nsxt_policy_tier0_gateway.test.path
+  depends_on   = [data.nsxt_policy_realization_info.realization_info]
```
annakhm, line 61: Does the config fail without this? Policy intent should not depend on realization state..

ksamoray: IDK - tests pass without it. But would I be able to retrieve locale id while related stuff isn't realized?

website/docs/r/policy_tier0_gateway_gre_tunnel.html.markdown
```diff
@@ -0,0 +1,128 @@
...
+  type           = "EXTERNAL"
+  mtu            = 1500
+  gateway_path   = nsxt_policy_tier0_gateway.test.path
+  segment_path   = nsxt_policy_vlan_segment.test.path
+  edge_node_path = data.nsxt_policy_edge_node.EN.path
+  subnets        = ["192.168.240.10/24"]
+  enable_pim     = "true"
+}
+
+data "nsxt_policy_realization_info" "realization_info" {
```
annakhm, line 56: This is not needed anymore

website/docs/r/policy_tier0_gateway_gre_tunnel.html.markdown
annakhm: I don't see why not.. everything is on policy level

website/docs/r/policy_tier0_gateway_gre_tunnel.html.markdown
ksamoray: Indeed, it's useless.

annakhm (approved):

PR #1092, opened 2024-01-29
Add the option to remove NSX during deletion

ksamoray:
While removing a host transport node collection, allow cleaning up the NSX bits from the hypervisors.
Also set revision attribute on update to allow updating.

### DIFF
--- a/nsxt/resource_nsxt_policy_host_transport_node_collection.go
+++ b/nsxt/resource_nsxt_policy_host_transport_node_collection.go
@@ -6,13 +6,18 @@ package nsxt
 import (
 	"fmt"
 	"log"
+	"time"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/fabric/compute_collections"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/sites/enforcement_points"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 )
+
+const removeOnDestroyDefault = false
 
 func resourceNsxtPolicyHostTransportNodeCollection() *schema.Resource {
 	return &schema.Resource{
@@ -88,6 +93,12 @@ func resourceNsxtPolicyHostTransportNodeCollection() *schema.Resource {
 				Type:        schema.TypeString,
 				Optional:    true,
 				Description: "Transport Node Profile Path",
+			},
+			"remove_nsx_on_destroy": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Indicate whether NSX service should be removed from hypervisors during resource deletion",
+				Default:     removeOnDestroyDefault,
 			},
 		},
 	}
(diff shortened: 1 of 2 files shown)


annakhm:
Since this attribute is not set in `Read` function, I think we need to set it explicitly in state during `import`, otherwise it will cause a diff. It would be best to verify this with actual import, since we don't have a test for this resource.

ksamoray:
> Since this attribute is not set in `Read` function, I think we need to set it explicitly in state during `import`, otherwise it will cause a diff. It would be best to verify this with actual import, since we don't have a test for this resource.

Yes import is an issue - IDK if we can retrieve the values which are in the script (obviously there's no value in NSX).
I could persist this value in a tag in NSX - IDK if we do that elsewhere, it would be ugly as this is visible to the user.

Any other option?

nsxt/resource_nsxt_policy_host_transport_node_collection.go
```diff
@@ -267,6 +277,14 @@ func resourceNsxtPolicyHostTransportNodeCollectionDelete(d *schema.ResourceData,
 		return err
 	}
 
+	removeNsxOnDestroy := d.Get("remove_nsx_on_destroy").(bool)
+	if removeNsxOnDestroy {
+		log.Printf("[INFO] Removing NSX from hosts associated with HostTransportNodeCollection with ID %s", id)
+		err = client.Removensx(siteID, epID, id)
```
salv-orlando, line 283: Is this API call synchronous or asynchronous?
I seem to recall the latter.

I'm trying to look at the problem from a different perspective.
I wonder if this attribute should set an expectation in the user that NSX will be uninstalled from nodes.
In that case we might need to find a way to confirm deletion was done successfully; if not fail the TF apply operation, and the TN collection should be kept in the state.

Does this make sense?

nsxt/resource_nsxt_policy_host_transport_node_collection.go
```diff
@@ -89,6 +89,12 @@ func resourceNsxtPolicyHostTransportNodeCollection() *schema.Resource {
 				Optional:    true,
 				Description: "Transport Node Profile Path",
 			},
+			"remove_nsx_on_destroy": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Indicate whether NSX service should be removed from hypervisors during resource deletion",
```
salv-orlando, line 95: nit for description:

Upon deletion, uninstall NSX from Transport Node Collection member hosts.

salv-orlando:
> > Since this attribute is not set in `Read` function, I think we need to set it explicitly in state during `import`, otherwise it will cause a diff. It would be best to verify this with actual import, since we don't have a test for this resource.
> 
> Yes import is an issue - IDK if we can retrieve the values which are in the script (obviously there's no value in NSX). I could persist this value in a tag in NSX - IDK if we do that elsewhere, it would be ugly as this is visible to the user.
> 
> Any other option?

I think the tag might be ok. This will be a first (to the best of my knowledge) for the TF provider, but it is rather common for other integrations. It will be very important to properly document the tag we will be using.

annakhm:
> > > Since this attribute is not set in `Read` function, I think we need to set it explicitly in state during `import`, otherwise it will cause a diff. It would be best to verify this with actual import, since we don't have a test for this resource.
> > 
> > 
> > Yes import is an issue - IDK if we can retrieve the values which are in the script (obviously there's no value in NSX). I could persist this value in a tag in NSX - IDK if we do that elsewhere, it would be ugly as this is visible to the user.
> > Any other option?
> 
> I think the tag might be ok. This will be a first (to the best of my knowledge) for the TF provider, but it is rather common for other integrations. It will be very important to properly document the tag we will be using.

I suggest assigning default value upon import, and documenting this. If the user wants to change it, they will need to change config and reapply. In my opinion using tags for this is an overkill.

nsxt/resource_nsxt_policy_host_transport_node_collection.go
ksamoray: This call is asynchronous. However, upon calling this API, the host transport node collection state becomes unavailable.
We could busy-wait on the associated compute collection status until removal is complete. Does that make sense?

ksamoray:
> > > > Since this attribute is not set in `Read` function, I think we need to set it explicitly in state during `import`, otherwise it will cause a diff. It would be best to verify this with actual import, since we don't have a test for this resource.
> > > 
> > > 
> > > Yes import is an issue - IDK if we can retrieve the values which are in the script (obviously there's no value in NSX). I could persist this value in a tag in NSX - IDK if we do that elsewhere, it would be ugly as this is visible to the user.
> > > Any other option?
> > 
> > 
> > I think the tag might be ok. This will be a first (to the best of my knowledge) for the TF provider, but it is rather common for other integrations. It will be very important to properly document the tag we will be using.
> 
> I suggest assigning default value upon import, and documenting this. If the user wants to change it, they will need to change config and reapply. In my opinion using tags for this is an overkill.

Sounds like a reasonable solution. Using a tag are hackish and if a user insist on non-default value after import, the config can be reapplied.

annakhm (approved):

PR #1094, opened 2024-01-30
Set edge_node data source member_index attribute

ksamoray:
When retrieving an edge node using the data source, set member_index attribute into state.


salv-orlando (approved):
LGTM

PR #1095, opened 2024-01-30
Increase timeout for IP allocation realization

ksamoray:
Tests fail occasionally due to timing out while waiting for realization of the IP allocation.

Set higher timeout than the default to avoid this failure.

### DIFF
--- a/nsxt/resource_nsxt_policy_ip_address_allocation_test.go
+++ b/nsxt/resource_nsxt_policy_ip_address_allocation_test.go
@@ -316,7 +316,8 @@ resource "nsxt_policy_ip_address_allocation" "test" {
 }
 
 data "nsxt_policy_realization_info" "realization_info" {
-  path = nsxt_policy_ip_address_allocation.test.path
+  timeout = 6000
+  path    = nsxt_policy_ip_address_allocation.test.path
 }`, attrMap["display_name"], attrMap["description"])
 }
 


nsxt/resource_nsxt_policy_ip_address_allocation_test.go
```diff
@@ -316,7 +316,8 @@ resource "nsxt_policy_ip_address_allocation" "test" {
 }
 
 data "nsxt_policy_realization_info" "realization_info" {
-  path = nsxt_policy_ip_address_allocation.test.path
+  timeout = 6000
```
salv-orlando, line 319: is this timeout in seconds or milliseconds?

ksamoray: Hmm seems like these are [seconds](https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/data_source_nsxt_policy_realization_info.go#L145). 
As we default this attribute to [1200](https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/data_source_nsxt_policy_realization_info.go#L57), I was assuming that they're msecs.

So I guess that this PR fixes nothing, I'll look elsewhere.

salv-orlando: Sounds good! There might be cases where IP realization is delayed quite a lot, but 6000 seconds is really too much. The API will also return realization failure in case there is an error while allocating the IP address, such as pool exhaustion.

annakhm: We need to add realization error printout to debug this failure.

ksamoray: We have this [printout](https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/resource_nsxt_policy_ip_address_allocation.go#L177), it's the error that shows up in the tests:
```
    resource_nsxt_policy_ip_address_allocation_test.go:96: Step 1/4 error: Error running apply: exit status 1
        
        Error: Failed to get realized IP for path /infra/ip-pools/7d507c81-457f-482c-b868-2ec2893de029/ip-allocations/f0354ca4-5600-437a-8c52-fb2307a8abb3
        
          with nsxt_policy_ip_address_allocation.test,
          on terraform_plugin_test.tf line 19, in resource "nsxt_policy_ip_address_allocation" "test":
          19: resource "nsxt_policy_ip_address_allocation" "test" {
```

Maybe I'll turn on extra logging on the CI to troubleshoot this.

PR #1101, opened 2024-02-08
Add NSX version check to cluster IP resource

annakhm:
In addition, improve error reporting


2ez4szliu:
From the changes it seems that virtual IP is only supported on NSX >= 4.0.0, do we want to also add this to the doc?

ksamoray (approved):
LGTM

2ez4szliu (approved):

PR #1103, opened 2024-02-08
Handle errors on manager_cluster resource properly

ksamoray:
Output messages weren't outputed properly from provider.


nsxt/resource_nsxt_manager_cluster.go
```diff
@@ -356,7 +356,7 @@ func joinNodeToCluster(clusterID string, certSha256Thumbprint string, guestNode
 	}
 	_, err = client.Joincluster(joinClusterParams)
 	if err != nil {
-		return fmt.Errorf("Failed to join node to cluster: %s, node ip address: %s", err, guestNode.IPAddress)
+		return logAPIError(fmt.Sprintf("Failed to join node to cluster: %s, node ip address: %s", clusterID, guestNode.IPAddress), err)
```
annakhm, line 359: `Failed to join node` test is repeated on Create (but not on Update), maybe lets remove the duplicate text from Create flow?

ksamoray: Yeah makes sense

annakhm (approved):

PR #1104, opened 2024-02-08
[WIP]: Implement policy site resource

ksamoray:


nsxt/resource_nsxt_policy_site.go
```diff
@@ -0,0 +1,267 @@
...
+
+	if isNotFoundError(err) {
+		return false, nil
+	}
+
+	return false, logAPIError("Error retrieving resource", err)
+}
+
+func resourceNsxtPolicySiteCreate(d *schema.ResourceData, m interface{}) error {
+	id, err := getOrGenerateID(d, m, resourceNsxtPolicySiteExists)
```
annakhm, line 166: can we add here: 
        if !isPolicyGlobalManager(m) {
                return globalManagerOnlyError()
        }

nsxt/resource_nsxt_policy_site.go
```diff
@@ -0,0 +1,267 @@
...
+							Description: "Username",
+						},
+					},
+				},
+			},
+			"site_type": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Description:  "Persistent Site Type",
+				ValidateFunc: validation.StringInSlice(siteTypeValues, false),
```
annakhm, line 94: No default here? What happens if this is unspecified? Should the value be Computed in this case?

ksamoray: [Spec](https://vdc-download.vmware.com/vmwb-repository/dcr-public/8e2a4b1a-659f-4be6-a4f4-7ca42b04cb84/9a2c1372-7174-40f0-89d2-27615bfd5d08/api_includes/types_Site.html) doesn't specify a default here.
As UI in GM allows only on-prem addition, maybe we could discard this attribute entirely?
Screenshot 2024-02-15 at 12 10 55

annakhm: I think this depends on env, and perhaps default depends on env as well. If NSX auto-assigns this value, we can define it as `Computed`.

ksamoray: Turns out this this is required...
```
│ Error:  Failed to create Site f5bb6846-cf9d-4efd-9cd6-4d11de593fff: Field level validation errors: {value  of property site_type is not one of the allowed values [ONPREM_LM, SDDC_LM]} (code 255)
```

website/docs/r/policy_site.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+* `fail_if_rtep_misconfigured` - (Optional) Fail onboarding if RTEPs misconfigured. Default is true.
+* `fail_if_rtt_exceeded` - (Optional) Fail onboarding if maximum RTT exceeded. Default is true.
+* `maximum_rtt` - (Optional) Maximum acceptable packet round trip time (RTT). Default is 250.
+* `site_connection_info` - (Optional) Connection information.
+  * `fqdn` - (Optional) Fully Qualified Domain Name of the Management Node.
+  * `password` - (Optional) Password.
+  * `site_uuid` - (Optional) ID of Site.
+  * `thumbprint` - (Optional) Thumbprint of Enforcement Point.
+  * `username` - (Optional) Username.
+* `site_type` - (Required) Persistent Site Type. Allowed values are ONPREM_LM, SDDC_LM. Default is ONPREM_LM.
```
annakhm, line 46: It doesn't have a default according to the code (and its `Required`). Also lets surround the values in `` for consistency.

ksamoray: It's required in the code? The [spec](https://vdc-download.vmware.com/vmwb-repository/dcr-public/8e2a4b1a-659f-4be6-a4f4-7ca42b04cb84/9a2c1372-7174-40f0-89d2-27615bfd5d08/api_includes/types_Site.html) doesn't enlist it as so.

website/docs/r/policy_site.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+
+The following arguments are supported:
+
+* `display_name` - (Required) Display name of the resource.
+* `description` - (Optional) Description of the resource.
+* `tag` - (Optional) A list of scope + tag pairs to associate with this resource.
+* `nsx_id` - (Optional) The NSX ID of this resource. If set, this ID will be used to create the resource.
+* `fail_if_rtep_misconfigured` - (Optional) Fail onboarding if RTEPs misconfigured. Default is true.
+* `fail_if_rtt_exceeded` - (Optional) Fail onboarding if maximum RTT exceeded. Default is true.
+* `maximum_rtt` - (Optional) Maximum acceptable packet round trip time (RTT). Default is 250.
```
salv-orlando, line 39: The NSX API guide does not mention it, but I think we should specify the unit of measure.
msecs, I believe.

nsxt/resource_nsxt_policy_site.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+}
+
+func getSiteFromSchema(d *schema.ResourceData) gm_model.Site {
+	displayName := d.Get("display_name").(string)
+	description := d.Get("description").(string)
+	var tags []gm_model.Tag
+	tagListIntface := d.Get("tag")
+	if tagListIntface != nil {
+		tagList := tagListIntface.(*schema.Set).List()
```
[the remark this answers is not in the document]
ksamoray: We have more than one utility, as the SDK has different `model.Tag` structs for policy/MP/GM APIs. However there's no util func for GM - I could add one or convert to policy model and then use TypeConverter (we do this when there's a resource which runs for both M and policy).

website/docs/r/policy_site.html.markdown
ksamoray: IDK which are the units as I couldn't find that attribute in the UI

nsxt/resource_nsxt_policy_site.go
```diff
@@ -0,0 +1,270 @@
...
+func resourceNsxtPolicySiteUpdate(d *schema.ResourceData, m interface{}) error {
+	id := d.Id()
+	if id == "" {
+		return fmt.Errorf("Error obtaining Site ID")
+	}
+
+	connector := getPolicyConnector(m)
+	client := global_infra.NewSitesClient(connector)
+
+	obj := getSiteFromSchema(d)
```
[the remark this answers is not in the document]
ksamoray: I didn't try to set it (I need an extra LM to test this). But connectivity info is updatable in UI.

nsxt/resource_nsxt_policy_site.go
```diff
@@ -0,0 +1,270 @@
...
+	return &schema.Resource{
+		Create: resourceNsxtPolicySiteCreate,
+		Read:   resourceNsxtPolicySiteRead,
+		Update: resourceNsxtPolicySiteUpdate,
+		Delete: resourceNsxtPolicySiteDelete,
+		Importer: &schema.ResourceImporter{
+			State: nsxtPolicyPathResourceImporter,
+		},
+
+		Schema: map[string]*schema.Schema{
```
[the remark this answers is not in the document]
ksamoray: Haven't found use for those yet, usually we add read-only attrs only if they're used as reference anywhere else or just seem very useful. As I haven't complete yet the federation object implementation, I might find their use later on maybe. Anyway it's always easier to add attributes than removing them.

annakhm (approved):

PR #1105, opened 2024-02-08
Datasource for transport node collection and state

### LINKED ISSUE #1090: Datasource for transport node collections state

As Luca is testing fabric resources, the problem came that segment resources create the VLAN segments but because the bits are not yet installed on the hosts, the corresponding dvpg by the same name is not available in vSphere. Currently there is no data source for checking the status of the installation.

Add data source for `transport-node-collection/~/state` with `state` info, and `aggregate_progress_percentage` if in progress, failure otherwise.

`compute-collections/~/member-status`

### Additional context

_No response_

qiyueyao:
Adds a data source for Policy Host Transport Node Collection and its state.

Fixes #1090


nsxt/data_source_nsxt_policy_host_transport_node_collection_state.go
```diff
@@ -0,0 +1,132 @@
...
+
+	objID := d.Get("id").(string)
+	objName := d.Get("display_name").(string)
+	objSitePath := d.Get("site_path").(string)
+	// For local manager, if site path is not provided, use default site
+	if objSitePath == "" {
+		objSitePath = defaultSite
+	}
+	if objID == "" && objName == "" {
+		return fmt.Errorf("please specify id or display_name in order to identify Transport Node Collection")
```
annakhm, line 96: Wouldn't `site_path` be enough to identify transport node collection, in case there is single collection per site? In this case, I would suggest to allow not to specify name or id.

nsxt/data_source_nsxt_policy_host_transport_node_collection_state.go
```diff
@@ -0,0 +1,132 @@
...
+		converter := bindings.NewTypeConverter()
+		dataValue, errors := converter.ConvertToGolang(obj, model.HostTransportNodeCollectionBindingType())
+		if len(errors) > 0 {
+			return errors[0]
+		}
+		hostTransportNodeCollectionResource := dataValue.(model.HostTransportNodeCollection)
+		objID = *hostTransportNodeCollectionResource.Id
+	}
+
+	objState, err := client.Get(objSitePath, getPolicyEnforcementPoint(m), objID)
```
annakhm, line 118: Would it make sense to move this under `else`? If we already have the object in hand from the search API, no need to pull it again.

nsxt/data_source_nsxt_policy_host_transport_node_collection_state.go
qiyueyao: Removed the restraint.

annakhm (approved):

qiyueyao:
Rebased and made some modifications to fix issues with specific cases, by providing display name. Seeking another review, thanks!

ksamoray:
Hi, in the *_state datasource, we'll need the DS to busy-wait until the state is "success" or whatever - see an example [here](https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/data_source_nsxt_compute_manager_realization.go#L140)

qiyueyao:
> Hi, in the *_state datasource, we'll need the DS to busy-wait until the state is "success" or whatever - see an example [here](https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/data_source_nsxt_compute_manager_realization.go#L140)

Got it, I'll change it to realization instead of merely returning state.

ksamoray:
> > Hi, in the *_state datasource, we'll need the DS to busy-wait until the state is "success" or whatever - see an example [here](https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/data_source_nsxt_compute_manager_realization.go#L140)
> 
> Got it, I'll change it to realization including wait. Found that the existing `realization` data sources only returns state, but not the error messages, do we need that for this data source? @ksamoray

I've got no opinion about - we output the errors AFAIK. @annakhm - got any idea whether we need to?
TBH I was never sure why we have that state - it'll always be `SUCCESS` as otherwise we fail the apply action.

Anyway I tested the DS - it does work so if @annakhm has no further comments, let's merge.

ksamoray (approved):

annakhm (approved):

qiyueyao:
> I've got no opinion about - we output the errors AFAIK. @annakhm - got any idea whether we need to? TBH I was never sure why we have that state - it'll always be `SUCCESS` as otherwise we fail the apply action.

Actually I encountered a `FAILED_TO_REALIZE` situation, when I messed the deployments, with validation error given _"8345: Cannot update TransportNode/... to remove TransportZone/... as this TransportZone has LogicalSwitches with 3 Vif Attachment(s). Please use the object name or UUID in Global Search to find all linked objects."_ Might be useful.
Merging the PR.

PR #1106, opened 2024-02-11
Make domain name unique to address test failures

ksamoray:
Test fail occasionally due to delays in deletion of domains in global manager topologies.


nsxt/resource_nsxt_policy_gateway_policy_test.go
```diff
@@ -386,7 +386,7 @@ func TestAccResourceNsxtGlobalPolicyGatewayPolicy_withSite(t *testing.T) {
 func TestAccResourceNsxtGlobalPolicyGatewayPolicy_withDomain(t *testing.T) {
 	name := "terraform-test"
 	siteName := getTestSiteName()
-	domainName := "new-domain"
+	domainName := fmt.Sprintf("new-domain-%d", getRandomSuffixInt())
```
annakhm, line 389: can we use `getAccTestResourceName()` for consistency with other tests?

ksamoray: Sure, I thought that there's some reason that it wasn't used :)

annakhm (approved):

PR #1107, opened 2024-02-11
Handle edge_node import failures

ksamoray:
Some optional attributes aren't returned from NSX with v4.2.0 backend, the PR handles this matter.

### DIFF
--- a/nsxt/resource_nsxt_edge_transport_node.go
+++ b/nsxt/resource_nsxt_edge_transport_node.go
@@ -1232,14 +1232,18 @@ func resourceNsxtEdgeTransportNodeRead(d *schema.ResourceData, m interface{}) er
 	}
 	node := base.(model.EdgeNode)
 
-	err = setEdgeDeploymentConfigInSchema(d, node.DeploymentConfig)
-	if err != nil {
-		return handleReadError(d, "TransportNode", id, err)
+	if node.DeploymentConfig != nil {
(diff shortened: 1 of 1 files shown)


nsxt/resource_nsxt_edge_transport_node.go
```diff
@@ -1232,14 +1232,18 @@ func resourceNsxtEdgeTransportNodeRead(d *schema.ResourceData, m interface{}) er
 	}
 	node := base.(model.EdgeNode)
 
-	err = setEdgeDeploymentConfigInSchema(d, node.DeploymentConfig)
-	if err != nil {
-		return handleReadError(d, "TransportNode", id, err)
+	if node.DeploymentConfig != nil {
+		err = setEdgeDeploymentConfigInSchema(d, node.DeploymentConfig)
```
salv-orlando, line 1236: is this optional?

salv-orlando (approved):
just a question about DeploymentConfig being optional or not.

nsxt/resource_nsxt_edge_transport_node.go
ksamoray: An edge can be deployed as an OVA or directly on HW, right? I think that `deployment_config` is only populated when NSX deploys an edge node on a compute manager.

salv-orlando: got it!

PR #1109, opened 2024-02-12
Documentation fixes

ksamoray:
For compute manager, lookup attribute was missing from example.
For host transport node resources, edit the import part to match the same convention used elsewhere.
For IPSec VPN service, replace deprecated attribute with supported one.


salv-orlando (approved):
LGTM

annakhm (approved):

PR #1111, opened 2024-02-12
Add validators to transport node resource

annakhm:
When user provides empty strings or empty lists, NSX fails the request without specifying what exactly is missing. It is preferrable to fail those invalid configs on plan stage.

### DIFF
--- a/nsxt/resource_nsxt_edge_transport_node.go
+++ b/nsxt/resource_nsxt_edge_transport_node.go
@@ -558,9 +558,10 @@ func getTransportZoneEndpointSchema() *schema.Schema {
 		Elem: &schema.Resource{
 			Schema: map[string]*schema.Schema{
 				"transport_zone": {
-					Type:        schema.TypeString,
-					Required:    true,
-					Description: "Unique ID identifying the transport zone for this endpoint",
+					Type:         schema.TypeString,
+					Required:     true,
+					Description:  "Unique ID identifying the transport zone for this endpoint",
+					ValidateFunc: validation.StringIsNotWhiteSpace,
 				},
 				"transport_zone_profiles": {
 					Type:        schema.TypeList,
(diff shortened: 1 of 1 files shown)


nsxt/resource_nsxt_edge_transport_node.go
```diff
@@ -1519,7 +1527,7 @@ func setIPAssignmentInSchema(spec *data.StructValue) (interface{}, error) {
 	return []interface{}{elem}, nil
 }
 
-func setHostSwitchProfileIDsInSchema(hspIDs []model.HostSwitchProfileTypeIdEntry) interface{} {
+func setHostSwitchProfileIDsInSchema(hspIDs []model.HostSwitchProfileTypeIdEntry) []interface{} {
```
GraysonWu, line 1530: To double check: this is fixing a different issue than the PR description, right? 
(I'm kind of surprised that it could work before)

GraysonWu:
> When user provides empty strings or empty lists, NSX fails the request without specifying what exactly is missing. It is preferrable to fail those invalid configs on plan stage.

I'm wondering if the same issue also exists in other resources.

nsxt/resource_nsxt_edge_transport_node.go
ksamoray: Why wouldn't it work? `interface{}` can be anything, including `[]interface{}`. And before it wasn't iterated anywhere, just assigned into the schema dicts.
As this PR iterates for validation, it requires this change.

ksamoray (approved):
LGTM

nsxt/resource_nsxt_edge_transport_node.go
salv-orlando: I think `interace{}` is so generic that even a slice will satisfy it.
`[]interface{}` is a bit more strongly typed as only a slice will satisfy it.

So the change is probably unrelated, but I think a good thing to add.

salv-orlando (approved):
LGTM

annakhm:
> > When user provides empty strings or empty lists, NSX fails the request without specifying what exactly is missing. It is preferrable to fail those invalid configs on plan stage.
> 
> I'm wondering if the same issue also exists in other resources.

I'm reluctant to apply this validator to released resources, since this can break backwards compatibility. But we can go over  resources in `Beta` and improve validation there.

GraysonWu (approved):

PR #1114, opened 2024-02-13
Add debug capabilities to main.go

ksamoray:
Add some necessary code for execution of the provider with a debugger, e.g delve.
More information about using debuggers is available [here](https://developer.hashicorp.com/terraform/plugin/debugging) and [here](https://opencredo.com/blogs/running-a-terraform-provider-with-a-debugger/).

### DIFF
--- a/main.go
+++ b/main.go
@@ -4,15 +4,33 @@
 package main
 
 import (
+	"context"
+	"flag"
+	"log"
+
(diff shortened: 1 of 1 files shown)


salv-orlando (commented):
LGTM, but I would check it after we release 3.5.0

annakhm (approved):

PR #1116, opened 2024-02-15
Fix port validation in compute manager

annakhm:
Since port is int rather than string, int validation is required


ksamoray (approved):
LGTM

PR #1121, opened 2024-02-18
Fix node user password misbehavior

ksamoray:
Node user tests were not specifying a password during update and therefore were failing.

### DIFF
--- a/nsxt/resource_nsxt_node_user.go
+++ b/nsxt/resource_nsxt_node_user.go
@@ -37,10 +37,11 @@ func resourceNsxtUsers() *schema.Resource {
 				Computed:    true,
 			},
 			"password": {
-				Type:        schema.TypeString,
-				Description: "Password for the user",
-				Sensitive:   true,
-				Optional:    true,
+				Type:         schema.TypeString,
+				Description:  "Password for the user",
+				Sensitive:    true,
+				Optional:     true,
+				ValidateFunc: validation.StringIsNotWhiteSpace,
 			},
 			"password_change_frequency": {
 				Type:         schema.TypeInt,
--- a/nsxt/resource_nsxt_node_user_test.go
+++ b/nsxt/resource_nsxt_node_user_test.go
@@ -60,7 +60,23 @@ func TestAccResourceNsxtNodeUser_basic(t *testing.T) {
 				),
 			},
 			{
-				Config: testAccNodeUserUpdate(testUsername),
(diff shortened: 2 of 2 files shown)


nsxt/resource_nsxt_node_user.go
```diff
@@ -212,7 +212,7 @@ func resourceNsxtNodeUserUpdate(d *schema.ResourceData, m interface{}) error {
 	}
 
 	// If password is changed, handle password change.
-	if password != oldPassword {
+	if password != oldPassword && password != "" {
```
annakhm, line 215: should the resource error out if password is empty?

ksamoray: So what happens is as follows:
When updating in the test [here](https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/resource_nsxt_node_user_test.go#L63), password isn't specified at all within the test. This is legit, I think, and therefore the test fails as old_password != new_password.
So the change ignores it - I don't think that the provider should fail in that case.

ksamoray: But wouldn't that overwrite the password in the state?

annakhm: Shouldn't we nullify the password on NSX in this case? If empty pwd is allowed?

ksamoray: Password cannot be nullified. So maybe the test should be fixed in this case? As the test doesn't set the password to the same as the existing one.

annakhm: My take is that test should be fixed, and possibly validation code added in the resource?

nsxt/resource_nsxt_node_user_test.go
```diff
@@ -181,14 +197,15 @@ resource "nsxt_node_user" "test" {
 }`, attrMap["active"], attrMap["full_name"], attrMap["password"], username, attrMap["password_change_frequency"], attrMap["password_change_warning"])
 }
 
-func testAccNodeUserUpdate(username string) string {
+func testAccNodeUserUpdate(username, password string) string {
```
annakhm, line 200: sorry for pedant comment, but can be just take `password` from `attrMap` without passing it as param here?

ksamoray: @annakhm note that there are two update cases, with and without password change (as that requires some extra logic). If we use the password from attrMap we need an indication whether password should change or not.

annakhm (approved):

Issue #1097, opened 2024-01-31
Multi-TEP High Availability


msyretis:
### Is your feature request related to a problem? Please describe.

Multi-TEP High Availability
https://docs.vmware.com/en/VMware-NSX/4.1/administration/GUID-72C2F2B2-7DC6-49A2-AD74-2FBAC93E3FAC.html

I would like if possible to be able to create the profile and reference it in the transport node profile.
this saves me a manual step of doing this by hand/ansible.

### Describe the solution you'd like

a data and resource block for the TEP HA host switch profile

```
{
"enabled": "true",
"failover_timeout":"5",
"auto_recovery" : "true",
"auto_recovery_initial_wait" : "300",
"auto_recovery_max_backoff" : "86400",
"resource_type": "PolicyVtepHAHostSwitchProfile",
"display_name": "VtepHAProfile1"
}
```

and then be able to reference to it in the transport node profile

```
        "host_switches": [
            {
                "host_switch_name": "vtepha_vds",
                "host_switch_id": "50 22 ee c4 f6 40 79 8b-0e a4 2b da 6a 4c 36 b3",
                "host_switch_type": "VDS",
                "host_switch_mode": "ENS_INTERRUPT",
                "host_switch_profile_ids": [
                    {
                        "key": "UplinkHostSwitchProfile",
                        "value": "/infra/host-switch-profiles/b32e6ce6-f1ba-4e31-a2c4-33d550505cdd"
                    },
                    {
                        "key": "VtepHAHostSwitchProfile",
                        "value": "/infra/host-switch-profiles/vtephaprofile1"
                    }
                ],
```

### Describe alternatives you've considered

Ansible or manually get/put with postman

salv-orlando:
I think we can surely consider this. @ksamoray do you think it might be feasible?

Targeting this to milestone 3.6.0 as 3.5.0 is quite close.

PR #1122, opened 2024-02-19
Implement VTEP HA resource and data source

ksamoray:
Fixes: #1097

### DIFF
--- a/nsxt/resource_nsxt_policy_vtep_ha_host_switch_profile.go
+++ b/nsxt/resource_nsxt_policy_vtep_ha_host_switch_profile.go
@@ -1,0 +1,229 @@
+/* Copyright © 2024 Broadcom, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+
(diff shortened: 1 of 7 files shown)


nsxt/resource_nsxt_policy_vtep_ha_host_switch_profile.go
```diff
@@ -0,0 +1,241 @@
...
+	if err == nil {
+		converter := bindings.NewTypeConverter()
+		baseInterface, errs := converter.ConvertToGolang(structValue, model.PolicyBaseHostSwitchProfileBindingType())
+		if errs != nil {
+			return false, errs[0]
+		}
+		base := baseInterface.(model.PolicyBaseHostSwitchProfile)
+
+		resourceType := base.ResourceType
+		if resourceType != infra.HostSwitchProfiles_LIST_HOSTSWITCH_PROFILE_TYPE_POLICYVTEPHAHOSTSWITCHPROFILE {
```
annakhm, line 90: I don't think this check is needed, and its actually harmful, since two profiles with same id but different types will not be permitted by NSX. Exists function should return `true` if `id` is present, regardless of profile type.

annakhm (approved):

annakhm (approved):

PR #1124, opened 2024-02-20
Fix edge_transport_node import issue

ksamoray:
standard_host_switch attribute was not being populated during import/read.


salv-orlando (approved):
LGTM!

PR #1126, opened 2024-02-22
Mark site_path as Computed in project resource

annakhm:
site is auto-assigned by NSX if edge cluster is specified


ksamoray (approved):
LGTM

PR #1137, opened 2024-03-02
Add state argument to discovered node data source

annakhm:
This argument is not used in API call, it is added to ensure dependency on compute manager realization data source when deployed together with compute manager. Without this dependency discovered node data source might fail to resolve.

In addition, rename `nsxt_discover_node` to `nsxt_discovered_node`


ksamoray:
Mind renaming also the files funcs etc to discovered_nodes? It's better to remain consistent.

ksamoray (approved):

PR #1138, opened 2024-03-05
Fix incorrect state update following NSX error

annakhm:
When QoS profile resource is updated with incorrect value that NSX rejects, the value would still get updated in the state, even though Update returned an error.
This is terraform SDK bug:
https://github.com/hashicorp/terraform-plugin-sdk/issues/476

This change implements the suggested workaround.

We should consider to always implement this workaround as part of `handleUpdateError`, however first it would help to understand why this problem appears only for certain resources.


ksamoray:
@annakhm should we apply this on other resources with the update issue?

annakhm:
> @annakhm should we apply this on other resources with the update issue?

We can only apply this when Update function of the provider gets executed (it seems). When attribute validation fails the provider has no say and can't squeeze this in..

PR #1139, opened 2024-03-06
Fix empty QoS profile issue in segment

annakhm:
Sometimes NSX sends a segment binding map with empty segment for QoS profile. This change works around this NSX issue by ignoring binding map with empty profile path.


ksamoray (approved):
LGTM

PR #1148, opened 2024-03-13
Increase default retry timeouts and intervals

annakhm:
Fabric resources in general require bigger timeouts for the sake of async operation (f.e. several seconds might be required to clear edge node vm after edge delete operation). For now, we are increasing default timeout values for all resources in order to minimize errors. In future, we might want to introduce more granular approach to tumeouts.


ksamoray (approved):

annakhm:
We decided not to increase timeouts for all resources. We'll consider per-resource timeouts to focus on fabric.

PR #1150, opened 2024-03-14
Stall delete operation until deletion is complete

ksamoray:
When destroying a topology which has an edge nodes that were provisioned on a compute manager, there's a case where the compute manager is deleted before the edge VM is destroyed - that causes a failure.

This change waits for the deletion is complete, to satisfy the compute manager dependency.


annakhm (approved):

annakhm:
Lets wait with merging, not sure about returned code

PR #1160, opened 2024-03-24
Revise nsxt_management_cluster datasource

ksamoray:
Management cluster datasource is required to retrieve the managemnet endpoint sha256 thumbprint.
This change un-deprecates this DS as it has no replacement and changes the API in use to the NSX golang SDK.

### DIFF
--- a/nsxt/data_source_nsxt_management_cluster.go
+++ b/nsxt/data_source_nsxt_management_cluster.go
@@ -5,15 +5,15 @@ package nsxt
 
 import (
 	"fmt"
-	"net/http"
+	"net"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx"
 )
 
 func dataSourceNsxtManagementCluster() *schema.Resource {
 	return &schema.Resource{
-		Read:               dataSourceNsxtManagementClusterRead,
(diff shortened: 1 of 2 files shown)


nsxt/data_source_nsxt_management_cluster.go
```diff
@@ -30,42 +30,31 @@ func dataSourceNsxtManagementCluster() *schema.Resource {
...
 func dataSourceNsxtManagementClusterRead(d *schema.ResourceData, m interface{}) error {
-	nsxClient := m.(nsxtClients).NsxtClient
-	if nsxClient == nil {
-		return dataSourceNotSupportedError()
-	}
+	connector := getPolicyConnector(m)
+	client := nsx.NewClusterClient(connector)
 
-	clusterObj, resp, err := nsxClient.NsxComponentAdministrationApi.ReadClusterConfig(nsxClient.Context)
+	hostIP, err := net.LookupIP(m.(nsxtClients).Host[len("https://"):])
```
annakhm, line 36: Shouldn't we skip this if host is already IP address? Can we reuse `resolveHostIPs` function here?

ksamoray: net.LookupIP does that exactly[already](https://github.com/golang/go/blob/master/src/net/lookup.go#L309), no reason to double check IMO.

annakhm (approved):

Issue #1146, opened 2024-03-12
nsx_edge_transport_node fails to create edge with two standard_host_switch entries


martinrohrbach:
### Describe the bug

We've eagerly awaited the release of the new provider to test nsx_edge_transport_node to deploy new edges. The challenge for us is that we use two switches per edge and that did not work with the nsx_transport_node resource. 

This is what we tested with:

```
resource "nsxt_edge_transport_node" "test2” {
  description  = "edge-test"
  display_name = "edge-test"

  standard_host_switch {
    host_switch_id = "xxx"
    ip_assignment {
      static_ip_pool = nsxt_policy_ip_pool.ip-pool-for-test-edge.realized_id
    }
    transport_zone_endpoint {
      transport_zone          = "xxx“
      transport_zone_profiles = ["xxx"]
    }
    host_switch_profile = ["xxx"]
    pnic {
      device_name = "fp-eth0"
      uplink_name = "uplink"
    }
  }

  standard_host_switch {
    host_switch_id = "yyy"
    ip_assignment {
      assigned_by_dhcp = true
    }
    transport_zone_endpoint {
      transport_zone          = "yyy“
      transport_zone_profiles = ["yyy“]
    }
    host_switch_profile = ["yyy"]
    pnic {
      device_name = "fp-eth0"
      uplink_name = "uplink"
    }
  }

  deployment_config {
    form_factor = "SMALL"
    node_user_settings {
      cli_password  = "VMware1!VMware1!"
      root_password = "VMware1!VMware1!"
    }
    vm_deployment_config {
        <redacted>
    }
  }
 
  node_settings {
    hostname             = "edge-test"
    allow_ssh_root_login = true
    enable_ssh           = true
  }
}
```

Note that those two switches have completey different configs, i.e. they are on a different transport zone, have different host_switch_ids and everything. 

Deploying the test edge with just either one of the standard_host_switch configs works find and we end up with a new edge and a switch called "nsxDefaultHostSwitch".

However, if we use both standard_host_switch at the same time, we get the following API error:

```
Failed to create Transport node edge-test: Multiple HostSwitches were found with the same HostSwitch nsxDefaultHostSwitch in TransportNode.
```

If we do this through the API or in the GUI, we can specify a distinct name for each switch, obviously some default is used here which leads to the clash. 

Can you expose a property to actually set a distinct name for each switch (while keeping the default as it is)?

### Reproduction steps

[...]

PR #1163, opened 2024-03-25
Add host_switch_name attribute to transport node

ksamoray:
NSX requires host_switch_name specification even though this attribute is marked as deprecated.
Without it, NSX issues an error when using multiple host switches. As this is an issue even if NSX addresses this later, it's better to add this attribute to the provider resources.

Fixes: #1146


annakhm:
Looks like the assignment of the new attribute is missing?

annakhm (approved):

PR #1173, opened 2024-04-04
Wait for subnet realization in ip_allocation test

ksamoray:
When an IP address is consument from an IP pool, wait for the subnet to be realized before consuming IPs from the pool.


salv-orlando (approved):
LGTM!

PR #1178, opened 2024-04-09
Switch sub-config with no IP assignment bug

ksamoray:
When a switch sub-config is created with no specification of IP assignment, the code mishandles it and tries set it anyway.


salv-orlando (approved):
LGTM

PR #1181, opened 2024-04-11
Change input of TNC realization from id to path

annakhm:
Since corresponding resource is policy resource, it makes more sense to use path and remove the need for site_path parameter, which is part of the path.

In addition, make sure the error returned is never empty.

### DIFF
--- a/nsxt/data_source_nsxt_policy_host_transport_node_collection_realization.go
+++ b/nsxt/data_source_nsxt_policy_host_transport_node_collection_realization.go
@@ -11,7 +11,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
-	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/sites/enforcement_points"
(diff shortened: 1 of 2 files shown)


salv-orlando (approved):
LGTM

nsxt/data_source_nsxt_policy_host_transport_node_collection_realization.go
```diff
@@ -72,31 +68,25 @@ func dataSourceNsxtPolicyHostTransportNodeCollectionRealizationRead(d *schema.Re
...
+
 	stateConf := &resource.StateChangeConf{
 		Pending: pendingStates,
 		Target:  targetStates,
 		Refresh: func() (interface{}, string, error) {
-			state, err := client.Get(objSitePath, getPolicyEnforcementPoint(m), objID)
+			state, err := client.Get(site, ep, objID)
 			if err != nil {
-				return state, model.TransportNodeCollectionState_STATE_IN_PROGRESS, logAPIError("Error while waiting for realization of Transport Node Collection", err)
+				return state, model.TransportNodeCollectionState_STATE_IN_PROGRESS, logAPIError("Error getting collection state", err)
```
qiyueyao, line 89: Perhaps detail to "Error getting Transport Node Collection state"?

qiyueyao (commented):
LGTM, thanks!

nsxt/data_source_nsxt_policy_host_transport_node_collection_realization.go
qiyueyao: Page did not refresh, if merged then nvm~

PR #1184, opened 2024-04-15
Support revision in flood protection binding

annakhm:
When revision is specified, there is no need to recreate the object manually.

In addition, remove parallelism in test because of sequence number conflict.


ksamoray (approved):
LGTM

PR #1185, opened 2024-04-16
Remove hard-coded maxint default in flood profile

annakhm:
It causes compilation error on some platforms. Set default value to Computed instead.


ksamoray (approved):
LGTM

PR #1191, opened 2024-04-18
Add custom groups support to upgrade run resource

ksamoray:
Add the logic to create and update custom host groups while running the upgrade process.

### DIFF
--- a/nsxt/resource_nsxt_upgrade_run.go
+++ b/nsxt/resource_nsxt_upgrade_run.go
@@ -15,6 +15,7 @@ import (
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/fabric"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/model"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/upgrade"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/upgrade/plan"
@@ -519,6 +520,13 @@ func updateUpgradeUnitGroups(upgradeClientSet *upgradeClientSet, d *schema.Resou
 		}
 	}
 
+	groupList, err := upgradeClientSet.GroupClient.List(nil, nil, nil, nil, nil, nil, nil, nil)
+	if err != nil {
+		return err
+	}
+	groupIDMap := make(map[string]bool)
+	groupNameMap := make(map[string]bool)
+
 	preUpgradeGroupID := ""
 	for _, groupI := range d.Get(componentToGroupKey[component]).([]interface{}) {
 		group := groupI.(map[string]interface{})
@@ -531,11 +539,8 @@ func updateUpgradeUnitGroups(upgradeClientSet *upgradeClientSet, d *schema.Resou
 			if groupName == "" {
 				return fmt.Errorf("couldn't find upgrade unit group without id or display_name")
 			}
+			groupNameMap[groupName] = true
 			// This is a custom group, try to find it by name
-			groupList, err := upgradeClientSet.GroupClient.List(nil, nil, nil, nil, nil, nil, nil, nil)
-			if err != nil {
-				return err
-			}
 			for i, group := range groupList.Results {
 				if *group.DisplayName == groupName {
 					if groupGet == nil {
@@ -568,7 +573,7 @@ func updateUpgradeUnitGroups(upgradeClientSet *upgradeClientSet, d *schema.Resou
 				if group["hosts"] != nil {
 					schemaUnits = interface2StringList(group["hosts"].([]interface{}))
 				}
-				getHostDefaultUpgradeGroup, err := getHostDefaultUpgradeGroupGetter(m)
+				getHostDefaultUpgradeGroup, err := getHostDefaultUpgradeGroupGetter(m, groupList)
 				if err != nil {
 					return fmt.Errorf("failed to retrieve host upgrade groups, error is %v", err)
 				}
@@ -595,6 +600,7 @@ func updateUpgradeUnitGroups(upgradeClientSet *upgradeClientSet, d *schema.Resou
 				groupGet.UpgradeUnits = groupMembers
 			}
 		} else {
+			groupIDMap[groupID] = true
 			group, err := upgradeClientSet.GroupClient.Get(groupID, nil)
 			if err != nil {
 				return err
@@ -661,7 +667,35 @@ func updateUpgradeUnitGroups(upgradeClientSet *upgradeClientSet, d *schema.Resou
 		}
 		preUpgradeGroupID = groupID
 	}
+
+	// After group update is complete, check for empty custom host groups which aren't defined in the schema, and have no members - these can be deleted.
+	componentType := "HOST"
+	groupList, err = upgradeClientSet.GroupClient.List(&componentType, nil, nil, nil, nil, nil, nil, nil)
+	if err != nil {
+		return err
+	}
+	for _, group := range groupList.Results {
+		// Non-empty groups cannot be deleted anyway so skip
+		if *group.UpgradeUnitCount == 0 {
+			if !groupIDMap[*group.Id] && !groupNameMap[*group.DisplayName] && !isPredefinedGroup(m, group) {
+				err = upgradeClientSet.GroupClient.Delete(*group.Id)
+				if err != nil {
+					return err
+				}
+			}
+		}
+	}
 	return nil
+}
+
+func isPredefinedGroup(m interface{}, group model.UpgradeUnitGroup) bool {
+	if group.DisplayName != nil && *group.DisplayName == hostUpgradeUnitDefaultGroup && *group.Type_ == "HOST" {
+		return true
+	}
+	connector := getPolicyConnector(m)
+	client := fabric.NewComputeCollectionsClient(connector)
+	_, err := client.Get(*group.Id)
+	return !isNotFoundError(err)
 }
 
 func updateComponentUpgradePlanSetting(settingClient plan.SettingsClient, d *schema.ResourceData, component string) error {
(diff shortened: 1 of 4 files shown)


nsxt/resource_nsxt_host_upgrade_group.go
```diff
@@ -0,0 +1,183 @@
...
+			"description":  getDescriptionSchema(),
+			"display_name": getDisplayNameSchema(),
+			"tag":          getTagsSchema(),
+			"enabled": {
+				Type:        schema.TypeBool,
+				Description: "Flag to indicate whether upgrade of this group is enabled or not",
+				Optional:    true,
+				Default:     true,
+			},
+			"extended_configuration": {
```
annakhm, line 36: I think we can reuse schema for key/value pairs, we already have a func `getSwitchingProfileIdsSchema`, we might want to rename it and generalize the description. Also `advanced_configuration` of edge transport node might benefit from this

nsxt/resource_nsxt_host_upgrade_group_assignment.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil {
+		return err
+	}
+
+	hostIDs := getUnitIDsFromUnits(group.UpgradeUnits)
+	if slices.Contains(hostIDs, hostID) {
+		return fmt.Errorf("host %s already exists in group %s", hostID, groupID)
+	}
+	group.UpgradeUnits = append(group.UpgradeUnits, model.UpgradeUnit{Id: &hostID})
+	_, err = client.Update(groupID, group)
```
annakhm, line 69: I feel simultaneous updates might cause errors here, so we might need to wrap this code (including `Get`) in retry. Auto-retry that we have today will not re-read the revision, so Update would fail again if revision got updated. What do you think?

nsxt/resource_nsxt_host_upgrade_group.go
ksamoray: Switching profile uses a different schema than edge transport node and upgrade - [SwitchingProfileTypeIdEntry](https://vdc-download.vmware.com/vmwb-repository/dcr-public/8e2a4b1a-659f-4be6-a4f4-7ca42b04cb84/9a2c1372-7174-40f0-89d2-27615bfd5d08/api_includes/types_SwitchingProfileTypeIdEntry.html) vs [KeyValuePair](https://vdc-download.vmware.com/vmwb-repository/dcr-public/8e2a4b1a-659f-4be6-a4f4-7ca42b04cb84/9a2c1372-7174-40f0-89d2-27615bfd5d08/api_includes/types_KeyValuePair.html).
I'll consolidate edge and upgrade and leave switching profiles as is.

nsxt/resource_nsxt_host_upgrade_group_assignment.go
ksamoray: Makes a lot of sense. Do we have any example for such retry mechanism in the code?
BTW it seems like the `_revision` attribute is ignored by NSX - I'll open a bug for this.

ksamoray: I think that I've found something.

nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -503,9 +523,76 @@ func updateUpgradeUnitGroups(upgradeClientSet *upgradeClientSet, d *schema.Resou
...
+		var groupGet *model.UpgradeUnitGroup
+		var err error
+		isCreate := false
+		if groupID == "" {
+			groupName := group["display_name"].(string)
+			if groupName == "" {
+				return fmt.Errorf("couldn't find upgrade unit group without id or display_name")
+			}
+			// This is a custom group, try to find it by name
+			groupList, err := upgradeClientSet.GroupClient.List(nil, nil, nil, nil, nil, nil, nil, nil)
```
annakhm, line 535: is pagination relevant here?

nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -503,9 +523,76 @@ func updateUpgradeUnitGroups(upgradeClientSet *upgradeClientSet, d *schema.Resou
...
+				}
+				for _, nsxUnit := range nsxUnits {
+					if !slices.Contains(schemaUnits, nsxUnit) {
+						groupID, err := getHostDefaultUpgradeGroup(m, nsxUnit)
+						if isNotFoundError(err) {
+							return fmt.Errorf("couldn't find default group for host %s as default group was not found", nsxUnit)
+						} else if err != nil {
+							return handleDeleteError("Host Upgrade Group Binding", nsxUnit, err)
+						}
+						return addHostToGroup(m, groupID, nsxUnit, false)
```
annakhm, line 579: Is the `return` here intentional?

nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -722,3 +812,74 @@ func resourceNsxtUpgradeRunUpdate(d *schema.ResourceData, m interface{}) error {
...
+
+	doUpdate := func() error {
+		group, err := client.Get(groupID, nil)
+		if err != nil {
+			return err
+		}
+
+		hostIDs := getUnitIDsFromUnits(group.UpgradeUnits)
+		if slices.Contains(hostIDs, hostID) {
+			return fmt.Errorf("host %s already exists in group %s", hostID, groupID)
```
annakhm, line 865: should we swallow this error?

nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -722,3 +812,74 @@ func resourceNsxtUpgradeRunUpdate(d *schema.ResourceData, m interface{}) error {
...
+
+	if node.ComputeCollectionId != nil {
+		return *node.ComputeCollectionId, nil
+	}
+
+	// This host is not a part of a compute cluster:
+	// it should be assigned to the 'Group 1 for ESXI' group (this value is hardcoded in NSX)
+	groupClient := upgrade.NewUpgradeUnitGroupsClient(connector)
+	componentType := "HOST"
+	hostGroups, err := groupClient.List(&componentType, nil, nil, nil, nil, nil, nil, nil)
```
annakhm, line 838: It seems wasteful to run this API for each `getHostDefaultUpgradeGroup` call, can the logic be change to run it once? We only care for default groups here, and default groups are not expected to change throughout the update operation, is that correct?

annakhm:
I didn't see any update in `setUpgradeRunOutput` function (sorry if I missed it), but looks like custom group needs to be updated with its new attributes

nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -503,9 +523,76 @@ func updateUpgradeUnitGroups(upgradeClientSet *upgradeClientSet, d *schema.Resou
...
 		group := groupI.(map[string]interface{})
 		groupID := group["id"].(string)
-		groupGet, err := upgradeClientSet.GroupClient.Get(groupID, nil)
-		if err != nil {
-			return err
+		var groupGet *model.UpgradeUnitGroup
+		var err error
+		isCreate := false
+		if groupID == "" {
+			groupName := group["display_name"].(string)
```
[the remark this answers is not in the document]
ksamoray: No - this isn't doable as the display_name is used to identify a pre-existing custom group, and as the upgrade support doesn't maintain/support state.
If the upgrade implementation would support state, we could identify the group by id and update the group.

nsxt/resource_nsxt_upgrade_run.go
ksamoray: Do we expect more than 1000 groups?

nsxt/resource_nsxt_upgrade_run.go
ksamoray: It's a bug...

nsxt/resource_nsxt_upgrade_run.go
ksamoray: You mean, load the entire host/group map and cache it for the entire execution?
As the "original" upgrade group for each host could be different according to cluster membership (or lack of cluster association).

annakhm: Yes, only run the API once and use the list for all calls of this function

ksamoray:
> I didn't see any update in `setUpgradeRunOutput` function (sorry if I missed it), but looks like custom group needs to be updated with its new attributes

Yeah I missed this. Although as we do not support the state existence, that's a bit meaningless for now.

ksamoray:
> I didn't see any update in `setUpgradeRunOutput` function (sorry if I missed it), but looks like custom group needs to be updated with its new attributes

But this is only the status [type](https://dp-downloads.broadcom.com/api-content/apis/API_NTDCRA_001/4.1.2/html/api_includes/types_UpgradeUnitStatus.html) - which doesn't have the extra group  attributes. So I'm not sure that a change is required here.

annakhm:
> > I didn't see any update in `setUpgradeRunOutput` function (sorry if I missed it), but looks like custom group needs to be updated with its new attributes
> 
> Yeah I missed this. Although as we do not support the state existence, that's a bit meaningless for now.

I think that although we ask to remove state post-upgrade, `terraform plan` should not produce non-empty diff after upgrade. is this true today?

nsxt/resource_nsxt_upgrade_run.go
annakhm: Doesn't sound feasible, but maybe a comment about it would be helpful

annakhm:
LGTM, except for one doubt I still have: after upgrade is complete and there are no new changes in config, would `terraform plan` show non-empty diff?

nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -503,9 +523,83 @@ func updateUpgradeUnitGroups(upgradeClientSet *upgradeClientSet, d *schema.Resou
 	for _, groupI := range d.Get(componentToGroupKey[component]).([]interface{}) {
```
annakhm, line 523: Looks like the code doesn't delete custom groups that were created with previous apply (which might have failed) but are not present in current config, is that correct?
To solve this, `d.GetChange("host_group")` might be useful

ksamoray:
Terraform doesn't show an empty diff, also without these changes as far as I understand (there are dependencies on computed groups within the upgrade process - I think that this is the reason?).
As for group deletion: it's possible to check the groups for empty groups (with no upgrade units) which do not match the "predefined group pattern" (either have the id of a compute collection, or that hardcoded name which NSX uses for non-clustered hosts). These can be deleted I guess. Is that what you had in mind?

annakhm:
> Terraform doesn't show an empty diff, also without these changes as far as I understand (there are dependencies on computed groups within the upgrade process - I think that this is the reason?). As for group deletion: it's possible to check the groups for empty groups (with no upgrade units) which do not match the "predefined group pattern" (either have the id of a compute collection, or that hardcoded name which NSX uses for non-clustered hosts). These can be deleted I guess. Is that what you had in mind?

I would suggest to rely on `GetChange` routing to check if groups are explicitly deleted + deleting groups that we know are recreated as a result of update.

annakhm (approved):

PR #1198, opened 2024-04-26
Consistent naming in TN data source

annakhm:


ksamoray:
This DS retrieves all kinds of transport nodes - so if we rename it, we should at least filter out non-edge TNs.

ksamoray:
If we choose to rename, can we also rename the realization DS `nsxt_transport_node_realization`?

annakhm:
> This DS retrieves all kinds of transport nodes - so if we rename it, we should at least filter out non-edge TNs.

This PR changes the TN to be generic, by removing `edge` from naming

ksamoray:
> > This DS retrieves all kinds of transport nodes - so if we rename it, we should at least filter out non-edge TNs.
> 
> This PR changes the TN to be generic, by removing `edge` from naming

Got it... Seems like "someone" made a mess here :)

ksamoray (approved):
Thanks!

PR #1201, opened 2024-05-01
Manager info data source

ksamoray:
A container for various information about the NSX manager.
For now, retrieves the software version of the manager.


ksamoray:
@annakhm maybe it's better to consolidate into `nsxt_management_cluster`?

annakhm:
I don't think we should consolidate it with `nsxt_management_cluster`. But I'm thinking there might be other NSX parameters worth exposing, in which case we should consider generic name for the data source (`nsxt_manager_info`? `nsxt_about`? )

annakhm (approved):

PR #1205, opened 2024-05-07
Add reflect lib for auto generation

wsquan171:
This PR is a continuation of previous work by @annakhm on test-reflect branch on the following aspects.

### Functional
- Support conversion of primitive schema types: int, bool, string
- Support conversion of nested schema types: list, set
- Nested struct in sdk model should be represented as list with at most 1 element, with metadata `SchemaType` set to `struct`
- List or set of primitive data types (int, bool, string) should use `*ExtendedSchema` as `Schema.Elem`
- List or set of struct data types should use `*ExtendedResource` as `Schema.Elem`, and set `Metadata.ReflectType` to the type of element type in the slice, rather than a slice type.
- UT under `metadata/metaddata_test.go` can be referenced as example use cases.

### Non-Functional
- Refactored reflect code into own package for adding isolated UT
- Moved version check logic to `utils` package, so both `nsxt` and `metadata` package can share the code
- Added Github action for UT

### TBD in a future PR
- Support of polymorphous struct or list of structs
- Support of Schema.TypeMap

### DIFF
--- a/nsxt/metadata/metadata.go
+++ b/nsxt/metadata/metadata.go
@@ -147,10 +147,16 @@ func StructToSchema(elem reflect.Value, d *schema.ResourceData, metadata map[str
 // currently supports nested subtype and trivial types
 func SchemaToStruct(elem reflect.Value, d *schema.ResourceData, metadata map[string]*ExtendedSchema, parent string, parentMap map[string]interface{}) {
 	for key, item := range metadata {
-		if item.Metadata.ReadOnly || item.Metadata.Skip {
+		if item.Metadata.ReadOnly {
+			log.Printf("[INFO] skip key %s as read only", key)
+			continue
+		}
+		if item.Metadata.Skip {
+			log.Printf("[INFO] skip key %s", key)
 			continue
 		}
 		if item.Metadata.IntroducedInVersion != "" && util.NsxVersionLower(item.Metadata.IntroducedInVersion) {
+			log.Printf("[INFO] skip key %s as NSX does not have support", key)
 			continue
 		}
 
@@ -208,11 +214,20 @@ func SchemaToStruct(elem reflect.Value, d *schema.ResourceData, metadata map[str
 		}
 		if item.Metadata.SchemaType == "list" || item.Metadata.SchemaType == "set" {
 			var itemList []interface{}
-			if len(parent) > 0 {
-				itemList = parentMap[key].([]interface{})
+			if item.Metadata.SchemaType == "list" {
+				if len(parent) > 0 {
+					itemList = parentMap[key].([]interface{})
+				} else {
+					itemList = d.Get(key).([]interface{})
+				}
 			} else {
-				itemList = d.Get(key).([]interface{})
+				if len(parent) > 0 {
+					itemList = parentMap[key].(*schema.Set).List()
+				} else {
+					itemList = d.Get(key).(*schema.Set).List()
+				}
 			}
+
 			if len(itemList) == 0 {
 				continue
 			}
@@ -233,7 +248,11 @@ func SchemaToStruct(elem reflect.Value, d *schema.ResourceData, metadata map[str
 				}
 
 				for i, v := range itemList {
-					sliceElem.Index(i).Set(reflect.ValueOf(v))
+					if childElem.Metadata.SchemaType == "int" {
+						sliceElem.Index(i).Set(reflect.ValueOf(v).Convert(reflect.TypeOf(int64(0))))
+					} else {
+						sliceElem.Index(i).Set(reflect.ValueOf(v))
+					}
 					log.Printf("[INFO] appending %v to %s", v, key)
 				}
 			}
--- a/nsxt/metadata/metadata.go
+++ b/nsxt/metadata/metadata.go
@@ -130,6 +130,13 @@ func SchemaToStruct(elem reflect.Value, d *schema.ResourceData, metadata map[str
 		if item.Metadata.IntroducedInVersion != "" && util.NsxVersionLower(item.Metadata.IntroducedInVersion) {
 			continue
 		}
+		if len(parent) == 0 {
+			_, ok := d.GetOk(key)
+			if !ok {
+				log.Printf("[INFO] Skipping key %s with unset value", key)
+				continue
+			}
+		}
 
 		log.Printf("[INFO] inspecting key %s with type %s", key, item.Metadata.SchemaType)
 		if len(parent) > 0 {
@@ -178,7 +185,6 @@ func SchemaToStruct(elem reflect.Value, d *schema.ResourceData, metadata map[str
 			nestedSchema := nestedSchemaList[0].(map[string]interface{})
 
 			childElem := item.Schema.Elem.(*ExtendedResource)
-			log.Printf("[INFO] calling recur %s", key)
 			SchemaToStruct(nestedObj.Elem(), d, childElem.Schema, key, nestedSchema)
 			log.Printf("[INFO] assigning struct %v to %s", nestedObj, key)
 			elem.FieldByName(item.Metadata.SdkFieldName).Set(nestedObj)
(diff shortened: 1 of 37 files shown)


nsxt/resource_nsxt_policy_mac_discovery_profile.go
```diff
@@ -14,13 +15,116 @@ import (
...
+	},
+	"unknown_unicast_flooding_enabled": {
+		Schema: schema.Schema{
+			Type:     schema.TypeBool,
+			Optional: true,
+		},
+		Metadata: metadata.Metadata{
+			SchemaType:          "bool",
+			SdkFieldName:        "UnknownUnicastFloodingEnabled",
+			IntroducedInVersion: "4.0.0",
```
annakhm, line 119: I don't think this attribute was introduced in 4.0.0 (if this is copied from PoC, it was just a test for `IntroducedInVersion` functionality)

nsxt/metadata/metadata.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+// SchemaToStruct converts terraform schema to NSX model struct
+// currently supports nested subtype and trivial types
+func SchemaToStruct(elem reflect.Value, d *schema.ResourceData, metadata map[string]*ExtendedSchema, parent string, parentMap map[string]interface{}) {
+	for key, item := range metadata {
+		if item.Metadata.ReadOnly || item.Metadata.Skip {
+			continue
+		}
+		if item.Metadata.IntroducedInVersion != "" && util.NsxVersionLower(item.Metadata.IntroducedInVersion) {
```
annakhm, line 153: I think we need debug logging here and in the above conditional

nsxt/metadata/metadata.go
```diff
@@ -0,0 +1,256 @@
...
+
+			childElem := item.Schema.Elem.(*ExtendedResource)
+			SchemaToStruct(nestedObj.Elem(), d, childElem.Schema, key, nestedSchema)
+			log.Printf("[INFO] assigning struct %v to %s", nestedObj, key)
+			elem.FieldByName(item.Metadata.SdkFieldName).Set(nestedObj)
+		}
+		if item.Metadata.SchemaType == "list" || item.Metadata.SchemaType == "set" {
+			var itemList []interface{}
+			if len(parent) > 0 {
+				itemList = parentMap[key].([]interface{})
```
annakhm, line 212: I don't think this conversion can work with `TypeSet` (f.e., to convert `TypeSet` to string list, we use `d.Get(schemaAttrName).(*schema.Set).List()`
Is there an example of `TypeSet` within this PR?

nsxt/resource_nsxt_policy_mac_discovery_profile.go
```diff
@@ -14,13 +15,116 @@ import (
...
+				UpdateValue: "50",
+			},
+		},
+	},
+	"mac_limit_policy": {
+		Schema: schema.Schema{
+			Type:         schema.TypeString,
+			ValidateFunc: validation.StringInSlice(macDiscoveryProfileMacLimitPolicyValues, false),
+			Optional:     true,
+			Default:      "ALLOW",
```
annakhm, line 84: I suggest we use the constant here and below

nsxt/resource_nsxt_policy_mac_discovery_profile.go
```diff
@@ -98,28 +163,20 @@ func resourceNsxtPolicyMacDiscoveryProfileCreate(d *schema.ResourceData, m inter
 		return err
 	}
 
+	// TODO - consider including standard object attributes in the schema
+	tags := getPolicyTagsFromSchema(d)
```
annakhm, line 167: Do we want to include `tags`, `display_name` and `description` in the extended schema?

annakhm:
Lets make sure we cover the following use cases, either in unit tests or with a resource:

- list of native types (int, string, bool)
- set of native types
- list of structs
- set of structs
- pointer to struct in the model (list of with maxItems = 1)
- double-nested struct or list of structs
- list of polymorphic structs (next PR)
- set of polymorphic structs (next PR)
- pointer to polymorphic struct (next PR)

nsxt/resource_nsxt_policy_mac_discovery_profile.go
wsquan171: Thanks, fixed.

nsxt/metadata/metadata.go
wsquan171: Fixed and added in UT

nsxt/metadata/metadata.go
wsquan171: added

nsxt/resource_nsxt_policy_mac_discovery_profile.go
wsquan171: done

nsxt/resource_nsxt_policy_mac_discovery_profile.go
wsquan171: It's not hard to improve `GetExtendedSchema` and make it return common extended schema for some well-known policy attrs, but it'll need some refactor on the test side of stuff to make everything e2e working. Can we tackle this after the polymorphic struct is addressed?

nsxt/metadata/metadata.go
```diff
@@ -0,0 +1,275 @@
...
+						reflect.MakeSlice(reflect.TypeOf([]int64{}), len(itemList), len(itemList)))
+				}
+
+				for i, v := range itemList {
+					if childElem.Metadata.SchemaType == "int" {
+						sliceElem.Index(i).Set(reflect.ValueOf(v).Convert(reflect.TypeOf(int64(0))))
+					} else {
+						sliceElem.Index(i).Set(reflect.ValueOf(v))
+					}
+					log.Printf("[INFO] appending %v to %s", v, key)
```
annakhm, line 256: I think those printouts should be `TRACE` level, and we might want to preface them with some keyword so that the context of conversion is clear

nsxt/metadata/metadata.go
```diff
@@ -0,0 +1,275 @@
...
+			} else {
+				if len(parent) > 0 {
+					itemList = parentMap[key].(*schema.Set).List()
+				} else {
+					itemList = d.Get(key).(*schema.Set).List()
+				}
+			}
+
+			if len(itemList) == 0 {
+				continue
```
annakhm, line 232: I wonder if here we might run into a problem of assigning `nil` value to array instead of assigning empty array?
We need to be able to remove a list of values from a struct, and in some cases `nil` value would be ignored by the Patch call. 
We can leave this as is for now, and consider adding another metadata value that would control this behavior when the need arises.

nsxt/metadata/metadata.go
```diff
@@ -0,0 +1,275 @@
...
+	for key, item := range metadata {
+		if item.Metadata.Skip {
+			continue
+		}
+
+		log.Printf("[INFO] inspecting key %s", key)
+		if len(parent) > 0 {
+			log.Printf("[INFO] parent %s key %s", parent, key)
+		}
+		if elem.FieldByName(item.Metadata.SdkFieldName).IsNil() {
```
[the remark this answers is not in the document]
annakhm: Agree with Salvatore, we should ease troubleshooting for bad metadata arguments that will inevitably occur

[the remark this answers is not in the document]
wsquan171: added check for `elem.FieldByName(item.Metadata.SdkFieldName).IsValid()` and returns error if the field is not found

nsxt/metadata/metadata.go
```diff
@@ -0,0 +1,275 @@
...
+			continue
+		}
+		if item.Metadata.IntroducedInVersion != "" && util.NsxVersionLower(item.Metadata.IntroducedInVersion) {
+			log.Printf("[INFO] skip key %s as NSX does not have support", key)
+			continue
+		}
+
+		log.Printf("[INFO] inspecting key %s with type %s", key, item.Metadata.SchemaType)
+		if len(parent) > 0 {
+			log.Printf("[INFO] parent %s key %s", parent, key)
```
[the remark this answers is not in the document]
wsquan171: added the info of parent as part of context prefix as Anna suggested.

nsxt/metadata/metadata.go
wsquan171: Nice catch. I actually think there is no harm in passing an empty slice if tf schema is empty. For create this is a no op, and for update this is an explict erasure. We can revisit this part if other handling for erasing is needed for some NSX resources.

nsxt/metadata/metadata.go
wsquan171: moved most to TRACE, except for the ones logged before returning an error.

annakhm (approved):

PR #1206, opened 2024-05-10
Update syslog config spec in edge TN resource

annakhm:
Port should follow the API spec and be defined as string rather than int. Spec allows port range, however NSX rejects multiple port config, so we'll leave the single port validation for now, while allowing to change it later without deprecating the attribute. In addition, this change removes name attribute for syslog since NSX seems to ignore it.


ksamoray (approved):

salv-orlando (approved):
LGTM

PR #1207, opened 2024-05-13
Add VPC data source

ksamoray:


nsxt/data_source_policy_vpc_test.go
```diff
@@ -0,0 +1,128 @@
...
+					if err := testAccDataSourceNsxtPolicyVPCCreate(name); err != nil {
+						t.Error(err)
+					}
+				},
+				Config: testAccNsxtPolicyVPCReadTemplate(name),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(testResourceName, "display_name", name),
+					resource.TestCheckResourceAttr(testResourceName, "description", name),
+					resource.TestCheckResourceAttrSet(testResourceName, "path"),
+					//resource.TestCheckResourceAttrSet(testResourceName, "default_gateway_path"),
```
salv-orlando, line 43: Why is this attribute commented? Is it giving you troubles?

nsxt/data_source_nsxt_policy_vpc.go
```diff
@@ -0,0 +1,102 @@
...
+		objGet, err := client.Get(defaultOrgID, ctx.ProjectID, objID)
+		if err != nil {
+			return handleDataSourceReadError(d, "VPC", objID, err)
+		}
+		obj = objGet
+	} else if objName == "" {
+		return fmt.Errorf("error obtaining VPC ID or name during read")
+	} else {
+		// Get by full name/prefix
+		objList, err := client.List(defaultOrgID, ctx.ProjectID, nil, nil, nil, nil, nil, nil)
```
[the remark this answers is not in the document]
ksamoray: I've tried to implement the project datasource with search API at the time, which didn't work. So I didn't even try with this one - maybe it's worthwhile trying.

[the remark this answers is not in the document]
ksamoray: client.List() call (L55) takes a project parameter, so uniqueness is only checked at the scope of these query results.

nsxt/data_source_nsxt_policy_vpc.go
```diff
@@ -0,0 +1,102 @@
...
+		for _, objInList := range objList.Results {
+			if strings.HasPrefix(*objInList.DisplayName, objName) {
+				prefixMatch = append(prefixMatch, objInList)
+			}
+			if *objInList.DisplayName == objName {
+				perfectMatch = append(perfectMatch, objInList)
+			}
+		}
+		if len(perfectMatch) > 0 {
+			if len(perfectMatch) > 1 {
```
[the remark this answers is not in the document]
ksamoray: client.List() call (L55) takes a project parameter, so uniqueness is only checked at the scope of these query results.

nsxt/data_source_policy_vpc_test.go
ksamoray: I didn't set the DGW path parameter in Create() (L84), I intended to remove this as although I could gather the T0 id, I had to create a path from it as I do with IP block. It's quite ugly... If you think that there's some importance to this, I can add it.
TBH I'd rather rewrite this once we have a resource for VPC.

nsxt/data_source_nsxt_policy_vpc.go
```diff
@@ -0,0 +1,102 @@
...
+
+func dataSourceNsxtPolicyVPC() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNsxtPolicyVPCRead,
+		Schema: map[string]*schema.Schema{
+			"id":                   getDataSourceIDSchema(),
+			"display_name":         getDataSourceDisplayNameSchema(),
+			"description":          getDataSourceDescriptionSchema(),
+			"path":                 getPathSchema(),
+			"context":              getContextSchema(),
```
annakhm, line 23: Should project be required here?

nsxt/data_source_nsxt_policy_vpc.go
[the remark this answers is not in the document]
annakhm: Should this support pagination?

nsxt/data_source_nsxt_policy_vpc.go
ksamoray: Indeed

nsxt/data_source_nsxt_policy_vpc.go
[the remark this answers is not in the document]
ksamoray: Replaced logic with search API

website/docs/d/policy_vpc.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+
+In addition to arguments listed above, the following attributes are exported:
+
+* `description` - The description of the resource.
+* `path` - The NSX path of the policy resource.
+* `short_id` - Defaults to id if id is less than equal to 8 characters or defaults to random generated id if not set.
+* `site_info` - Information related to sites applicable for given Project.
+    * `edge_cluster_paths` - The edge cluster on which the networking elements for the Org will be created.
+    * `site_path` - This represents the path of the site which is managed by Global Manager. For the local manager, if set, this needs to point to 'default'.
+* `default_gateway_path` - The tier 0 has to be pre-created before Project is created.
```
annakhm, line 45: This description does not seem relevant to data source

website/docs/d/policy_vpc.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+
+## Attributes Reference
+
+In addition to arguments listed above, the following attributes are exported:
+
+* `description` - The description of the resource.
+* `path` - The NSX path of the policy resource.
+* `short_id` - Defaults to id if id is less than equal to 8 characters or defaults to random generated id if not set.
+* `site_info` - Information related to sites applicable for given Project.
+    * `edge_cluster_paths` - The edge cluster on which the networking elements for the Org will be created.
```
annakhm, line 43: `Org` => `VPC`?

website/docs/d/policy_vpc.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+    * `site_path` - This represents the path of the site which is managed by Global Manager. For the local manager, if set, this needs to point to 'default'.
```
annakhm, line 44: I'm not sure this is relevant for `VPC`

website/docs/d/policy_vpc.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+  }
+  display_name = "vpc1"
+}
+```
+
+## Argument Reference
+
+* `id` - (Optional) The ID of VPC to retrieve. If ID is specified, no additional argument should be configured.
+* `display_name` - (Optional) The Display Name prefix of the VPC to retrieve.
+* `context` - (Optional) The context which the object belongs to
```
annakhm, line 32: This should be `Required`

nsxt/data_source_nsxt_policy_vpc.go
```diff
@@ -0,0 +1,57 @@
...
+			"id":                   getDataSourceIDSchema(),
+			"display_name":         getDataSourceDisplayNameSchema(),
+			"description":          getDataSourceDescriptionSchema(),
+			"path":                 getPathSchema(),
+			"context":              getContextSchema(true, false),
+			"site_info":            getSiteInfoSchema(),
+			"default_gateway_path": getPathSchema(),
+			"short_id": {
+				Type:     schema.TypeString,
+				Optional: true,
```
annakhm, line 25: According to the code this is fully computed, in which case `Optional: true` should be removed, so that terraform does not allow user to specify it

website/docs/d/policy_vpc.html.markdown
ksamoray: I'm not sure either - I reused the SiteInfo from Project as these two use the same schema. Should I remove site_path entirely from this DS?

nsxt/data_source_nsxt_policy_vpc.go
ksamoray: This value is assigned by NSX if not set by user, and therefore has to be computed. It can also be set by the user (it should be unique - NSX will enforce that).
The UI doesn't allow updating it (on create there's the text "This field can be set at VPC creation time and cannot be edited later. Max 8 characters allowed.").
BTW there's an open issue for this in the Project.

website/docs/d/policy_vpc.html.markdown
annakhm: @tvigneron could you please give your opinion?

annakhm (approved):

PR #1208, opened 2024-05-15
Update LB SSL tests for NSX v4.2.0

ksamoray:
SSL protocol support has changed in this version and test should adapt accordingly


salv-orlando (approved):
LGTM

Issue #1210, opened 2024-05-16
nsxt_policy_tier0_inter_vrf_routing bgp_route_leaking in_filter and out_filter


rvvliet78:
### Describe the bug

When adding an in_filter to the bgp_route_leaking it isn't getting applied.
However when adding only an out_filter, it gets applied as an in_filter

### Reproduction steps

Code
`resource "nsxt_policy_tier0_inter_vrf_routing" "inter-vrf-routing_tenant" {
  display_name = "${nsxt_policy_tier0_gateway.vrftenant-tier0.display_name}-inter-vrf-route"
  gateway_path = nsxt_policy_tier0_gateway.vrftenant-tier0.path
  target_path  = data.nsxt_policy_tier0_gateway.parent_tier0_vrf_gateway.path

  bgp_route_leaking {
    address_family = "IPV4"
    in_filter = [nsxt_policy_gateway_route_map.routemap_in.path]
    out_filter = [nsxt_policy_gateway_route_map.routemap_out.path]
  }
}`

Apply
`Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  module.vrf_tenant.nsxt_policy_tier0_inter_vrf_routing.inter-vrf-routing_tenant will be updated in-place
  ~ resource "nsxt_policy_tier0_inter_vrf_routing" "inter-vrf-routing_tenant" {
        id           = "da3e067a-7152-4794-84cb-53c21cafc277"
        # (6 unchanged attributes hidden)

      ~ bgp_route_leaking {
          ~ in_filter      = [
              + "/infra/tier-0s/84f40d06-72a2-44c2-adb6-913fd11def9c/route-maps/3f0de474-ee94-460c-9b14-e7375db56561",
            ]
          ~ out_filter     = [
              + "/infra/tier-0s/84f40d06-72a2-44c2-adb6-913fd11def9c/route-maps/19153ec6-dbd0-4c70-8e45-fcc1f2e754c8",
            ]
            # (1 unchanged attribute hidden)
        }
    }

Plan: 0 to add, 1 to change, 0 to destroy.
module.vrf_tenant.nsxt_policy_tier0_inter_vrf_routing.inter-vrf-routing_tenant: Modifying... [id=da3e067a-7152-4794-84cb-53c21cafc277]
[...]

PR #1211, opened 2024-05-16
Update out_filter in nsxt_policy_gateway_route_map

ksamoray:
The code assigns the values into the wrong variable and therefore changes aren't applied correctly.

Fixes: #1210


annakhm (approved):

annakhm:
would be nice for the test to cover this

PR #1212, opened 2024-05-17
Add ignore_tags feature for segments

annakhm:
Other tools (like VCD) may add tags to certain NSX objects managed by terraform, and expect those tags to be persisted. However, terraform provider assumes itself to be the only source of truth, and in case described above it detects drift in tags, and deletes them with apply.
This PR allows user to specify a list of tag scopes that should be ignored by the provider, more specifically:
1. diff should not be detected when such tags are present on NSX but not in terraform config
2. when provider applies tags from its own config, tags from the `ignore` list should not be deleted.

Implementation suggested here changes the Read function to sort tags into two buckets: regular tags and ignored tags. Regular tags are managed as before, whilw ignored tags (computed attribute) are appended to regular tag list on each apply.

Future enhancements:
1. Add regexp to compare scopes
2. Add this feature to more resources if needed

### DIFF
--- a/nsxt/policy_common.go
+++ b/nsxt/policy_common.go
@@ -395,6 +395,43 @@ func getPolicySecurityPolicySchema(isIds, withContext, withRule bool) map[string
 	return result
 }
 
+func getIgnoreTagsSchema() *schema.Schema {
+	return &schema.Schema{
+		Type:     schema.TypeList,
+		Optional: true,
+		MaxItems: 1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"scopes": {
+					Type:        schema.TypeList,
+					Description: "List of scopes to ignore",
+					Required:    true,
+					Elem: &schema.Schema{
+						Type: schema.TypeString,
+					},
+				},
+				"detected": {
+					Type:        schema.TypeSet,
+					Description: "Tags matching scopes to ignore",
+					Computed:    true,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"scope": {
+								Type:     schema.TypeString,
+								Computed: true,
+							},
+							"tag": {
+								Type:     schema.TypeString,
+								Computed: true,
+							},
+						},
+					},
+				},
+			},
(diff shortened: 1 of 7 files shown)


ksamoray:
If VCD or such uses the same tag names across various resources, would it make sense to set these at the provider level, rather than on each resource individually?
That would be less cumbersome and prone to errors maybe - as if an environment is managed by both VCD and TF, these tags can be set at the provider level, then from there and on affect any resource. Instead in this solution a user should remember to set the `ignore_tags` attribute everywhere.

annakhm:
> If VCD or such uses the same tag names across various resources, would it make sense to set these at the provider level, rather than on each resource individually? That would be less cumbersome and prone to errors maybe - as if an environment is managed by both VCD and TF, these tags can be set at the provider level, then from there and on affect any resource. Instead in this solution a user should remember to set the `ignore_tags` attribute everywhere.

I have chosen to expose attribute on resource level for two reasons:
- We need to store the state of ignored tags per resource, so new computed attribute is needed anyway
- Scopes might be different for individual resources

nsxt/policy_common.go
```diff
@@ -395,6 +395,44 @@ func getPolicySecurityPolicySchema(isIds, withContext, withRule bool) map[string
...
+					Elem: &schema.Schema{
+						Type: schema.TypeString,
+					},
+				},
+				"detected": {
+					Type:        schema.TypeSet,
+					Description: "Tags matching scopes to ignore",
+					Computed:    true,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
```
ksamoray, line 419: Can `getTagsSchemaInternal()` be recycled here in some way?
Other than this LGTM

annakhm: This one is fully computed, I didn't want to add to many params to `getTagsSchemaInternal`

ksamoray (approved):

salv-orlando (commented):
All good, but I think we should have this capability validated in acceptance tests.
It won't be a trivial test, but also not a very complex one

PR #1216, opened 2024-05-24
Escape password when requesting for session token

annakhm:
When NSX password contains certain special characters(such as &), those need to be encoded when sent to session/create API


salv-orlando (approved):
lgtm

ksamoray (approved):

PR #1217, opened 2024-05-27
VPC E-W security policy resource

ksamoray:
Implement the resource for east-west VPC security policy.

### DIFF
--- a/nsxt/provider.go
+++ b/nsxt/provider.go
@@ -496,6 +496,7 @@ func Provider() *schema.Provider {
 			"nsxt_policy_gateway_flood_protection_profile_binding":     resourceNsxtPolicyGatewayFloodProtectionProfileBinding(),
 			"nsxt_policy_compute_sub_cluster":                          resourceNsxtPolicyComputeSubCluster(),
 			"nsxt_policy_tier0_inter_vrf_routing":                      resourceNsxtPolicyTier0InterVRFRouting(),
+			"nsxt_vpc_security_policy":                                 resourceNsxtVPCSecurityPolicy(),
 		},
 
 		ConfigureFunc: providerConfigure,
--- a/nsxt/resource_nsxt_policy_predefined_security_policy.go
+++ b/nsxt/resource_nsxt_policy_predefined_security_policy.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
+	nsxt "github.com/vmware/vsphere-automation-sdk-go/services/nsxt"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 
 	"github.com/vmware/terraform-provider-nsxt/api/infra/domains"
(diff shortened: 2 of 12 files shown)


nsxt/provider.go
```diff
@@ -1201,28 +1202,35 @@ func getGlobalPolicyEnforcementPointPath(m interface{}, sitePath *string) string
...
 
-func getProjectIDFromSchema(d *schema.ResourceData) string {
+func getContextDataFromSchema(d *schema.ResourceData) (string, string) {
 	ctxPtr := d.Get("context")
 	if ctxPtr != nil {
 		contexts := ctxPtr.([]interface{})
 		for _, context := range contexts {
 			data := context.(map[string]interface{})
+			vpcID := ""
+			if data["vpi_id"] != nil {
```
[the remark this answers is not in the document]
ksamoray: Indeed

annakhm:
This PR does not cover the provider-defined VPC scope yet, correct?

ksamoray:
> This PR does not cover the provider-defined VPC scope yet, correct?

Indeed - the provider-level VPC specification is a global, non per-resource change.

nsxt/provider.go
```diff
@@ -496,6 +496,7 @@ func Provider() *schema.Provider {
 			"nsxt_policy_gateway_flood_protection_profile_binding":     resourceNsxtPolicyGatewayFloodProtectionProfileBinding(),
 			"nsxt_policy_compute_sub_cluster":                          resourceNsxtPolicyComputeSubCluster(),
 			"nsxt_policy_tier0_inter_vrf_routing":                      resourceNsxtPolicyTier0InterVRFRouting(),
+			"nsxt_policy_vpc_security_policy":                          resourceNsxtPolicyVPCSecurityPolicy(),
```
annakhm, line 499: can we remove the first `policy` for all VPC resources? There is no other VPC.. Would love to hear also from @salv-orlando and @tvigneron

qiyueyao:
Hi, a quick maybe dumb question: does this series of change support `data_source_nsxt_policy_security_policy` and `data_source_nsxt_policy_gateway_policy`? Looks like for Get with ID, it's handled, but without ID, it still lists the `domain` 🤔

ksamoray:
> Hi, a quick maybe dumb question: does this series of change support `data_source_nsxt_policy_security_policy` and `data_source_nsxt_policy_gateway_policy`? Looks like for Get with ID, it's handled, but without ID, it still lists the `domain` 🤔

@qiyueyao I haven't got to handle these data sources yet but yeah - unless they use search API, they'll be affected by this change as the API wrapper is modified.
I believe that I should add a check to the DS, that the VPC is not specified - as these datasources aren't supposed to serve VPC.
VPC should have a pair of `data_source_nsxt_policy_vpc.*` datasources.

nsxt/provider.go
salv-orlando: I agree, also proposed that in design review.

tvigneron: It seems a great idea - for vpc resources let's get rid of the Policy. We don't want the name to be unecessarily long

nsxt/resource_nsxt_vpc_security_policy_test.go
```diff
@@ -0,0 +1,174 @@
...
+	tag2 := "def"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccOnlyVPC(t)
+		},
+		Providers: testAccProviders,
+		CheckDestroy: func(state *terraform.State) error {
+			return testAccNsxtPolicySecurityPolicyCheckDestroy(state, updatedName, defaultDomain)
```
ksamoray, line 36: Should pass the resource name or use a VPC func here

nsxt/resource_nsxt_policy_predefined_security_policy.go
```diff
@@ -409,20 +464,29 @@ func resourceNsxtPolicyPredefinedSecurityPolicyDelete(d *schema.ResourceData, m
...
+		childVPC, err := createChildVPCWithSecurityPolicy(context, *policy.Id, policy)
+		if err != nil {
+			return fmt.Errorf("Failed to create H-API for VPC Security Policy: %s", err)
+		}
+		orgRoot := model.OrgRoot{
+			ResourceType: strPtr("OrgRoot"),
+			Children:     []*data.StructValue{childVPC},
+		}
+
+		client := nsxt.NewOrgRootClient(connector)
```
annakhm, line 478: This would work with VPC admin priviledge?

annakhm (approved):

website/docs/r/vpc_security_policy.html.markdown
```diff
@@ -0,0 +1,131 @@
...
+
+An existing security policy can be [imported][docs-import] into this resource, via the following command:
+
+[docs-import]: https://www.terraform.io/cli/import
+
+```
+terraform import nsxt_vpc_security_policy.policy1 PATH
+```
+
+The above command imports the VPC security policy named `policy1` under NSX domain `domain` with the NSX Policy path `PATH`.
```
annakhm, line 131: `domain` is irrelevant here

annakhm:
LGTM. There are some doc fixes needed, but feel free to address them in one of the following PRs if you prefer

PR #1218, opened 2024-05-28
Add utility to parse standard policy path

annakhm:
This utility will be used in generated resources


nsxt/policy_utils.go
```diff
@@ -188,6 +188,55 @@ func getResourceIDFromResourcePath(rPath string, rType string) string {
 	return ""
 }
 
+func parseStandardPolicyPath(path string) ([]string, error) {
```
salv-orlando, line 191: Do you just need to collect the list of identifiers in the policy path, or will it also matter figuring out what is what?

For instance in:
/orgs/myorg/projects/myproj/infra/tier-1s/mygw1"
parent[2] is a router id

but in:
/orgs/myorg/projects/myproj/vpcs/myvpc/tier-1s/mygw1
parent[2] is a VPC id

ksamoray:
Can we utilize this (or share code with) in [nsxtPolicyPathResourceImporterHelper](https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/policy_utils.go#L237)? I've already added VPC support in PR  #1217

nsxt/policy_utils.go
annakhm: Parent list will be used as parameters for API calls, so in case of orgs/projects/vpc the ids will be used as parameters in API call in the same way as gw id or segment id would be used.

annakhm:
> Can we utilize this (or share code with) in [nsxtPolicyPathResourceImporterHelper](https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/policy_utils.go#L237)? I've already added VPC support in PR #1217

Good point, I think `nsxtPolicyPathResourceImporterHelper` can use somewhat enhanced version of `parseStandardPolicyPath`. I can refactor it in a follow up OR when all involved PRs are merged

nsxt/policy_utils_test.go
```diff
@@ -0,0 +1,101 @@
...
+		{
+			path:    "/global-infra/tier-1s/t15/ipsec-vpn-services/default/sessions/xxx-yyy-xxx",
+			parents: []string{"t15", "default", "xxx-yyy-xxx"},
+		},
+		{
+			path:    "/infra/evpn-tenant-configs/{config-id}",
+			parents: []string{"{config-id}"},
+		},
+		{
+			path:    "/infra/tier-0s/{tier-0-id}/locale-services/{locale-service-id}/service-interfaces/{interface-id}",
```
salv-orlando, line 49: silly question: are the '{ }' brackets useful for testing something in particular, or is it just a copy & paste thing?

nsxt/policy_utils_test.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+type policyPathTest struct {
+	path    string
+	parents []string
+}
+
+func TestParseStandardPolicyPath(t *testing.T) {
+
+	testData := []policyPathTest{
```
salv-orlando, line 19: nit: add negative test with invalid path and verify code fails as expected.

nsxt/policy_utils.go
```diff
@@ -188,6 +188,67 @@ func getResourceIDFromResourcePath(rPath string, rType string) string {
...
+			// vpc paths do not contain infra
+			infraPath = false
+		}
+	}
+	if len(segments) <= idx {
+		return nil, fmt.Errorf("unexpected policy path %s", path)
+	}
+	if infraPath {
+		// continue after infra marker
+		if segments[idx] != "infra" && segments[idx] != "global-infra" {
```
salv-orlando, line 225: I don't remember a case where we can have global-infra for a project-scoped resource.
At the same time, I don't think that a special case to handle only "infra" when in org scope is mandatory

salv-orlando (commented):
LGTM, lmk if we can do a little bit more for testing.

nsxt/policy_utils_test.go
annakhm: I thought we might test special characters as well :)

nsxt/policy_utils_test.go
annakhm: There is a negative test below

PR #1223, opened 2024-06-02
VPC N-S security policy resource

ksamoray:
Implement the resource for north-south VPC security policy.

### DIFF
--- a/nsxt/policy_common.go
+++ b/nsxt/policy_common.go
@@ -310,13 +310,13 @@ func getSecurityPolicyAndGatewayRuleSchema(scopeRequired bool, isIds bool, nsxID
 	return ruleSchema
 }
 
-func getPolicyGatewayPolicySchema() map[string]*schema.Schema {
-	secPolicy := getPolicySecurityPolicySchema(false, true, true, true)
+func getPolicyGatewayPolicySchema(withDomain bool) map[string]*schema.Schema {
+	secPolicy := getPolicySecurityPolicySchema(false, true, true, withDomain)
 	// GW Policies don't support scope
 	delete(secPolicy, "scope")
 	secPolicy["category"].ValidateFunc = validation.StringInSlice(gatewayPolicyCategoryWritableValues, false)
 	// GW Policy rules require scope to be set
-	secPolicy["rule"] = getSecurityPolicyAndGatewayRulesSchema(true, false, true)
+	secPolicy["rule"] = getSecurityPolicyAndGatewayRulesSchema(withDomain, false, true)
 	return secPolicy
 }
 
--- a/nsxt/resource_nsxt_policy_predefined_gateway_policy.go
+++ b/nsxt/resource_nsxt_policy_predefined_gateway_policy.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	nsxt "github.com/vmware/vsphere-automation-sdk-go/services/nsxt"
 
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
@@ -264,6 +265,21 @@ func createChildDomainWithGatewayPolicy(domain string, policyID string, policy m
 }
 
 func gatewayPolicyInfraPatch(context utl.SessionContext, policy model.GatewayPolicy, domain string, m interface{}) error {
+	connector := getPolicyConnector(m)
+	if context.ClientType == utl.VPC {
+		childVPC, err := createChildVPCWithGatewayPolicy(context, *policy.Id, policy)
+		if err != nil {
+			return fmt.Errorf("failed to create H-API for VPC Gateway Policy: %s", err)
+		}
+
+		orgRoot := model.OrgRoot{
+			ResourceType: strPtr("OrgRoot"),
+			Children:     []*data.StructValue{childVPC},
+		}
+
+		client := nsxt.NewOrgRootClient(connector)
+		return client.Patch(orgRoot, nil)
+	}
 	childDomain, err := createChildDomainWithGatewayPolicy(domain, *policy.Id, policy)
 	if err != nil {
 		return fmt.Errorf("Failed to create H-API for Predefined Gateway Policy: %s", err)
@@ -293,7 +309,7 @@ func updatePolicyPredefinedGatewayPolicy(id string, d *schema.ResourceData, m in
 		return fmt.Errorf("Failed to extract domain from Gateway Policy path %s", path)
 	}
 
-	predefinedPolicy, err := getGatewayPolicyInDomain(getSessionContext(d, m), id, domain, connector)
+	predefinedPolicy, err := getGatewayPolicy(getSessionContext(d, m), id, domain, connector)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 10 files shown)


nsxt/resource_nsxt_policy_predefined_gateway_policy.go
```diff
@@ -263,7 +264,22 @@ func createChildDomainWithGatewayPolicy(domain string, policyID string, policy m
 	return dataValue.(*data.StructValue), nil
 }
 
-func gatewayPolicyInfraPatch(context utl.SessionContext, policy model.GatewayPolicy, domain string, m interface{}) error {
+func gatewayPolicyInfraPatch(context utl.SessionContext, policy model.GatewayPolicy, domain string, m interface{}, withDomain bool) error {
```
salv-orlando, line 267: How are you using withDomain in this function?

nsxt/policy_common.go
```diff
@@ -310,13 +310,13 @@ func getSecurityPolicyAndGatewayRuleSchema(scopeRequired bool, isIds bool, nsxID
...
-func getPolicyGatewayPolicySchema() map[string]*schema.Schema {
-	secPolicy := getPolicySecurityPolicySchema(false, true, true, true)
+func getPolicyGatewayPolicySchema(withDomain bool) map[string]*schema.Schema {
+	secPolicy := getPolicySecurityPolicySchema(false, true, true, withDomain)
 	// GW Policies don't support scope
 	delete(secPolicy, "scope")
 	secPolicy["category"].ValidateFunc = validation.StringInSlice(gatewayPolicyCategoryWritableValues, false)
 	// GW Policy rules require scope to be set
-	secPolicy["rule"] = getSecurityPolicyAndGatewayRulesSchema(true, false, true)
+	secPolicy["rule"] = getSecurityPolicyAndGatewayRulesSchema(withDomain, false, true)
```
salv-orlando, line 319: I thought the first parameter is for scope required.
if withDomain = False and then we have a VPC resource, does this make scope optional?

nsxt/resource_nsxt_policy_vpc_gateway_policy.go
```diff
@@ -0,0 +1,163 @@
...
+		},
+
+		Schema: getPolicyGatewayPolicySchema(false),
+	}
+}
+
+func resourceNsxtPolicyVPCGatewayPolicyCreate(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+
+	// Initialize resource Id and verify this ID is not yet used
```
annakhm, line 35: If `domain` is the only difference between this function and `resourceNsxtPolicyGatewayPolicyCreate`, would it make sense to parametrize this single value here and below?

nsxt/resource_nsxt_policy_predefined_gateway_policy.go
ksamoray: It's redundant, I'll remove that argument.

nsxt/policy_common.go
ksamoray: I guess that I've made it so to make scope optional for VPC - but I'm not sure that it should be so.

nsxt/resource_nsxt_policy_vpc_gateway_policy.go
ksamoray: Both funcs barely have any logic. I don't see much value in adding excess of functions which in the end, do nothing, pretty much like in `Delete()`

nsxt/resource_nsxt_vpc_gateway_policy.go
```diff
@@ -0,0 +1,163 @@
...
+
+func resourceNsxtVPCGatewayPolicyRead(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+
+	id := d.Id()
+	if id == "" {
+		return fmt.Errorf("error obtaining VPC Gateway Policy ID")
+	}
+
+	obj, err := getGatewayPolicyInDomain(getSessionContext(d, m), id, "", connector)
```
annakhm, line 60: I understand the reluctance to rename common helpers, but I personally find `domain` confusing here.. I think a comment explaining that we're not actually looking in domain for VPC would help future code readers. 
For a long run, I think we might want to consider moving `domain` into the context

website/docs/r/vpc_gateway_policy.html.markdown
```diff
@@ -0,0 +1,123 @@
...
+  tcp_strict      = false
+
+  tag {
+    scope = "color"
+    tag   = "orange"
+  }
+
+  rule {
+    display_name       = "rule1"
+    destination_groups = [nsxt_policy_group.group1.path, nsxt_policy_group.group2.path]
```
annakhm, line 50: `nsxt_vpc_group` would be the more obvious use case?

website/docs/r/vpc_gateway_policy.html.markdown
```diff
@@ -0,0 +1,123 @@
...
+  * `vpc_id` - (Required) The ID of the VPC which the object belongs to
+* `comments` - (Optional) Comments for this Gateway Policy including lock/unlock comments.
+* `locked` - (Optional) A boolean value indicating if the policy is locked. If locked, no other users can update the resource.
+* `sequence_number` - (Optional) An int value used to resolve conflicts between security policies
+* `stateful` - (Optional) A boolean value to indicate if this Policy is stateful. When it is stateful, the state of the network connects are tracked and a stateful packet inspection is performed.
+* `tcp_strict` - (Optional) A boolean value to enable/disable a 3 way TCP handshake is done before the data packets are sent.
+* `rule` (Optional) A repeatable block to specify rules for the Gateway Policy. Each rule includes the following fields:
+  * `display_name` - (Required) Display name of the resource.
+  * `description` - (Optional) Description of the resource.
+  * `destination_groups` - (Optional) Set of group paths that serve as the destination for this rule. IPs, IP ranges, or CIDRs may also be used starting in NSX-T 3.0. An empty set can be used to specify "Any".
```
annakhm, line 84: lets remove `3.0` mention here and below as its irrelevant for VPC

website/docs/r/vpc_gateway_policy.html.markdown
```diff
@@ -0,0 +1,123 @@
...
+  * `destinations_excluded` - (Optional) A boolean value indicating negation of destination groups.
+  * `direction` - (Optional) The traffic direction for the policy. Must be one of: `IN`, `OUT` or `IN_OUT`. Defaults to `IN_OUT`.
+  * `disabled` - (Optional) A boolean value to indicate the rule is disabled. Defaults to `false`.
+  * `ip_version` - (Optional) The IP Protocol for the rule. Must be one of: `IPV4`, `IPV6` or `IPV4_IPV6`. Defaults to `IPV4_IPV6`.
+  * `logged` - (Optional) A boolean flag to enable packet logging.
+  * `notes` - (Optional) Text for additional notes on changes for the rule.
+  * `profiles` - (Optional) A list of context profiles for the rule. Note: due to platform issue, this setting is only supported with NSX 3.2 onwards.
```
annakhm, line 91: Lets remove NSX version mention here as well

nsxt/resource_nsxt_vpc_gateway_policy.go
ksamoray: I think that moving domain (or generally anything that's in the URL part of the API) into context makes some sense, but I'd rather avoid this now - it's a major effort and a compatibility issue.
For the func name change - I'll change it, I missed that somehow.

website/docs/r/vpc_gateway_policy.html.markdown
salv-orlando: FQDN network policies should work for VPCs as well.

website/docs/r/vpc_gateway_policy.html.markdown
```diff
@@ -0,0 +1,123 @@
...
+
+An existing VPC Gateway Policy can be [imported][docs-import] into this resource, via the following command:
+
+[docs-import]: https://www.terraform.io/cli/import
+
+```
+terraform import nsxt_vpc_gateway_policy.policy1 PATH
+```
+
+The above command imports the VPC gateway policy named `policy1` under NSX domain `domain` with the NSX Policy path `PATH`.
```
annakhm, line 123: lets remove `domain` here

annakhm (approved):

PR #1224, opened 2024-06-03
Handle invalid context errors

ksamoray:
When a resource or datasource is defined with an invalid context, e.g with a VPC for a resource with no VPC implementation, the provider crashes as the client creation is not validated.
    
This PR addresses this issue.


salv-orlando (approved):
LGTM!

PR #1226, opened 2024-06-04
Configure project and VPC at provider level

ksamoray:
Allow configuration of the attributes above in the provider globally for the whole context, instead of the resource level.


ksamoray:
I think that I should add some tests for this case as well.

annakhm:
I suggest erroring out in case different vpcs are specified on provider level and in the resource

PR #1228, opened 2024-06-05
Fix set IP block visibility on READ if NSX version too low

### LINKED ISSUE #1227: visibility not set on creation of nsxt_policy_ip_block
### Describe the bug

If you create a `nsxt_policy_ip_block` resource with the `visibility` attribute set, the value does not seem to be set on the initial creation.  If you do a plan after creating the resource, you will see an update to set the `visibility` attribute.

### Reproduction steps

1. Apply a plan that creates a new `nsxt_policy_ip_block` resource with the `visibility` attribute set.
2. Perform another terraform plan and observe that an update to the `visibility` attribute is needed.

### Expected behavior

The value of `visibility` is set when the resource is initially created and a second plan and apply is not needed.

### Additional context

_No response_

wsquan171:
This change fixes setting policy IP block visibility on READ if NSX is below 4.2, where the value will always be nil. This will cause tf to always show diff if the resource spec has non-empty value.

Closes #1227

### DIFF
--- a/nsxt/resource_nsxt_policy_ip_block.go
+++ b/nsxt/resource_nsxt_policy_ip_block.go
@@ -92,7 +92,9 @@ func resourceNsxtPolicyIPBlockRead(d *schema.ResourceData, m interface{}) error 
 	d.Set("path", block.Path)
 	d.Set("revision", block.Revision)
 	d.Set("cidr", block.Cidr)
-	d.Set("visibility", block.Visibility)
+	if util.NsxVersionHigherOrEqual("4.2.0") {
+		d.Set("visibility", block.Visibility)
+	}
 
 	return nil
 }


ksamoray:
LGTM, would like @annakhm to have a look as well as this behavior is not consistent throughout the provider code.
Maybe we could at least warn that this attribute has been ignored?

salv-orlando (approved):
LGTM!

annakhm:
What happens in NSX 4.2 if this setting is unspecified? Does NSX send back a default?

annakhm:
> @annakhm this param is unknown to NSX < 4.2. On read it will be empty value, and if included on write it will error.

My question was for NSX 4.2. This param is optional, so user can skip it. However, for such enumerations NSX would normally  send back a default (not always though). I wanted to verify we don't need to set this param as `Computed` for NSX 4.2, regardless of earlier version

wsquan171:
> I wanted to verify we don't need to set this param as `Computed` for NSX 4.2, regardless of earlier version

For NSX 4.2, if the value is unset, NSX will send back as-is (also unset). This doesn't need to be computed.

annakhm (approved):

PR #1229, opened 2024-06-10
Upgrade guide document

ksamoray:
Add a document describing the upgrade process.
Plus improve precheck failure and warning reporting, as it lacks documentation on NSX side.

### DIFF
--- a/website/docs/guides/upgrade.html.markdown
+++ b/website/docs/guides/upgrade.html.markdown
@@ -1,0 +1,200 @@
+---
+layout: "nsxt"
+page_title: "VMware NSX Terraform Provider NSX upgrade support"
+description: |-
+  VMware NSX Terraform Provider NSX upgrade support
+---
+
+# NSX Provider for NSX Upgrade
+
+NSX Terraform Provider offers support for [upgrading NSX components](https://docs.vmware.com/en/VMware-NSX/4.1/upgrade/GUID-E04242D7-EF09-4601-8906-3FA77FBB06BD.html).
+
+The provider support consists of several resources and data sources which can perform a complete upgrade cycle.
+The NSX Terraform provider support upgrades for NSX on ESX hosts, NSX Edge appliances, and NSX manager appliances.
+
+**NOTE:** The upgrade process using the Terraform NSXT provider assumes that the Terraform NSX provider uses NSX upgrade 
+components exclusively. Therefore, concurrent changess to upgrade components via other means, e.g. UI, might fail the process.
+**NOTE:** The NSX upgrade process should start each upgrade with a clean state. So whenever an upgrade is executed, e.g. 
+from v3.2.3 to v4.1.1 the Terraform state file should be retained until the upgrade is completed. Then, if the same plan 
+is used later on to upgrade from v4.1.1 to v4.2.0, the Terraform state file should be deleted prior to applying the plan.
+
+## Upgrade process steps
+
+### Preparation for upgrade
+
+Upgrade preparation consists of several actions, and is performed using [nsxt_upgrade_prepare](../resources/upgrade_prepare.html.markdown):
+* Uploading the upgrade and precheck bundles.
+* Acceptance of the user agreement.
+* Reporting of failed pre-checks.
+
+**NOTE:** Only one `nsxt_upgrade_prepare` resource should be used in the upgrade plan.
+
+```hcl
+resource "nsxt_upgrade_prepare" "prepare_res" {
+  upgrade_bundle_url    = "http://url-to-upgrade-bundle.mub"
+  precheck_bundle_url   = "http://url-to-precheck-bundle.pub"
+  accept_user_agreement = true
+}
+```
+
+**NOTE:** Acceptance of the license agreement is required to initiate the upgrade process.
+**NOTE:** Precheck bundle is usable only while upgrading a NSX manager of version v4.1.1 and above.
+
+### Checking the readiness for upgrade
+
+When preparation is complete, it is required to validate the readiness for upgrade using the 
+[nsxt_upgrade_prepare_ready](../data-sources/upgrade_prepare_ready.html.markdown) data source.
+When NSX preparation results with failures or warnings, the data source will fail the upgrade.
+Warnings should be acknowledged as described below.
+
+For example:
+```
+Error: 
+There are unacknowledged warnings in prechecks:
+  Component: EDGE, code: pUBCheck, description: Checks if new version of pub available on VMware Download site.
+  Component: MP, code: backupOperationCheck, description: Warns if backup is not taken recently and blocks management plane upgrade if backup is in progress
+
+Please address these errors from NSX or using nsxt_upgrade_precheck_acknowledge resource
+```
+
+Prechecks which are in `failure' or in warning state can also be obtained from the `nsxt_upgrade_prepare` resource, 
+using its `failed_prechecks` attribute.
+
+### Acknowledgement of upgrade issues and warnings
+
+The execution of `nsxt_upgrade_prepare` and `nsxt_upgrade_prepare_ready` could result with various warnings which would 
+halt the upgrade process. However, warnings can be suppressed with the [nsxt_upgrade_precheck_acknowledge](../resources/upgrade_precheck_acknowledge.html.markdown) resource.
+
+The resource uses the `precheck_ids` to conclude which prechecks are safe to ignore for this upgrade cycle.
+The `nsxt_upgrade_precheck_acknowledge` resource requires the `target_version` from the `nsxt_upgrade_prepare` resource 
+to make sure that the acknowledged warnings are from the correct execution, and to enforce execution of the 
+`nsxt_upgrade_precheck_acknowledge` resource after preparation is complete.
+
+The `nsxt_upgrade_prepare_ready` data source verifies that the preparation phase of the upgrade was concluded successfully 
+and enable the execution of the upgrade. In case that there are still open issues, these will raise as described above.
+
+```hcl
+resource "nsxt_upgrade_precheck_acknowledge" "test" {
+  precheck_ids   = ["backupOperationCheck", "pUBCheck"]
+  target_version = nsxt_upgrade_prepare.prepare_res.target_version
+}
+
+data "nsxt_upgrade_prepare_ready" "ready" {
+  upgrade_prepare_id = nsxt_upgrade_prepare.prepare_res.id
+  depends_on         = [nsxt_upgrade_precheck_acknowledge.test]
+}
+```
+
+### Running the upgrade
+
+In order to configure and execute upgrade of NSXT edges, hosts, and managers, use [nsxt_upgrade_run](../resources/upgrade_run.html.markdown) resource. 
+
+This resource will start with configuration of upgrade unit groups and upgrade plan settings for `EDGE`and `HOST` 
+components. Host component groups may be created within this resource with custom host lists and attributes, or it can 
+use the predefined groups which NSX provides.
+
+Following the configuration of the edge, host groups, it will execute the upgrade for each component. If there are 
+either `EDGE` or `HOST` groups excluded from the upgrade, the corresponding component will be paused after enabled 
+groups get upgraded. At the same time, `MP` upgrade won't be processed, because that requires `EDGE` and `HOST` to be 
+fully upgraded. Refer to the exposed attribute for current upgrade state details. For example, check `upgrade_plan`
+for current upgrade plan, which also includes the plan not specified in this resource and `state` for upgrade status of 
+each component and UpgradeUnitGroups in the component. For more details, please check NSX admin guide.
+
+The order of the groups within the `nsxt_upgrade_run` resource defined the order of the upgrade execution.
+
+In the example below, host_group `HostWithDifferentSettings` will use different settings than the defaulted ones. The 
+group configuration is initiated with the `nsxt_upgrade_run` while the other groups in this example are predefined and 
+consumed using the data sources `nsxt_edge_upgrade_group` and `nsxt_host_upgrade_group`.
+
+**NOTE:** Only one `nsxt_upgrade_run` resource should be used in the upgrade plan.
+
+```hcl
+resource "nsxt_upgrade_run" "run" {
+  upgrade_prepare_ready_id = data.nsxt_upgrade_prepare_ready.ready.id
+
+  edge_group {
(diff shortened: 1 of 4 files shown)


website/docs/guides/upgrade.html.markdown
```diff
@@ -0,0 +1,158 @@
+---
+layout: "nsxt"
+page_title: "VMware NSX Terraform Provider NSX upgrade support"
+description: |-
+  The VMware NSX Terraform Provider NSX upgrade support
```
salv-orlando, line 5: Maybe "The" is not needed.

website/docs/guides/upgrade.html.markdown
```diff
@@ -0,0 +1,158 @@
...
+---
+
+# NSX Provider for NSX Upgrade
+
+NSX Terraform Provider offers support for [upgrading NSX components](https://docs.vmware.com/en/VMware-NSX/4.1/upgrade/GUID-E04242D7-EF09-4601-8906-3FA77FBB06BD.html).
+
+The provider support consists of several resources and data sources which can perform a complete upgrade cycle when applied correctly.
```
salv-orlando, line 12: maybe "when applied correctly" is redundant?

website/docs/guides/upgrade.html.markdown
```diff
@@ -0,0 +1,158 @@
...
+The NSX upgrade process upgrades the manager components of NSX, Edge appliances, and the NSX bits on the compute hosts.
```
salv-orlando, line 13: Another pedant suggestion, consider replacing with:

"The NSX Terraform provider support upgrades for NSX on ESX hosts, NSX Edge appliances, and NSX manager appliances"

website/docs/guides/upgrade.html.markdown
```diff
@@ -0,0 +1,158 @@
...
+to make sure that the acknowledged warnings are from the correct execution, and to enforce execution of the 
+`nsxt_upgrade_precheck_acknowledge` resource after preparation is complete.
+
+The `nsxt_upgrade_prepare_ready` data source verifies that the preparation phase of the upgrade was concluded 
+successfully and enable the execution of the upgrade. 
+
+```hcl
+resource "nsxt_upgrade_precheck_acknowledge" "test" {
+  precheck_ids   = ["backupOperationCheck", "pUBCheck"]
+  target_version = nsxt_upgrade_prepare.test.target_version
```
salv-orlando, line 57: Fore completeness, you might want to show also the nsxt_upgrade_prepare resource.
And if you're actually referring to the previous one, it might worth replacing `nsxt_upgrade_prepare.test` with `nsxt_upgrade_prepare.prepare_res`

website/docs/guides/upgrade.html.markdown
```diff
@@ -0,0 +1,158 @@
...
+
+The preparation stage may result with various warnings which would halt the upgrade process. These warnings can be 
+suppressed with the `nsxt_upgrade_precheck_acknowledge` resource.
+
+For example:
+```
+There are unacknowledged warnings in precheck: [backupOperationCheck], please address these errors from NSX or using nsxt_upgrade_precheck_acknowledge resource
+```
+
+The resource uses the `precheck_ids` to conclude which prechecks are safe to ignore for this upgrade cycle.
```
salv-orlando, line 46: Can we add a link to NSX docs showing the list of precheck_ids?

website/docs/guides/upgrade.html.markdown
```diff
@@ -0,0 +1,158 @@
...
+### Running the upgrade
+
+In order to configure and execute upgrade of NSXT edges, hosts, and managers, use `nsxt_upgrade_run` resource. 
+
+This resource will start with configuration of upgrade unit groups and upgrade plan settings for `EDGE`and `HOST` 
+components. Host component groups may be created within this resource with custom host lists and attributes, or it can 
+use the predefined groups which NSX provides.
+
+Following the configuration of the edge, host groups, it will execute the upgrade for each component. If there are 
+either `EDGE` or `HOST` groups disabled from the upgrade, the corresponding component will be paused after enabled
```
salv-orlando, line 75: nit: replace disabled with excluded

website/docs/guides/upgrade.html.markdown
```diff
@@ -0,0 +1,158 @@
...
+    post_upgrade_check = true
+  }
+
+  host_upgrade_setting {
+    parallel           = true
+    post_upgrade_check = true
+  }
+
+  host_group {
+    display_name = "TEST123"
```
salv-orlando, line 120: did you add this group for a different purpose, it's not mentioned in the text above, but it looks like you could have written:

host_group `HostWithDifferentSettings` and `TEST123` will use different settings than the defaulted ones

website/docs/guides/upgrade.html.markdown
```diff
@@ -0,0 +1,158 @@
...
+In order to configure and execute upgrade of NSXT edges, hosts, and managers, use `nsxt_upgrade_run` resource. 
+
+This resource will start with configuration of upgrade unit groups and upgrade plan settings for `EDGE`and `HOST` 
+components. Host component groups may be created within this resource with custom host lists and attributes, or it can 
+use the predefined groups which NSX provides.
+
+Following the configuration of the edge, host groups, it will execute the upgrade for each component. If there are 
+either `EDGE` or `HOST` groups disabled from the upgrade, the corresponding component will be paused after enabled 
+groups get upgraded. At the same time, `MP` upgrade won't be processed, because that requires `EDGE` and `HOST` to be 
+fully upgraded. Refer to the exposed attribute for current upgrade state details. For example, check `upgrade_plan`
```
salv-orlando, line 77: We can also have multiple nsxt_ugprade_run resources? In that case could we add to the sample below a 2nd resource that will upgrade the disabled edge group and then MP?

website/docs/guides/upgrade.html.markdown
```diff
@@ -0,0 +1,158 @@
...
+for current upgrade plan, which also includes the plan not specified in this resource and `state` for upgrade status of 
+each component and UpgradeUnitGroups in the component. For more details, please check NSX admin guide.
+
+In the example below, host_group `HostWithDifferentSettings` will use different settings than the defaulted ones. The
```
salv-orlando, line 81: It might be good to spend one or two lines discussing the default settings.

website/docs/guides/upgrade.html.markdown
```diff
@@ -0,0 +1,158 @@
...
+This resource will start with configuration of upgrade unit groups and upgrade plan settings for `EDGE`and `HOST` 
+components. Host component groups may be created within this resource with custom host lists and attributes, or it can 
+use the predefined groups which NSX provides.
+
+Following the configuration of the edge, host groups, it will execute the upgrade for each component. If there are 
+either `EDGE` or `HOST` groups disabled from the upgrade, the corresponding component will be paused after enabled 
+groups get upgraded. At the same time, `MP` upgrade won't be processed, because that requires `EDGE` and `HOST` to be 
+fully upgraded. Refer to the exposed attribute for current upgrade state details. For example, check `upgrade_plan`
+for current upgrade plan, which also includes the plan not specified in this resource and `state` for upgrade status of 
+each component and UpgradeUnitGroups in the component. For more details, please check NSX admin guide.
```
salv-orlando, line 79: I'm not sure I follow line 77-79, are these instructions for operations to perform on NSX?

website/docs/guides/upgrade.html.markdown
```diff
@@ -0,0 +1,158 @@
...
+
+data "nsxt_upgrade_postcheck" "pc2" {
+  upgrade_run_id = nsxt_upgrade_run.run.id
+  type           = "HOST"
+}
+
+check "edge_post_check" {
+  assert {
+    condition     = data.nsxt_upgrade_postcheck.pc1.failed_group == null
+    error_message = "test"
```
salv-orlando, line 148: super pedant nit: can we have a better error message - both above and at line 155.

website/docs/guides/upgrade.html.markdown
ksamoray: Wouldn't it become cumbersome if each HCL example will include all the proceeding DS/resources?

website/docs/guides/upgrade.html.markdown
ksamoray: I couldn't find such document... Maybe @tvigneron could help?

website/docs/guides/upgrade.html.markdown
ksamoray: It's just a copy-paste error, but maybe I'll reuse it as you suggested.

website/docs/guides/upgrade.html.markdown
```diff
@@ -0,0 +1,159 @@
...
+
+The provider support consists of several resources and data sources which can perform a complete upgrade cycle.
+The NSX Terraform provider support upgrades for NSX on ESX hosts, NSX Edge appliances, and NSX manager appliances.
+
+
+## Upgrade process steps
+
+### Preparation for upgrade
+
+Upgrade preparation consists of several actions, and is performed using [nsxt_upgrade_prepare](../resources/upgrade_prepare.html.markdown):
```
ksamoray, line 20: For consistency, either link other resources/DS discussed or remove link here.

website/docs/guides/upgrade.html.markdown
```diff
@@ -0,0 +1,159 @@
...
+  VMware NSX Terraform Provider NSX upgrade support
+---
+
+# NSX Provider for NSX Upgrade
+
+NSX Terraform Provider offers support for [upgrading NSX components](https://docs.vmware.com/en/VMware-NSX/4.1/upgrade/GUID-E04242D7-EF09-4601-8906-3FA77FBB06BD.html).
+
+The provider support consists of several resources and data sources which can perform a complete upgrade cycle.
+The NSX Terraform provider support upgrades for NSX on ESX hosts, NSX Edge appliances, and NSX manager appliances.
+
```
annakhm, line 14: Lets emphasize that upgrade configuration and state should be dedicated to upgrade only, and not shared with regular terraform config and state
Also do we want to recommend that state is cleared post successful upgrade?

website/docs/guides/upgrade.html.markdown
annakhm: I don't think we should support multiple `upgrade` or `prepare` resource instances, this would complicate the logic

website/docs/guides/upgrade.html.markdown
```diff
@@ -0,0 +1,159 @@
...
+### Running the upgrade
+
+In order to configure and execute upgrade of NSXT edges, hosts, and managers, use `nsxt_upgrade_run` resource. 
+
+This resource will start with configuration of upgrade unit groups and upgrade plan settings for `EDGE`and `HOST` 
+components. Host component groups may be created within this resource with custom host lists and attributes, or it can 
+use the predefined groups which NSX provides.
+
+Following the configuration of the edge, host groups, it will execute the upgrade for each component. If there are 
+either `EDGE` or `HOST` groups excluded from the upgrade, the corresponding component will be paused after enabled
```
annakhm, line 75: We should mention that order of groups in config defines the order of upgrade

website/docs/guides/upgrade.html.markdown
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  host_group {
+    id       = data.nsxt_host_upgrade_group.hg1.id
+    parallel = true
+  }
+  host_group {
+    display_name = "HostWithDifferentSettings"
+    parallel     = false
+    hosts        = ["375c260d-3196-487c-b5eb-046de958f6af"]
```
annakhm, line 106: I wonder if we should add host data source for this..

website/docs/guides/upgrade.html.markdown
ksamoray: Those are detailed in the resource doc (which is linked above). I think that this doc should describe mostly the process and for the resource details it references the relevant documentation.

website/docs/guides/upgrade.html.markdown
ksamoray: TBH I've borrowed those from the upgrade_run resource [doc](https://github.com/vmware/terraform-provider-nsxt/blob/master/website/docs/r/upgrade_run.html.markdown) :)
The information described is available both via UI and when setting verbose logging for the provider execution - but it's better if I ask @GraysonWu what was his intention.

website/docs/guides/upgrade.html.markdown
ksamoray: Done - I added a note at the beginning of the doc, about state management for upgrade.

website/docs/guides/upgrade.html.markdown
ksamoray: You mean, within the example? As we do have a host data source (transport_node, and host_transport_node).
Or do you mean a list-providing DS, e.g. `nsxt_host_transport_nodes`?

website/docs/guides/upgrade.html.markdown
salv-orlando: Agreed, let's not make things too difficult for us

website/docs/guides/upgrade.html.markdown
salv-orlando: As discussed elsewhere, we will figure out a different solution for this.

website/docs/guides/upgrade.html.markdown
salv-orlando: Sounds good!

website/docs/guides/upgrade.html.markdown
annakhm: No, I think single one is enough. If we can use the existing DS for the transport node here, I think its preferrable

website/docs/guides/upgrade.html.markdown
```diff
@@ -0,0 +1,185 @@
...
+```
+
+**NOTE:** Acceptance of the license agreement is required to initiate the upgrade process.
+**NOTE:** Precheck bundle is usable only while upgrading a NSX manager of version v4.1.1 and above.
+
+### Checking the readiness for upgrade
+
+When preparation is complete, it is required to validate the readiness for upgrade using the 
+[nsxt_upgrade_prepare_ready](../data-sources/upgrade_prepare_ready.html.markdown) data source.
+When NSX preparation results with failures or warnings, the data source will result will fail the upgrade.
```
annakhm, line 47: `will result will fail` => `will fail`

annakhm (approved):

PR #1231, opened 2024-06-13
VPC Groups resource implementation

ksamoray:


nsxt/resource_nsxt_vpc_group_test.go
```diff
@@ -0,0 +1,166 @@
...
+					resource.TestCheckResourceAttr(testResourceName, "display_name", name),
+					resource.TestCheckResourceAttr(testResourceName, "description", "Acceptance Test"),
+					resource.TestCheckResourceAttrSet(testResourceName, "path"),
+					resource.TestCheckResourceAttrSet(testResourceName, "revision"),
+					resource.TestCheckResourceAttr(testResourceName, "conjunction.#", "0"),
+					resource.TestCheckResourceAttr(testResourceName, "tag.#", "0"),
+					resource.TestCheckResourceAttr(testResourceName, "criteria.#", "1"),
+					resource.TestCheckResourceAttr(testResourceName, "criteria.0.ipaddress_expression.#", "1"),
+					resource.TestCheckResourceAttr(testResourceName, "criteria.0.ipaddress_expression.0.ip_addresses.#", "1"),
+					resource.TestCheckResourceAttr(testResourceName, "criteria.1.macaddress_expression.#", "0"),
```
salv-orlando, line 80: Can you add to the test more criteria, such as those on subnet and subnet ports?

website/docs/r/vpc_group.html.markdown
```diff
@@ -0,0 +1,5 @@
+---
+subcategory: "VPC"
+layout: "nsxt"
+page_title: "NSXT: nsxt_vpc_group"
+description: A resource to configure a VPC Group and its members.
```
salv-orlando, line 5: How should it work here? Do we need to link to the policy_group resource page?

ksamoray: That requires completion

website/docs/r/vpc_security_policy.html.markdown
```diff
@@ -0,0 +1,131 @@
...
+
+An existing security policy can be [imported][docs-import] into this resource, via the following command:
+
+[docs-import]: https://www.terraform.io/cli/import
+
+```
+terraform import nsxt_vpc_security_policy.policy1 PATH
+```
+
+The above command imports the VPC security policy named `policy1` under NSX domain `domain` with the NSX Policy path `PATH`.
```
annakhm, line 131: `domain` is irrelevant he` is irrelevare

nsxt/resource_nsxt_vpc_group_test.go
```diff
@@ -0,0 +1,166 @@
...
+		},
+		Providers: testAccProviders,
+		CheckDestroy: func(state *terraform.State) error {
+			return testAccNsxtPolicyGroupCheckDestroy(state, name, defaultDomain)
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: testAccNsxtVPCGroupAddressCreateTemplate(name, resourceName, true),
+				Check: resource.ComposeTestCheckFunc(
+					testAccNsxtPolicyGroupExists(testResourceName, defaultDomain),
```
annakhm, line 57: I'd suggest to add context validation to one of the tests

website/docs/r/vpc_group.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `display_name` - (Required) Display name of the resource.
+* `description` - (Optional) Description of the resource.
+* `tag` - (Optional) A list of scope + tag pairs to associate with this Group.
+* `nsx_id` - (Optional) The NSX ID of this resource. If set, this ID will be used to create the group resource.
+* `context` - (Optional) The context which the object belongs to
```
annakhm, line 61: This is marked as `Required` in vpc security policy docs.. This will change when provider supports the context clause, just a reminder to go over the docs and change appropriately if needed

annakhm:
LGTM, just some minor comments

website/docs/r/vpc_group.html.markdown
ksamoray: As long as PR #1226 isn't merged, it's better if we keep things consistent with the code. I'll change that.

nsxt/resource_nsxt_vpc_group.go
```diff
@@ -0,0 +1,36 @@
...
+import "github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+func resourceNsxtVPCGroup() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNsxtVPCGroupCreate,
+		Read:   resourceNsxtVPCGroupRead,
+		Update: resourceNsxtVPCGroupUpdate,
+		Delete: resourceNsxtVPCGroupDelete,
+		Importer: &schema.ResourceImporter{
+			State: nsxtPolicyPathResourceImporter,
```
annakhm, line 15: This comment is not specific to group, but rather to all vpc resources - it seems to me that we need VPC-specific Importer that would verify that the path is VPC path

ksamoray: You mean, call `nsxtPolicyPathResourceImporterHelper()` and fail if no `project_id` or `vpc_id` are specified? Yeah that could be helpful.

annakhm: yes, a  wrapper for `nsxtPolicyPathResourceImporter` that also verifies we have a VPC path. I don't think we should validate the object ids against what we might have and not have in config.
maybe its a good idea to create a generic importer that verifies specified path structure, and vpc would be just a use case for that.

annakhm (approved):

PR #1232, opened 2024-06-13
Add validators and improve validator testing

annakhm:
The new validator will serve auto-generated resources based on formats specified in NSX spec


ksamoray (approved):
LGTM

PR #1233, opened 2024-06-18
Resource sharing support

ksamoray:
Add resources to support sharing of resources between projects.


nsxt/resource_nsxt_policy_shared_resource.go
```diff
@@ -0,0 +1,207 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Share ID to associate the resource to",
+				ForceNew:    true,
+			},
+			"resource_object": {
+				Type:        schema.TypeList,
+				Description: "List of resources to be shared",
+				Required:    true,
+				MinItems:    1,
```
salv-orlando, line 42: I think we usually let NSX validate the number of items, but in this case I agree with doing it in the provider.
It is clearly pointless to create a shared resource without any resource :/

nsxt/resource_nsxt_policy_shared_resource.go
```diff
@@ -0,0 +1,207 @@
...
+			"revision":     getRevisionSchema(),
+			"tag":          getTagsSchema(),
+			"context":      getContextSchema(false, false, false),
+			"share_id": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Share ID to associate the resource to",
+				ForceNew:    true,
+			},
+			"resource_object": {
```
salv-orlando, line 38: This is plural in the API (resource_objects). Can we keep it plural here as well?

nsxt/resource_nsxt_policy_shared_resource.go
```diff
@@ -0,0 +1,207 @@
...
+	return &schema.Resource{
+		Create: resourceNsxtPolicySharedResourceCreate,
+		Read:   resourceNsxtPolicySharedResourceRead,
+		Update: resourceNsxtPolicySharedResourceUpdate,
+		Delete: resourceNsxtPolicySharedResourceDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceNsxtPolicySharedResourceImporter,
+		},
+
+		Schema: map[string]*schema.Schema{
```
salv-orlando, line 24: I see a minor but non-null value in adding `sharing-shared-with` as a computed attribute. What do you reckon?

nsxt/resource_nsxt_policy_shared_resource.go
ksamoray: We always use singular notation for a list of complex object like in this case, and a plural notation for lists of simple objects - this is due to the way that TF represents them. e.g.
```
   ip_addresses = ["1.1.1.1", "2.2.2.2"]
```

.vs.
```
   some_cidr_obj {
     addr = "1.1.1.0"
     prefix_len = "24"
  }
  some_cidr_obj {
     addr = "2.1.1.0"
     prefix_len = "24"
  }
```

nsxt/resource_nsxt_policy_shared_resource.go
ksamoray: Yeah it could be useful. I'll add ir.

ksamoray: So I've tried to add this - but NSX returns nothing in this attr. I'll remove it then.

website/docs/r/policy_share.html.markdown
```diff
@@ -0,0 +1,58 @@
...
+* `shared_with` - (Required) Path of the context
+* `sharing_strategy` - (Optional) Sharing Strategy. Accepted values: 'NONE_DESCENDANTS', 'ALL_DESCENDANTS'. Default: 'NONE_DESCENDANTS'.
+
+## Attributes Reference
+
+In addition to arguments listed above, the following attributes are exported:
+
+* `id` - ID of the resource.
+* `revision` - Indicates current revision number of the object as seen by NSX-T API server. This attribute can be useful for debugging.
+* `path` - The NSX path of the policy resource.
```
annakhm, line 46: If share has a path, would it make sense to use `share_path` rather than `share_id` in `nsxt_policy_shared_resource`? Can `path` be more specific than `id` for the share?

ksamoray: That parameter is passed into the SDK API as an id, so in the end even if we pass a path it will be parsed out into an id :)
But I'll change it for consistency as path is used for the shared objects.

salv-orlando:
@annakhm can you take a quick look before the PR is submitted?

annakhm:
LGTM, the only concern I have is usability. I feel it might be a better user experience to configure sharing as part of the resource itself. But that comes with its own set of problems.

annakhm (approved):

PR #1234, opened 2024-06-19
Add helper for future vpc resources

annakhm:
This helper will be used in autogenerated vpc resources that don't have explicit parent path in schema


ksamoray:
I'd put this in [utl_file_template.yaml](https://github.com/vmware/terraform-provider-nsxt/blob/master/api/utl_file_template.yaml) which will place this [here](https://github.com/vmware/terraform-provider-nsxt/blob/master/api/utl/api_util.go) after `make api-wrapper`.
Capitalize the func name, so it will be accessible (and while so, also the linter won't complain about..)

annakhm:
> I'd put this in [utl_file_template.yaml](https://github.com/vmware/terraform-provider-nsxt/blob/master/api/utl_file_template.yaml) which will place this [here](https://github.com/vmware/terraform-provider-nsxt/blob/master/api/utl/api_util.go) after `make api-wrapper`. Capitalize the func name, so it will be accessible (and while so, also the linter won't complain about..)

That's a good idea, but I prefer to wait with it until we have a single home place for the wrapper generator and templates

ksamoray (approved):

PR #1235, opened 2024-06-20
Add VPC path importer func

ksamoray:
VPC resources use policy path to import resources into the TF state. This importer func does some validation that the required attributes for a policy resource import do exist.

### DIFF
--- a/nsxt/policy_utils.go
+++ b/nsxt/policy_utils.go
@@ -382,6 +382,18 @@ func nsxtPolicyPathResourceImporter(d *schema.ResourceData, m interface{}) ([]*s
 	return rd, nil
 }
 
+func nsxtVPCPathResourceImporter(d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
+	rd, err := nsxtPolicyPathResourceImporterHelper(d, m)
+	if err != nil {
+		return rd, err
+	}
+	projectID, vpcID := getContextDataFromSchema(d)
+	if projectID == "" || vpcID == "" {
+		return rd, fmt.Errorf("imported resource policy path should have both project_id and vpc_id fields")
+	}
+	return rd, nil
+}
+
 func nsxtPolicyPathResourceImporterHelper(d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
 	importID := d.Id()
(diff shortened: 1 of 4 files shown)


nsxt/policy_utils.go
```diff
@@ -382,6 +382,20 @@ func nsxtPolicyPathResourceImporter(d *schema.ResourceData, m interface{}) ([]*s
 	return rd, nil
 }
 
+func nsxtVPCPathResourceImporter(d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
+	rd, err := nsxtPolicyPathResourceImporterHelper(d, m)
+	if errors.Is(err, ErrNotAPolicyPath) {
+		return rd, nil
```
annakhm, line 388: I see this behavior in other importer as well - could you please remind me why we swallow `ErrNotAPolicyPath` error?

ksamoray: In importers which support importing via both `path` and `id`, it's used to handle the `id` case. In this case it's a bug, I'll fix this.

nsxt/policy_utils.go
```diff
@@ -382,6 +382,18 @@ func nsxtPolicyPathResourceImporter(d *schema.ResourceData, m interface{}) ([]*s
 	return rd, nil
 }
 
+func nsxtVPCPathResourceImporter(d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
+	rd, err := nsxtPolicyPathResourceImporterHelper(d, m)
+	if err != nil {
+		return rd, err
+	}
+	projectID, vpcID := getContextDataFromSchema(d)
```
salv-orlando, line 390: don't we need something like this for all project-scoped resources?

salv-orlando (approved):
LGTM

nsxt/policy_utils.go
ksamoray: If you refer to the importer, then it's more complicated with project/multitenancy resources, as we use the same resource in both project and non-project contexts while VPC resources are VPC-only.
Also as with legacy we used to import resources by id (instead of by path as we currently do), so that import-id could be an NSX id and not a path. We don't have this case with VPC either.

annakhm (approved):

PR #1238, opened 2024-06-25
Add OmitIfEmpty option in metadata library

annakhm:
NSX would not accept empty value for some attributes. For those, the attribute should  be omitted in request. This PR add such support this behavior when flag is specified in metadata.


annakhm:
> I guess this only applies to strings? And from stuct to schema this will be automatically bypassed since the stuct would have a nil value?

Correct, I don't think this can be relevant to numbers and bools

wsquan171 (approved):

PR #1239, opened 2024-06-26
Support vpc nat resource and its friends

annakhm:


annakhm:
Note that `ip_address_allocation` resource will be regenerated with new spec. Hence I didn't add detailed tests for it - basic testing is already covered by the nat resource test.

annakhm:
Reposting the netto support for those objects with https://github.com/vmware/terraform-provider-nsxt/pull/1247

PR #1241, opened 2024-06-27
Ignore display name for policy_ldap_identity_source when NSX version >= 4.2.1

2ez4szliu:
Fix an issue with NSX version >= 4.2.1 that ldap_identity_source without display_name cannot be successfully created because NSX has forced the display_name to match its ID.

### DIFF
--- a/nsxt/resource_nsxt_policy_ldap_identity_source.go
+++ b/nsxt/resource_nsxt_policy_ldap_identity_source.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/vmware/terraform-provider-nsxt/nsxt/util"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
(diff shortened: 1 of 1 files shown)


nsxt/resource_nsxt_policy_ldap_identity_source.go
```diff
@@ -219,6 +220,9 @@ func resourceNsxtPolicyLdapIdentitySourceProbeAndUpdate(d *schema.ResourceData,
 			LdapServers:            ldapServers,
 			ResourceType:           nsxModel.LdapIdentitySource_RESOURCE_TYPE_ACTIVEDIRECTORYIDENTITYSOURCE,
 		}
+		if util.NsxVersionLower("4.2.1") {
```
annakhm, line 223: This does not seem to change behavior because DisplayName is still assigned in line 212

annakhm:
Hi @2ez4szliu, can you please explain in more detail what NSX behavior we're trying to work around here?

nsxt/resource_nsxt_policy_ldap_identity_source.go
2ez4szliu: Thanks, it's fixed

2ez4szliu:
> Hi @2ez4szliu, can you please explain in more detail what NSX behavior we're trying to work around here?

Hi @annakhm, the issue is 3403819, if I understand correctly, before 421 we can specify arbitrary display_name for ldap_identity_source, but since 421 it has to match a specific value. As @salv-orlando suggested we can just ignore parsing display name on provider side, in that case the NSX resource created will use the correct name by default

ksamoray:
As 4.2.1 isn't released yet, can't it be resolved within the platform?

annakhm:
I think its preferrable to remove display name from the resource (it is still in Beta)

annakhm:
Suggested alternative https://github.com/vmware/terraform-provider-nsxt/pull/1254

salv-orlando:
Closing this PR as it is superseded by PR #1254.
Thanks a lot @2ez4szliu for your contribution!

PR #1242, opened 2024-07-01
[WIP] Skip attribute processing when no values is assigned in schema

ksamoray:
When a value is not assigned to the schema for some attribute, using  d.Get("attr_name").(type) will return a zero or a null string.

That would break when this value is not allowed by NSX and usually we protect against this with an if statement which would assign this attribute a value only when it's set in the schema. With required attributes that cannot happen as NSX will check that a value is assigned prior to the Create() call.

### DIFF
--- a/nsxt/metadata/metadata.go
+++ b/nsxt/metadata/metadata.go
@@ -306,7 +306,12 @@ func SchemaToStruct(elem reflect.Value, d *schema.ResourceData, metadata map[str
 			if len(parent) > 0 {
 				value = parentMap[key].(string)
 			} else {
-				value = d.Get(key).(string)
+				v, exists := d.GetOk(key)
+				if !exists {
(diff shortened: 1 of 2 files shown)


annakhm:
We need to be able to set empty value as in some cases, this is the way to nullify value on update.
We have `OmitIfEmpty` boolean flag to control this behavior

ksamoray:
@annakhm Isn't it better to nullify by default and set when needed?

annakhm:
> @annakhm Isn't it better to nullify by default and set when needed?

There is a difference between empty value (translates to field name with empty value in API call) and nil value (field name is not mentioned in the request). The latter is sometimes ignored with NSX `Patch` behavior (which is correct behavior, since attributes that are not specified should not be touched with `Patch`. In those cases, explicit empty value is needed to remove assigned value.

PR #1246, opened 2024-07-03
VPC subnet resource implmentation

ksamoray:
PR contains two commits - the first has the autogenerated files and the second, all the manual edits required to make it work.
As the generator is still being debugged, I'd rather keep it so - so I can regenerate whenever something is fixed in the generator (at least of as long as this PR isn't merged).

### DIFF
--- a/nsxt/resource_nsxt_vpc_subnet.go
+++ b/nsxt/resource_nsxt_vpc_subnet.go
@@ -1,0 +1,350 @@
+/* Copyright © 2024 Broadcom, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"log"
+	"reflect"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+	clientLayer "github.com/vmware/vsphere-automation-sdk-go/services/nsxt/orgs/projects/vpcs"
+
+	utl "github.com/vmware/terraform-provider-nsxt/api/utl"
+	"github.com/vmware/terraform-provider-nsxt/nsxt/metadata"
+)
+
+var vpcSubnetAccessModeValues = []string{
+	model.VpcSubnet_ACCESS_MODE_PRIVATE,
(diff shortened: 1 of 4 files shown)


annakhm:
LGTM but I think we need more test coverage - with `advanced_config`, `dhcp_config` and alternating presence of those

salv-orlando:
@ksamoray can you update the target branch for this PR to vpc-feature-branch

nsxt/resource_nsxt_vpc_subnet.go
```diff
@@ -209,13 +213,13 @@ func resourceNsxtVpcSubnet() *schema.Resource {
 		Update: resourceNsxtVpcSubnetUpdate,
 		Delete: resourceNsxtVpcSubnetDelete,
 		Importer: &schema.ResourceImporter{
-			State: nsxtParentPathResourceImporter,
+			State: nsxtVPCPathResourceImporter,
```
salv-orlando, line 216: If this has already been fixed in the generator maybe we can try and update the generated commit so we don't have to manually change the importer?

ksamoray: If I understand correctly, `nsxtParentPathResourceImporter` is useful for objects which have sub-objects of a level VPC obj, e.g `VpcSubnetPort` which is nested under `VpcSubnet` and therefore has  a parent object path.
IDK if we can conclude automatically from the spec if this is the case or not. @annakhm do you know?

nsxt/resource_nsxt_vpc_subnet.go
```diff
@@ -31,7 +31,7 @@ var vpcSubnetSchema = map[string]*metadata.ExtendedSchema{
 	"description":  metadata.GetExtendedSchema(getDescriptionSchema()),
 	"revision":     metadata.GetExtendedSchema(getRevisionSchema()),
 	"tag":          metadata.GetExtendedSchema(getTagsSchema()),
-	"context":      metadata.GetExtendedSchema(getContextSchema(false, false, true)),
+	"context":      metadata.GetExtendedSchema(getContextSchema(true, false, true)),
```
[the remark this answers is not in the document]
ksamoray: We could IMO, VPC object contexts are always required. IDK if the generator distinguishes VPC objects which require `context` attribute and those which require `parent_path`. @annakhm is this doable?

nsxt/resource_nsxt_vpc_subnet.go
annakhm: This fix is already present in generator - based on whether `parent_path` is specified or not, the correct importer should be chosen

annakhm (approved):

PR #1248, opened 2024-07-04
Handle OmitIfEmpty for bool and int types

ksamoray:
Instead of passing a false/zero value, add some tests for OmitIfEmpty case

### DIFF
--- a/nsxt/metadata/metadata.go
+++ b/nsxt/metadata/metadata.go
@@ -303,12 +303,18 @@ func SchemaToStruct(elem reflect.Value, d *schema.ResourceData, metadata map[str
 		}
 		if item.Metadata.SchemaType == "string" {
 			var value string
-			if len(parent) > 0 {
+			exists := false
+			if len(parent) > 0 && parentMap[key] != nil {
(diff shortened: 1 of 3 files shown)


nsxt/metadata/metadata.go
```diff
@@ -322,6 +322,10 @@ func SchemaToStruct(elem reflect.Value, d *schema.ResourceData, metadata map[str
 			} else {
 				value = d.Get(key).(bool)
 			}
+			if item.Metadata.OmitIfEmpty && !value {
```
annakhm, line 325: wouldn't this be true if the value of bool is specified as `false`?

ksamoray: When bool is unspecified it defaults to `false`, it's also evident [here](https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/metadata/metadata_test.go#L630) in the tests.
Nonetheless I've revised this PR to use `GetOk()` and handle cases where an attribute is omitted within a sub-object - before it could crash the provider IMO.

annakhm (approved):

PR #1254, opened 2024-07-11
Remove display_name from ldap resource

annakhm:
Since starting from NSX 4.2, display_name has to be identical to id, display_name attribute is redundant


ksamoray:
I think that calling `getOrGenerateID()` is not required anymore in `resourceNsxtPolicyLdapIdentitySourceCreate()` as nsx_id is required (although the code will work as is as the id will never be generated).

annakhm:
> I think that calling `getOrGenerateID()` is not required anymore in `resourceNsxtPolicyLdapIdentitySourceCreate()` as nsx_id is required (although the code will work as is as the id will never be generated).

You're right, however the extra `if` inside `getOrGenerateID` does no harm, and to bypass it I would need to replace this single line of code with 4, so I would leave it as is

ksamoray (approved):

PR #1255, opened 2024-07-11
VPC resource

annakhm:


nsxt/resource_nsxt_vpc.go
```diff
@@ -0,0 +1,626 @@
...
+
+var vpcSchema = map[string]*metadata.ExtendedSchema{
+	"nsx_id":       metadata.GetExtendedSchema(getNsxIDSchema()),
+	"path":         metadata.GetExtendedSchema(getPathSchema()),
+	"display_name": metadata.GetExtendedSchema(getDisplayNameSchema()),
+	"description":  metadata.GetExtendedSchema(getDescriptionSchema()),
+	"revision":     metadata.GetExtendedSchema(getRevisionSchema()),
+	"tag":          metadata.GetExtendedSchema(getTagsSchema()),
+	"context":      metadata.GetExtendedSchema(getContextSchema(true, false, false)),
+	"external_ipv4_blocks": {
```
ksamoray, line 33: Can validate as policy path

nsxt/resource_nsxt_vpc.go
```diff
@@ -0,0 +1,626 @@
...
+			Type:         schema.TypeString,
+			ValidateFunc: validatePolicyPath(),
+			Optional:     true,
+		},
+		Metadata: metadata.Metadata{
+			SchemaType:   "string",
+			SdkFieldName: "VpcServiceProfile",
+		},
+	},
+	"load_balancer_vpc_endpoint": {
```
ksamoray, line 166: Can we flatten this? In the end we have MaxItem=1 and a boolean within.

nsxt/resource_nsxt_vpc.go
```diff
@@ -0,0 +1,626 @@
...
+			},
+			Optional: true,
+		},
+		Metadata: metadata.Metadata{
+			SchemaType:   "struct",
+			SdkFieldName: "LoadBalancerVpcEndpoint",
+			ReflectType:  reflect.TypeOf(model.LoadBalancerVPCEndpoint{}),
+		},
+	},
+	"ip_address_type": {
```
ksamoray, line 192: We could drop this entirely, it has only `IPV4` as valid value

nsxt/resource_nsxt_vpc.go
```diff
@@ -0,0 +1,626 @@
...
+		Metadata: metadata.Metadata{
+			SchemaType:   "struct",
+			SdkFieldName: "SubnetProfiles",
+			ReflectType:  reflect.TypeOf(model.SubnetProfiles{}),
+		},
+	},
+	"private_ipv4_blocks": {
+		Schema: schema.Schema{
+			Type: schema.TypeList,
+			Elem: &schema.Schema{
```
ksamoray, line 315: We can validate this as policy path

nsxt/resource_nsxt_vpc.go
```diff
@@ -0,0 +1,626 @@
...
+				Type: schema.TypeString,
+			},
+			Optional: true,
+		},
+		Metadata: metadata.Metadata{
+			SchemaType:   "array",
+			SdkFieldName: "PrivateIpv4Blocks",
+		},
+	},
+	"short_id": {
```
ksamoray, line 325: This is also computed as NSX sets value to it if user doesn't

annakhm: I wonder if this one is also impossible to update..

ksamoray: NSX v4.1.2 UI protects it from updating. So I guess it's the same.

nsxt/resource_nsxt_vpc.go
annakhm: I think we can keep it here, when `ipv6` the enum list will grow automatically as result of resource evolution, and then no manual work will be required
however we might want to skip this in docs

nsxt/resource_nsxt_vpc_test.go
```diff
@@ -0,0 +1,290 @@
...
+	"listener_ip":                "230.64.2.1",
+	"enable_dhcp":                "false",
+	"dns_server_ips":             "20.4.12.8",
+}
+
+func TestAccResourceNsxtVpc_basic(t *testing.T) {
+	testResourceName := "nsxt_vpc.test"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
```
ksamoray, line 52: This should be tested for VPC only

nsxt/resource_nsxt_vpc.go
annakhm: for now the strategy is to follow the api..

annakhm:
waiting for #1286 to rebase and fix the tests

nsxt/resource_nsxt_vpc.go
annakhm: This seems to be computed only

nsxt/data_source_nsxt_policy_project.go
```diff
@@ -33,7 +33,14 @@ func dataSourceNsxtPolicyProject() *schema.Resource {
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
 				},
-				Optional: true,
+				Computed: true,
+			},
+			"external_ipv4_blocks": {
```
ksamoray, line 38: Change to project data source should be documented

PR #1256, opened 2024-07-14
Handle upgrade unit groups which are lost in reset

ksamoray:
Some host upgrade groups, e.g the one for non-clustered ESXi hosts change their ID attribute during reset operation. As this causes the upgrade to fail when they're referenced by the run resource, cache the upgrade group list before calling Reset() and lookup the new ID of the group which replaces the old one.

### DIFF
--- a/nsxt/resource_nsxt_upgrade_run.go
+++ b/nsxt/resource_nsxt_upgrade_run.go
@@ -11,18 +11,12 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
-	"github.com/vmware/vsphere-automation-sdk-go/lib/vapi/std/errors"
-	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx"
-	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/fabric"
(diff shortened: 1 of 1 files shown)


nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -564,48 +561,23 @@ func updateUpgradeUnitGroups(upgradeClientSet *upgradeClientSet, d *schema.Resou
...
-				nsxUnits := getUnitIDsFromUnits(groupGet.UpgradeUnits)
-
-				// Find and remove upgrade units which have been removed from the list. This is done by assigning the
-				// upgrade unit to its default group
-				var schemaUnits []string
-				if group["hosts"] != nil {
-					schemaUnits = interface2StringList(group["hosts"].([]interface{}))
-				}
-				getHostDefaultUpgradeGroup, err := getHostDefaultUpgradeGroupGetter(m, groupList)
+				return fmt.Errorf("upgrade unit group with name %s already exists", groupName)
```
salv-orlando, line 564: A generic concern: why are we now failing in this case, and why don't we need anymore logic for handling the case where custom group could have been updated?
It's likely simply because I do not fully understand the logic here.

salv-orlando: Got it, thanks!

nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -520,26 +521,22 @@ func updateUpgradeUnitGroups(upgradeClientSet *upgradeClientSet, d *schema.Resou
...
-	groupNameMap := make(map[string]bool)
 
 	preUpgradeGroupID := ""
 	for _, groupI := range d.Get(componentToGroupKey[component]).([]interface{}) {
 		group := groupI.(map[string]interface{})
 		groupID := group["id"].(string)
 		var groupGet *model.UpgradeUnitGroup
-		var err error
 		isCreate := false
 		if groupID == "" {
```
annakhm, line 535: I would suggest to add a comment explaining what is the case for groups with ids and groups without

annakhm (approved):

PR #1262, opened 2024-07-18
Revise VPC subnet resource with latest generator

ksamoray:
Regenerated with args:
VpcSubnet --client-name Subnet --vpc-only --no-api-wrapper

and reapplied manual changes over patch.


annakhm (approved):

annakhm:
We don't have validations here for `ip_addresses` and `dns_server_ips` - is this because spec does not specify any or spec doesn't match behavior?

annakhm (approved):

PR #1280, opened 2024-07-24
VPC 2.0 Service Profile support

annakhm:
Extended tests will be added as follow up


annakhm:
reposting as https://github.com/vmware/terraform-provider-nsxt/pull/1286

PR #1285, opened 2024-07-27
Add new member types to group criteria

annakhm:


ksamoray (commented):
Should we validate that those are only used for VPC?

annakhm:
> Should we validate that those are only used for VPC?

I would follow the API for this one, given that API spec chose not to redefine this list for VPC group.
In future we hope to auto-generate those list, and adding manual validations here would remove the benefit.

ksamoray (approved):

PR #1286, opened 2024-07-28
Vpc20 service profile support

annakhm:
Extended tests will be added as follow up

### DIFF
--- a/nsxt/resource_nsxt_vpc_service_profile.go
+++ b/nsxt/resource_nsxt_vpc_service_profile.go
@@ -1,0 +1,450 @@
+/* Copyright © 2024 Broadcom, Inc. All Rights Reserved.
+   SPDX-License-Identifier: MPL-2.0 */
+
+package nsxt
+
+import (
+	"fmt"
+	"log"
+	"reflect"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+	clientLayer "github.com/vmware/vsphere-automation-sdk-go/services/nsxt/orgs/projects"
+
+	utl "github.com/vmware/terraform-provider-nsxt/api/utl"
+	"github.com/vmware/terraform-provider-nsxt/nsxt/metadata"
+)
+
+var vpcServiceProfileModeValues = []string{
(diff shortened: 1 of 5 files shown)


nsxt/resource_nsxt_vpc_service_profile.go
```diff
@@ -0,0 +1,445 @@
...
+		Schema: schema.Schema{
+			Type:     schema.TypeBool,
+			Optional: true,
+		},
+		Metadata: metadata.Metadata{
+			SchemaType:   "bool",
+			SdkFieldName: "IsDefault",
+		},
+	},
+	"mac_discovery_profile": {
```
ksamoray, line 54: Various profiles are policy paths? If so, they can be validated with `validatePolicyPath()`

nsxt/resource_nsxt_vpc_service_profile.go
```diff
@@ -0,0 +1,445 @@
...
+		},
+		Schema: metadata.GetSchemaFromExtendedSchema(vpcServiceProfileSchema),
+	}
+}
+
+func resourceNsxtVpcServiceProfileExists(sessionContext utl.SessionContext, id string, connector client.Connector) (bool, error) {
+	var err error
+	parents := getVpcParentsFromContext(sessionContext)
+	client := clientLayer.NewVpcServiceProfilesClient(connector)
+	_, err = client.Get(parents[0], parents[1], id)
```
ksamoray, line 317: It is simpler to take org/project/VPC from context than loading these into parents array and using from there

nsxt/resource_nsxt_vpc_service_profile_test.go
```diff
@@ -0,0 +1,226 @@
...
+	"server_addresses": "11.11.11.111",
+	"cache_size":       "1024",
+	"log_level":        "INFO",
+}
+
+func TestAccResourceNsxtVpcServiceProfile_basic(t *testing.T) {
+	testResourceName := "nsxt_vpc_service_profile.test"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t); testAccOnlyMultitenancy(t) },
```
ksamoray, line 42: TBH I'm not sure if that should be tested with multitenancy or with VPC env. For now definitely with VPC as it runs on v9.0 (these tests will fail with v4.1.2).

annakhm: I think this should be part of VPC test suite later on too

nsxt/resource_nsxt_vpc_service_profile.go
annakhm: For the generator this way is a cleaner and more readable option I think. The only difference is function name, rather than rewriting the whole block.

ksamoray (approved):

PR #1291, opened 2024-07-30
Remove redundant code

annakhm:


ksamoray (approved):
Where did that came from? The template?

PR #1296, opened 2024-08-01
VPC connectivity profile and transit gateway

annakhm:


nsxt/resource_nsxt_vpc_connectivity_profile_test.go
```diff
@@ -0,0 +1,210 @@
...
+	"display_name":        getAccTestResourceName(),
+	"description":         "terraform created",
+	"enable":              "false",
+	"enable_default_snat": "false",
+}
+
+var accTestVpcConnectivityProfileUpdateAttributes = map[string]string{
+	"display_name":        getAccTestResourceName(),
+	"description":         "terraform updated",
+	"enable":              "false",
```
ksamoray, line 24: Nit: if parameters are same for create and update, maybe it's worthwhile to drop them

ksamoray (approved):
Just minor stuff, otherwise LGTM. We don't really need to fix these

nsxt/resource_nsxt_transit_gateway_test.go
```diff
@@ -0,0 +1,179 @@
...
+	"transit_subnets": "192.168.5.0/24",
+}
+
+var accTestTransitGatewayUpdateAttributes = map[string]string{
+	"display_name":    getAccTestResourceName(),
+	"description":     "terraform updated",
+	"transit_subnets": "192.168.7.0/24",
+}
+
+func TestAccResourceNsxtTransitGateway_basic(t *testing.T) {
```
ksamoray, line 26: Note that vpc20 CI job doesn't run those tests now, as only tests that match `^TestAcc.*V[Pp][Cc].*$` regexp are executed

nsxt/resource_nsxt_vpc_connectivity_profile_test.go
annakhm: I hope those will change once NSX API is fully functional

nsxt/resource_nsxt_transit_gateway_test.go
annakhm: Thanks, good to know. Can we add this specific one to the list? I'm not sure about the naming btw

PR #1304, opened 2024-08-06
Verify that id is actually from the prepare resource

ksamoray:
We added some id attributes in the upgrade resources to enforce dependency, yet we never validate these.

### DIFF
--- a/nsxt/data_source_nsxt_upgrade_prepare_ready.go
+++ b/nsxt/data_source_nsxt_upgrade_prepare_ready.go
@@ -5,6 +5,7 @@ package nsxt
 
 import (
 	"fmt"
+	"github.com/vmware/terraform-provider-nsxt/nsxt/util"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	nsxModel "github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/model"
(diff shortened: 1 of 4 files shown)


nsxt/util/utils.go
```diff
@@ -29,3 +32,23 @@ func NsxVersionHigherOrEqual(ver string) bool {
...
 	return currentVersion.Compare(requestedVersion) >= 0
 }
+
+func hash(s string) uint32 {
+	h := fnv.New32a()
+	h.Write([]byte(s))
+	return h.Sum32()
+}
+
+func GetVerifyableID(id, extra string) string {
```
annakhm, line 42: Do we need the `extra` parameter here? Maybe just randomize id within this function and call it `getRandomVerifiableID`?

nsxt/data_source_nsxt_upgrade_prepare_ready.go
```diff
@@ -80,10 +86,7 @@ func dataSourceNsxtUpgradePrepareReadyRead(d *schema.ResourceData, m interface{}
 	if len(errMessage) > 0 {
 		return fmt.Errorf(errMessage)
 	}
-	objID := d.Get("id").(string)
-	if objID == "" {
-		objID = newUUID()
-	}
+	objID := util.GetVerifyableID(newUUID(), "nsxt_upgrade_prepare_ready")
```
annakhm, line 89: nit: `Verifyable` => `Verifiable`

nsxt/util/utils.go
ksamoray: As I validate these ids already, it makes sense that the id is from the correct resource (so they won't depend on the `nsxt_upgrade_prepare_ready` DS where they should depend on `nsxt_upgrade_prepare` resource etc. This process has many components, it's better to enforce the order as much as possible IMO.

annakhm (approved):

PR #1305, opened 2024-08-07
Protect against empty import ID

annakhm:


ksamoray (approved):

ksamoray:
@annakhm  can we cherry-pick this into master? So we can run the entire test deck against it?

PR #1307, opened 2024-08-08
Set tests which require VPC context, to run in VPC env

ksamoray:
Plus a couple of cherry-picks from master, to address v4.2.0+ test failures


nsxt/resource_nsxt_policy_tier1_gateway.go
```diff
@@ -407,13 +407,16 @@ func policyTier1GatewayResourceToInfraStruct(context utl.SessionContext, d *sche
...
 		ForceWhitelisting:       &forceWhitelisting,
-		Tier0Path:               &tier0Path,
 		RouteAdvertisementTypes: routeAdvertisementTypes,
 		RouteAdvertisementRules: routeAdvertisementRules,
 		Ipv6ProfilePaths:        ipv6ProfilePaths,
 		ResourceType:            &t1Type,
 	}
 
+	if tier0Path != "" {
+		obj.Tier0Path = &tier0Path
```
annakhm, line 417: This can potentially break the scenario when T0 path is removed. This code path is implemented with H-API that might use full `partial patch` logic, which causes nsx to ignore unspecified attributes. We need to run this through test that removes existing `tier0_path` attribute to be sure this change is harmless.

ksamoray: We test this [here](https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/resource_nsxt_policy_tier1_gateway_test.go#L432), I think

ksamoray: Of course we run it - `ci/jenkins/lm-9.0.0` job runs these tests on VCF9. The purpose of this PR is to make this branch pass the tests on that platform :)

annakhm (approved):

PR #1312, opened 2024-08-10
Disallow empty tag{} clause in VPC resources

annakhm:
To avoid exessive merges, non-vpc resources will follow later.


salv-orlando (approved):
LGTM

PR #1314, opened 2024-08-12
Remove is_default flag from vpc service profile

annakhm:
This flag cannot be set to true via API and thus adds no value. In addition, add qos etc. profile tests to the test suite

### DIFF
--- a/nsxt/resource_nsxt_vpc_service_profile.go
+++ b/nsxt/resource_nsxt_vpc_service_profile.go
@@ -41,16 +41,6 @@ var vpcServiceProfileSchema = map[string]*metadata.ExtendedSchema{
 	"revision":     metadata.GetExtendedSchema(getRevisionSchema()),
 	"tag":          metadata.GetExtendedSchema(getTagsSchema()),
 	"context":      metadata.GetExtendedSchema(getContextSchema(true, false, false)),
-	"is_default": {
-		Schema: schema.Schema{
-			Type:     schema.TypeBool,
-			Optional: true,
-		},
-		Metadata: metadata.Metadata{
-			SchemaType:   "bool",
-			SdkFieldName: "IsDefault",
-		},
-	},
 	"mac_discovery_profile": {
 		Schema: schema.Schema{
 			Type:         schema.TypeString,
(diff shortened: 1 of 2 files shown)


nsxt/resource_nsxt_vpc_service_profile.go
```diff
@@ -41,16 +41,6 @@ var vpcServiceProfileSchema = map[string]*metadata.ExtendedSchema{
...
-	"is_default": {
-		Schema: schema.Schema{
-			Type:     schema.TypeBool,
-			Optional: true,
-		},
-		Metadata: metadata.Metadata{
-			SchemaType:   "bool",
-			SdkFieldName: "IsDefault",
-		},
-	},
```
salv-orlando, line 53: I don't think this has any value as a computed attribute, is that correct?
That would be because for any user-created profile this is always going to be False.

nsxt/resource_nsxt_vpc_service_profile_test.go
```diff
@@ -68,7 +68,35 @@ func TestAccResourceNsxtVpcServiceProfile_basic(t *testing.T) {
...
+					resource.TestCheckResourceAttr(testResourceName, "dns_forwarder_config.0.log_level", accTestPolicyVpcServiceProfileCreateAttributes["log_level"]),
+
+					resource.TestCheckResourceAttrSet(testResourceName, "qos_profile"),
+					resource.TestCheckResourceAttrSet(testResourceName, "spoof_guard_profile"),
+					resource.TestCheckResourceAttrSet(testResourceName, "ip_discovery_profile"),
+					resource.TestCheckResourceAttrSet(testResourceName, "mac_discovery_profile"),
+					resource.TestCheckResourceAttrSet(testResourceName, "security_profile"),
+					resource.TestCheckResourceAttrSet(testResourceName, "nsx_id"),
+					resource.TestCheckResourceAttrSet(testResourceName, "path"),
+					resource.TestCheckResourceAttrSet(testResourceName, "revision"),
```
salv-orlando, line 94: How difficult it would be to also check these attributes have the correct value?

nsxt/resource_nsxt_vpc_service_profile.go
annakhm: correct

nsxt/resource_nsxt_vpc_service_profile_test.go
annakhm: We don't know paths for those attributes. We could pull it from NSX or from the dependant resource, but it feels to me like an effort not justified

salv-orlando (approved):
Thanks for the feedback, Anna.
LGTM.

PR #1326, opened 2024-08-24
Policy Group: Check for nil entries in criteria block

salv-orlando:
As every element in the CriteriaSet schema is empty, users migth specify an empty block (ie: 'criteria {}'). This causes the provider to panic as the criteria validation routine tries to cast from nil.

With this change, the provider will reject this configuration as invalid.

### DIFF
--- a/nsxt/resource_nsxt_policy_group.go
+++ b/nsxt/resource_nsxt_policy_group.go
@@ -322,7 +322,7 @@ func getExtendedCriteriaSetSchema() *schema.Resource {
 				Type:        schema.TypeSet,
 				Description: "Identity Group expression",
 				Elem:        getIdentityGroupSchema(),
-				Optional:    true,
+				Required:    true,
 			},
 		},
 	}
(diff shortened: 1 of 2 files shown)


nsxt/resource_nsxt_policy_group.go
```diff
@@ -821,6 +824,7 @@ func getIdentityGroupsData(expressions []*data.StructValue) ([]map[string]interf
 }
 
 func validateGroupCriteriaAndConjunctions(criteriaSets []interface{}, conjunctions []interface{}) ([]criteriaMeta, error) {
+	log.Printf("[INFO] ##### CRITERIA SET: %v, CONJUNCTIONS: %v", len(criteriaSets), len(conjunctions))
```
ksamoray, line 827: Did you plan on leaving this? Or is it a debugging message?

salv-orlando: It's a debugging message. Thankfully I only left this one :)

salv-orlando: fixed

PR #1327, opened 2024-08-26
Fix upgrade run with pause malfunctions

ksamoray:
Upgrade run does not resume when all the groups of previous component have completed successfully


salv-orlando (approved):
LGTM

PR #1334, opened 2024-08-28
Drop policy prefix for VPC-related objects naming

annakhm:
Since MP resources are deprecated, it makes sense to drop policy prefix for new objects, especially those that are VPC related.


ksamoray:
Should be fixed:
[here](https://github.com/vmware/terraform-provider-nsxt/blob/bec90ffaf62e813564a03f32434ec6c077a9d459/nsxt/resource_nsxt_transit_gateway_attachment_test.go#L157)
[here](https://github.com/vmware/terraform-provider-nsxt/blob/bec90ffaf62e813564a03f32434ec6c077a9d459/nsxt/resource_nsxt_policy_project_test.go#L391)
and [here](https://github.com/vmware/terraform-provider-nsxt/blob/bec90ffaf62e813564a03f32434ec6c077a9d459/nsxt/resource_nsxt_policy_project_test.go#L487)

PR #1336, opened 2024-08-29
External address VPC resource

annakhm:


ksamoray (approved):
No way to test this as we need a VM/port?

nsxt/resource_nsxt_vpc_external_address.go
```diff
@@ -0,0 +1,109 @@
...
+		Read:   resourceNsxtVpcExternalAddressRead,
+		Update: resourceNsxtVpcExternalAddressUpdate,
+		Delete: resourceNsxtVpcExternalAddressDelete,
+		Importer: &schema.ResourceImporter{
+			State: nsxtParentPathResourceImporter,
+		},
+		Schema: map[string]*schema.Schema{
+			"parent_path":                getPolicyPathSchema(true, true, "port path for address binding"),
+			"allocated_external_ip_path": getPolicyPathSchema(true, false, "allocated ip path"),
+			"external_ip_address": {
```
salv-orlando, line 24: Did we check if this is actually computed, or can instead be used to provide an IP address not allocated from a block? (ie: a random IP selected by the user)

annakhm: Its marked as `read-only` in the spec

annakhm:
> No way to test this as we need a VM/port?

We can test this with predefined environment variable. I will post the test in a separate PR

PR #1339, opened 2024-08-30
Rename nsxt_transit_gateway to nsxt_policy_transit_gateway

ksamoray:
For consistency


annakhm:
Test failure here is because fix for VpcServiceProfile is not included in base branch for this PR

PR #1342, opened 2024-08-30
Support for distributed vlan connection resource

annakhm:


salv-orlando (approved):
LGTM.

We can discuss further on whether we want to do vlan ID validation in the provider or let it to NSX.

PR #1352, opened 2024-09-04
Validate no port on VPC subnet before deletion

annakhm:
This PR follows the same logic we have on segment resource in order to make sure all auto-created VPC subnet ports were deleted before attempting subnet deletion

### DIFF
--- a/nsxt/resource_nsxt_vpc_subnet.go
+++ b/nsxt/resource_nsxt_vpc_subnet.go
@@ -7,12 +7,15 @@ import (
 	"fmt"
 	"log"
 	"reflect"
+	"time"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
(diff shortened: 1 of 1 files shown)


nsxt/resource_nsxt_vpc_subnet.go
```diff
@@ -359,8 +362,38 @@ func resourceNsxtVpcSubnetDelete(d *schema.ResourceData, m interface{}) error {
...
+	pendingStates := []string{"pending"}
+	targetStates := []string{"ok", "error"}
+	stateConf := &resource.StateChangeConf{
+		Pending: pendingStates,
+		Target:  targetStates,
+		Refresh: func() (interface{}, string, error) {
+			portsClient := subnets.NewPortsClient(connector)
+			ports, err := portsClient.List(parents[0], parents[1], parents[2], id, nil, nil, nil, nil, nil, nil)
+			if err != nil {
+				return ports, "error", logAPIError("Error listing VPC subnet ports", err)
```
salv-orlando, line 375: Should this fail because of an intermittent error in NSX, I guess Terraform will still retry invoking provider to perform the whole operation (until the number of retries expires, of course) ?

annakhm: Yes, regular http retry works here for status codes that are configured as retriable

salv-orlando: Got it, thanks!

salv-orlando (approved):
LGTM!

PR #1356, opened 2024-09-04
Allow configuration of nsx_id in policy rule

annakhm:
Due to a incorrect code, this attributed was defined as read-only


salv-orlando (approved):
LGTM!

PR #1359, opened 2024-09-06
Enable VPC policy rule test with context profiles

annakhm:
This test creates context profile and shares it with the VPC

### DIFF
--- a/nsxt/utils_test.go
+++ b/nsxt/utils_test.go
@@ -719,6 +719,40 @@ func testAccNsxtPolicyMultitenancyContext() string {
 	return ""
 }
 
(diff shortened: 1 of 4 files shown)


nsxt/utils_test.go
```diff
@@ -719,6 +719,40 @@ func testAccNsxtPolicyMultitenancyContext() string {
 	return ""
 }
 
+func testAccNsxtMultitenancyContext(includeVpc bool) string {
```
ksamoray, line 722: Maybe it's a good idea to abandon `testAccNsxtPolicyMultitenancyContext` entirely and replace it with this (obviously not in this PR's scope).

ksamoray (approved):

PR #1361, opened 2024-09-10
Remove context block from shared resource

annakhm:
Context is determined fromthe parent share_path
For consistency with other resources, we don't expose the context block (which provides redundant info) when parent path is present and required.


ksamoray:
Test failure seem legit, can you take a look?

ksamoray (approved):

PR #1368, opened 2024-09-16
VPC subnet DHCP IPv4 static bindings resource

ksamoray:


nsxt/resource_nsxt_vpc_dhcp_v4_static_binding_config.go
```diff
@@ -0,0 +1,387 @@
...
+										},
+									},
+								},
+							},
+							Optional: true,
+						},
+						Metadata: metadata.Metadata{
+							SchemaType:   "list",
+							SdkFieldName: "Others",
+						},
```
annakhm, line 184: I think `ReflectType` is missing here

nsxt/resource_nsxt_vpc_dhcp_v4_static_binding_config.go
```diff
@@ -0,0 +1,387 @@
...
+															SdkFieldName: "Network",
+														},
+													},
+												},
+											},
+											Optional: true,
+										},
+										Metadata: metadata.Metadata{
+											SchemaType:   "list",
+											SdkFieldName: "StaticRoutes",
```
annakhm, line 131: `ReflectType` is missing here too

nsxt/resource_nsxt_vpc_dhcp_v4_static_binding_config_test.go
```diff
@@ -0,0 +1,219 @@
...
+resource "nsxt_vpc_dhcp_v4_static_binding" "test" {
+  parent_path = nsxt_vpc_subnet.test.path
+  display_name = "%s"
+  description  = "%s"
+  gateway_address = "%s"
+  host_name = "%s"
+  mac_address = "%s"
+  lease_time = %s
+  ip_address = "%s"
+
```
annakhm, line 202: I'd suggest to cover dhcp options here

nsxt/resource_nsxt_vpc_dhcp_v4_static_binding_config.go
```diff
@@ -0,0 +1,387 @@
...
+			Type:         schema.TypeString,
+			ValidateFunc: validation.IsIPv4Address,
+			Optional:     true,
+		},
+		Metadata: metadata.Metadata{
+			SchemaType:   "string",
+			SdkFieldName: "IpAddress",
+		},
+	},
+	"options": {
```
annakhm, line 86: not sure its worth it considering auto-generation, but we can consider reusing options schema from vpc subnet

nsxt/resource_nsxt_vpc_dhcp_v4_static_binding_config.go
```diff
@@ -0,0 +1,387 @@
...
+															ValidateFunc: validateCidrOrIPOrRange(),
+															Optional:     true,
+														},
+														Metadata: metadata.Metadata{
+															SchemaType:   "string",
+															SdkFieldName: "Network",
+														},
+													},
+												},
+											},
```
annakhm, line 126: We'll avoid edge case issues by making this required (no other reason to specify `option121`)

nsxt/resource_nsxt_vpc_dhcp_v4_static_binding_config.go
```diff
@@ -0,0 +1,387 @@
...
+							SdkFieldName: "Option121",
+							ReflectType:  reflect.TypeOf(model.DhcpOption121{}),
+						},
+					},
+					"other": {
+						Schema: schema.Schema{
+							Type: schema.TypeList,
+							Elem: &metadata.ExtendedResource{
+								Schema: map[string]*metadata.ExtendedSchema{
+									"code": {
```
annakhm, line 149: I think `code` should be `Required` as well (not 100% sure about `values`)

nsxt/resource_nsxt_vpc_dhcp_v4_static_binding_config.go
ksamoray: As these API types (`DhcpV4StaticBindingConfig`, `VpcSubnet`) use the same type for options, it makes sense but I wouldn't break autogeneration for this.
Maybe we should consider generation of a per-SDK-struct set of APIs, for example, as in this case the type of the attribute is `DhcpV4Options, we could generate a `getDhcpV4Options()` func which will return the schema def. So later it can be reused - maybe with some other utilities for that struct. And then reuse it wherever this struct is used.

nsxt/resource_nsxt_vpc_dhcp_v4_static_binding_config.go
ksamoray: These are lists - I didn't see any ReflectType settings for other lists in the code. What should its ReflectType be anyway?

website/docs/r/vpc_dhcp_v4_static_binding_config.html.markdown
```diff
@@ -0,0 +1,121 @@
...
+* `tag` - (Optional) A list of scope + tag pairs to associate with this resource.
+* `nsx_id` - (Optional) The NSX ID of this resource. If set, this ID will be used to create the resource.
+* `gateway_address` - (Optional) When not specified, gateway address is auto-assigned from segment configuration.
+* `host_name` - (Optional) Hostname to assign to the host.
+* `mac_address` - (Optional) MAC address of the host.
+* `lease_time` - (Optional) DHCP lease time in seconds.
+* `ip_address` - (Optional) IP assigned to host. The IP address must belong to the subnet, if any, configured on Segment.
+* `options` - (Optional) None
+  * `option121` - (Optional) None
+    * `static_route` - (Optional) Classless static route of DHCP option 121.
```
annakhm, line 63: `Required` here and below in `code`

ksamoray: Yep. Also network, gateway should be required.
BTW do you think that we should require these also in vpc_subnet? Everything is optional there.

annakhm: Yes, I think so

annakhm (approved):

PR #1374, opened 2024-09-19
Add default_security_profile attribute to project

ksamoray:
Add the attribute above and manage it vs the VpcSecurityProfiles client

### DIFF
--- a/nsxt/resource_nsxt_policy_project.go
+++ b/nsxt/resource_nsxt_policy_project.go
@@ -12,6 +12,7 @@ import (
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 	infra "github.com/vmware/vsphere-automation-sdk-go/services/nsxt/orgs"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/orgs/projects"
 )
 
 func resourceNsxtPolicyProject() *schema.Resource {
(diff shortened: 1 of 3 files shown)


nsxt/resource_nsxt_policy_project.go
```diff
@@ -143,7 +169,41 @@ func resourceNsxtPolicyProjectPatch(connector client.Connector, d *schema.Resour
 	log.Printf("[INFO] Patching Project with ID %s", id)
 
 	client := infra.NewProjectsClient(connector)
-	return client.Patch(defaultOrgID, id, obj)
+	err := client.Patch(defaultOrgID, id, obj)
```
annakhm, line 172: is this change needed?

nsxt/resource_nsxt_policy_project.go
```diff
@@ -143,7 +169,41 @@ func resourceNsxtPolicyProjectPatch(connector client.Connector, d *schema.Resour
...
+	if err != nil {
+		return err
+	}
+	return patchVpcSecurityProfile(d, connector, id)
```
annakhm, line 176: I suggest to wrap this in `HasChange ("default_security_profile")` to avoid unnecessary call

nsxt/resource_nsxt_policy_project.go
```diff
@@ -143,7 +169,41 @@ func resourceNsxtPolicyProjectPatch(connector client.Connector, d *schema.Resour
...
+			// This is safe as TF enforces
+			nsfw := northSouthFirewall.([]interface{})[0].(map[string]interface{})
+			enabled = nsfw["enabled"].(bool)
+		}
+	}
+	client := projects.NewVpcSecurityProfilesClient(connector)
+	_, err := client.Get(defaultOrgID, projectID, "default")
+
+	if isNotFoundError(err) && !enabled {
+		// No action required
```
annakhm, line 194: Are we dealing with the case in which default profile was not created yet? In this case, should we busy wait here?

annakhm (approved):
We could use H-API to avoid two roundtrips but I don't think is worth it

PR #1377, opened 2024-09-23
Project: use TypeSet for some attributes

salv-orlando:
This change updates external_ipv4_block and tgw_external_connections to use schema.TypeSet. This will ensure no duplicate entries are stored in NSX, even if specified in the Terraform configuration.


ksamoray (approved):
LGTM, but why isn't that rejected by NSX?

salv-orlando:
We should not be changing lists into sets, as resource order might always be important for NSX.

PR #1384, opened 2024-09-25
Disable MP Resources from NSX 9.0.0 onwards

salv-orlando:
This change removes support for NSX MP API client when the NSX manager version is higher or equal to 9.0.0

For MP resources, with NSX >= 9.0.0:
- terraform apply, destroy, and import operations will fail with the following error: "This resource is not supported with given provider settings. Management Plane resources are not supported since NSX 9.0.0;
- terraform plan operations for resources already in the Terraform state will fail with the same error;
- all terraform operations involving MP data sources will fail with the same error;
- terraform plan operations for resource not yet in the Terraform state will succeed, but will then fail at the apply stage. This because Terraform won't perform a read from the backend if the resource is not present at all in the state.


nsxt/provider.go
```diff
@@ -1090,6 +1090,12 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 		return nil, err
 	}
 
+	initNSXVersion(getPolicyConnector(clients))
+	if util.NsxVersionHigherOrEqual("9.0.0") {
+		// Disable MP client from 9.0.0 onwards
```
annakhm, line 1095: This is fine for now, but ideally we should not invoke `configureNsxtClient` when NSX is 9.0.0 and above.
This would require us to rewrite the session create API which is currently handle fired from inside `configureNsxtClient`. We should do this regardless, because it paves the way of removing the old SDK when the time comes.

annakhm:
Food for thought: imagine a user has MP resources in their state, and they upgrade to 9.0.0. They understand that those are no longer supported and try to delete them from the config and apply. With current PR, apply will fail on delete.
We might want to offer `noop` for the delete operation in this case (which will require a change in every MP resource).
Another option is to add a note in documentation explaining that resources need to be manually deleted from the state.
What do you think?

salv-orlando:
> Food for thought: imagine a user has MP resources in their state, and they upgrade to 9.0.0. They understand that those are no longer supported and try to delete them from the config and apply. With current PR, apply will fail on delete. Moreover, even refresh will fail. We might want to offer `noop` for the delete and read operations in this case (which will require a change in every MP resource). Another option is to add a note in documentation explaining that resources need to be manually deleted from the state as well as config. What do you think?

All MP resources should now fail on create, delete or import, and do a no-op on read and update.
For MP data sources, they will fail on read.

nsxt/provider.go
salv-orlando: I've removed this code. I will handle the change in session create API with a new commit in this PR or a new PR

annakhm:
Unfortunately this elegant solution won't work in conjunction with `on_demand_connection` option. In case of delayed connection, `initNSXVersion` will not work since NSX backend is not ready yet.

annakhm (approved):

PR #1385, opened 2024-09-26
Rename nsxt_policy_ip_address_allocation funcs

ksamoray:
Functions for the resource above were using a non-policy prefixed names which is inconsistent.


salv-orlando (approved):
LGTM

PR #1391, opened 2024-10-01
TGW attachment connection_path attribute should be required

ksamoray:
This is not properly set in the API spec but NSX enforces this requirement.


salv-orlando (approved):
LGTM

PR #1400, opened 2024-10-04
Improve validation for project importers

annakhm:
For resources residing under a project, we only support policy path as import ID. This change introduces PolicyPathOnly importer that only allows policy paths, as opposed to old validator that allows both policy path and ID

### DIFF
--- a/nsxt/policy_utils.go
+++ b/nsxt/policy_utils.go
@@ -407,6 +407,7 @@ func validateImportPolicyPath(policyPath string) error {
 	return nil
 }
 
+// This importer function accepts policy path and resource ID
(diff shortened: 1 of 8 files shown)


ksamoray:
Should be useful also for:
nsxt_policy_global_manager
nsxt_policy_host_transport_node_profile
nsxt_policy_share
nsxt_policy_site
nsxt_vpc_connectivity_profile
nsxt_vpc_service_profile

salv-orlando (approved):
not sure if you want to add also the resources suggested by Kobi, otherwise LGTM!

annakhm:
I have added vpc-related resources from Kobi's comment. Lets fix the other post-merge

ksamoray (approved):

PR #1403, opened 2024-10-08
Fix VPC20 tests for latest platform changes

ksamoray:


annakhm:
should this also disable test for transit gateway itself?

ksamoray:
> should this also disable test for transit gateway itself?

Makes sense - I guess that we do not run this test with VPC20 platform. It's something which I should address.

PR #1405, opened 2024-10-08
VPC and TGW NAT rules: make Action Required

salv-orlando:
NSX mandates rule action to be specified for both TGW and VPC NAT rules, but in the API spec this attribute is not specified as required.

This change addresses this.
Resources' documentaiton has been updated to mark the attribute as required.


nsxt/resource_nsxt_vpc_nat_rule.go
```diff
@@ -76,6 +76,7 @@ var policyVpcNatRuleSchema = map[string]*metadata.ExtendedSchema{
 			Type:         schema.TypeString,
 			ValidateFunc: validation.StringInSlice(policyVpcNatRuleActionValues, false),
 			Optional:     true,
```
ksamoray, line 78: It can't be both optional and required, I think

salv-orlando: Why? It's an Heisen-Attribute!

ksamoray (approved):

annakhm (approved):

PR #1414, opened 2024-10-10
Mark partial update when NSX update fails for vpc

annakhm:
After merging VPC branch, this will be refactored to apply to all resources as part of handleUpdateError routine.


salv-orlando (approved):
LGTM

PR #1434, opened 2024-10-16
Remove IP blocks from connectivity profiles NAT

annakhm:
This attribute is removed from first VPC2.0 release on platform


ksamoray (approved):
LGTM

PR #1436, opened 2024-10-16
fix: replace deprecated `rand.Seed`

tenthirtyam:
### Summary

`rand.Seed` has been deprecated since Go 1.20 and an alternative has been available since Go 1.0.

As of Go 1.20 there is no reason to call `Seed` with a random value. Programs that call `Seed` with a known value to get a specific sequence of results should use `New(NewSource(seed))` to obtain a local random generator.


ksamoray:
> LGTM. @ksamoray can you please have a look at this?

Already did - but would rather wait to have a CI for acceptance tests before merging this (and few of the other changes).

tenthirtyam:
@vmware/terraform-provider-nsxt-maintainers: reminder.

salv-orlando:
@ksamoray If possible can we run acceptance tests for this PR, so we can merge it?

ksamoray (approved):

Issue #1441, opened 2024-10-21
Cannot use nsxt_policy_vtep_ha_host_switch_profile in nsxt_policy_host_transport_node_profile


martinrohrbach:
### Describe the bug

We have adopted the nsxt_policy_host_transport_node_profile in our environment and after some initial problems (see my last issues ;) we are quite happy with that.

We are now looking into using the nsxt_policy_vtep_ha_host_switch_profile for our transport nodes. We've tested the profiles manually in our test environment by adding the VTEP HA profiles using the API and that worked fine. However, when we tried using the vTEP HA profile in the nsxt_policy_host_transport_node_profile resource, the provider fails:

```
│
│ Error:  Failed to update Policy Host Transport Node Profile 199f39ef-9406-4a69-b9a4-6c8876573ea5: Unable to find UplinkProfile associated with Id BaseHostSwitchProfile/ac3d4c95-7500-4062-8dd0-b0e2233908cd. (code 8500)
│
│   with nsxt_policy_host_transport_node_profile.trf-tnp["cluster1"],
│   on [transport-node-profiles.tf](http://transport-node-profiles.tf/) line 109, in resource "nsxt_policy_host_transport_node_profile" "trf-tnp":
│  109: resource "nsxt_policy_host_transport_node_profile" "trf-tnp" {
│
```

We've simply added the profile to the host_switch_profile property of the resource, but apparently that expects only "BaseHostSwitchProfiles" rather than VTEP HA ones. 

As I currently don't see a way to add the latter, is this possibly something that can easily be added (either by accepting them in host_switch_profile or by adding an additional property)?

Or is it already possible and we overlooked something?

### Reproduction steps

1. Create nsxt_policy_vtep_ha_host_switch_profile resource
2. Use the resouce as a profile for nsxt_policy_host_transport_node_profile.host_switch_profile
3. Provider fails to create transport node profile (see above)

### Expected behavior

We can use the VTEP HA profile to create a host transport profile.

ksamoray:
Hi @martinrohrbach,
Can you please include the following info:

- NSX, provider versions
- Output of the `terraform apply` execution while the following flags are set:
```
export TF_LOG=INFO
export TF_LOG_PROVIDER_NSX_HTTP=1
```

martinrohrbach:
Fix confirmed in 3.71, many thanks!

ksamoray:
@martinrohrbach YW, thanks for reporting this!

PR #1442, opened 2024-10-22
Add proper host_switch_profile type with id, while setting in hostswitch

ksamoray:
The code assumes that specified host_switch_profile is always of type uplink which is incorrect.

Fixes: #1441

### DIFF
--- a/nsxt/resource_nsxt_edge_transport_node.go
+++ b/nsxt/resource_nsxt_edge_transport_node.go
@@ -9,8 +9,6 @@ import (
 	"reflect"
 	"time"
 
-	"github.com/vmware/terraform-provider-nsxt/nsxt/util"
-
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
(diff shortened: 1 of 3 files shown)


nsxt/resource_nsxt_edge_transport_node.go
```diff
@@ -29,47 +31,56 @@ var ipAssignmentTypes = []string{
 	"static_ip_pool",
 }
 
+var mpTypeFromPolicyType = map[string]string{
```
annakhm, line 34: If the code assumes MP and Policy values are the same, do we need this translation?

ksamoray: We query the host switch profile type using this policy [API](https://dp-downloads.broadcom.com/api-content/apis/API_NTDCRA_001/4.2.1/html/api_includes/method_GetPolicyHostSwitchProfile.html) which returns policy types, e.g `PolicyUplinkHostSwitchProfile`.
The switch type consumes [HostSwitchSpec](https://dp-downloads.broadcom.com/api-content/apis/API_NTDCRA_001/4.2.1/html/api_includes/types_StandardHostSwitch.html) type has `host_switch_profile_id` attribute which is a string - which is enumerated by [HostSwitchProfileType](https://dp-downloads.broadcom.com/api-content/apis/API_NTDCRA_001/4.2.1/html/api_includes/types_HostSwitchProfileType.html) that has no "Policy" prefix to its values. Hence translation is required even though it's not MP.
However `mpTypeFromPolicyType` name is bad - it's just `HostSwitchProfileType` translation and not any Policy->MP translation. I'll rename this var.

annakhm (approved):

PR #1444, opened 2024-10-22
Clarify VPC group documentation

annakhm:


ksamoray:
Should we validate this in any way? Or we leave this for NSX

annakhm:
> Should we validate this in any way? Or we leave this for NSX

We wouldn't want to update those validations every time another value is added (which happens quite a lot), so I'd rather leave it to NSX.

ksamoray (approved):

PR #1454, opened 2024-10-29
Pre-check bundle should be optional also for 4.1.1 and beyond

ksamoray:
Upgrade bundle contains the pre-checks as well. This attribute can be still used when a user would like to run pre-checks from a later build than the upgrade itself.


salv-orlando (approved):
LGTM

PR #1457, opened 2024-11-03
Handle default_security_profile attribute in project read func

ksamoray:
nsxt_policy_project has a default_security_profile attribute but it was not handled correctly in read operation.


salv-orlando (approved):
LGTM

PR #1463, opened 2024-11-06
Fix crash related to empty block in lb_vs resource

annakhm:
Empty block can be configured when no required attributes are defined in the block. In this case nil protection is required, since empty block is presented as nil rather than empty map in the schema structure.

### DIFF
--- a/nsxt/resource_nsxt_policy_lb_virtual_server.go
+++ b/nsxt/resource_nsxt_policy_lb_virtual_server.go
@@ -903,6 +903,10 @@ func getPolicyLbRuleVariablePersistenceLearnActionSchema() *schema.Schema {
 func getPolicyClientSSLBindingFromSchema(d *schema.ResourceData) *model.LBClientSslProfileBinding {
 	bindings := d.Get("client_ssl").([]interface{})
 	for _, binding := range bindings {
+		if binding == nil {
+			// empty clause
+			continue
+		}
 		// maximum count is one
 		data := binding.(map[string]interface{})
 		chainDepth := int64(data["certificate_chain_depth"].(int))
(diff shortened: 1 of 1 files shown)


nsxt/resource_nsxt_policy_lb_virtual_server.go
```diff
@@ -903,6 +903,10 @@ func getPolicyLbRuleVariablePersistenceLearnActionSchema() *schema.Schema {
 func getPolicyClientSSLBindingFromSchema(d *schema.ResourceData) *model.LBClientSslProfileBinding {
 	bindings := d.Get("client_ssl").([]interface{})
 	for _, binding := range bindings {
+		if binding == nil {
```
ksamoray, line 906: This is unsafe as `d.Get()` result can be nil as well and casting it will result a crash.
As we have multiple instances in the scope of this resource and multiple others in other resources, can we use a func instead of the range which will do all the validations, e.g check that input (`d.Get()` result) is not null, that members aren't null as well etc?
e.g:
```
func iterateSchemaSlice(s interface{}, f func(interface{})) {
    if s == nil {
        return
    }
    for _, i := range s.([]interface{}) {
        if i == nil {
            continue
        }
        f(i)
    }
}
```

And call it as:

```
  iterateSchemaSlice(d.Get("client_ssl"), func(binding interface{}){
		data := binding.(map[string]interface{})
		chainDepth := int64(data["certificate_chain_depth"].(int))    
  })
```

annakhm: If the key exists in schema, then nil is not expected to be returned, as stated [here](https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/schema/resource_data.go#L57)

salv-orlando: Thanks for the clarification Anna.
A pedant note. Do you think it might be better to rewrite this to:

```
if binding != nil {
		// maximum count is one
		data := binding.(map[string]interface{})
		chainDepth := int64(data["certificate_chain_depth"].(int))
}
```

annakhm: In most cases that are addressed in this patch, the code block that would need to be indented is rather big, so I would leave it as is for the sake of readability

salv-orlando (approved):
LGTM, no additional comment from me

PR #1468, opened 2024-11-21
Add realized_id attribute to nsxt_policy_transport_zone data source

ksamoray:
This data source is replacing MP nsxt_transport_zone which is deprecated. The nsxt_edge_transport_node resource requires this value, as it's not a policy resource and it is provided by nsxt_policy_transport_zone resource but not by the data source.


website/docs/d/policy_transport_zone.html.markdown
```diff
@@ -54,3 +54,4 @@ In addition to arguments listed above, the following attributes are exported:
 * `is_default` - A boolean flag indicating if this Transport Zone is the default.
 * `transport_type` - The transport type of this transport zone.
 * `path` - The NSX path of the policy resource.
+* `realized_id` - The id of realized pool object. This id should be used in `nsxt_edge_transport_node` resource.
```
annakhm, line 57: `pool` is irrelevant here

salv-orlando (approved):
LGTM

PR #1471, opened 2024-11-26
Ignore create_service_account with NSX 9.0.0

annakhm:
In compute manager resource, NSX no longer accepts false value for create_service_account attribute.
Since terraform sdk can't distinguish between unset value and false value for the attribute, we will ignore the attribute entirely for NSX 9.0 and above


salv-orlando (approved):
LGTM

ksamoray (approved):

Issue #1459, opened 2024-11-04
nsxt_edge_transport_node not usable for existing edge


martinrohrbach:
### Describe the bug

We have a setup where we use edges from a compute manager that is not registered with the NSX instance that will use the edge (and we also cannot register the computer manager because it is already registered with a different NSX instance and we cannot use multi NSX due to vLCM integration).

It is absolutely possible to use an existing edge for a new transport node, the API docs (https://dp-downloads.broadcom.com/api-content/apis/API_NTDCRA_001/4.2.1/html/api_includes/method_CreateTransportNodeWithDeploymentInfo.html) have this to say:

> The request should either provide node_deployement_info or node_id.
> 
> If the host node (hypervisor) or edge node (router) is already added in
> system then it can be converted to transport node by providing node_id in
> request.

However, I cannot specify a "node_id" using the nsxt_edge_transport_node resource afaics.

Would it be possible to add this property to the resource so we can either pass in a "deployment_config" OR a "node_id" (or "existing_deployment_id" or something) for the creation of the edge transport node?
 

### Reproduction steps

1. Create and register an edge VM in NSX (outside terraform)
2. Use nsxt_edge_transport_node without "deployment_config" to use the existing edge VM
3. Fails with "General error has occured"

### Expected behavior

We can create a edge transprt node from an existing edge VM.

annakhm:
Hi @martinrohrbach, the doc you quoted also [defines](https://[dp-downloads.broadcom.com/api-content/apis/API_NTDCRA_001/4.2.1/html/api_includes/types_TransportNode.html](https://dp-downloads.broadcom.com/api-content/apis/API_NTDCRA_001/4.2.1/html/api_includes/types_TransportNode.html)) `node_id` as both `Deprecated` and `ReadOnly`.
We normally don't expose deprecated attributes in the provider. Looks like there's a mismatch in NSX documentation.

martinrohrbach:
Hi @annakhm you're right, I didn't notice that. When you say "normally" what does that mean for this issue? 

I actually took the API call that the provider crafted, removed the "node_deployment_info" block, added the "node_id" and POSTed thet to /api/v1/trasport-nodes. That had exactly the effect that I wanted and the edge is now configured and usable and it's also what the docs suggest according to the snippet I posted above. 

All in all I think the read-only and deprecated flags are wrong and that's where the mismatch comes from. 

[...]

PR #1473, opened 2024-11-27
Add support for Edge transport nodes which were created externally

ksamoray:
Normally users create Edge Transport Nodes within NSX, which deploys them into the regsitered compute manager. However, users have the option to do the same by deploying the Edge appliance anywhere, outside NSX, and registering it with NSX using the Edge CLI or OVA parameters. Later, this Edge appliance can be converted into a transport node using NSX API - this change attempts to utilize this capability.

Fixes: #1459

### DIFF
--- a/nsxt/resource_nsxt_edge_transport_node.go
+++ b/nsxt/resource_nsxt_edge_transport_node.go
@@ -97,6 +97,13 @@ func resourceNsxtEdgeTransportNode() *schema.Resource {
 			"description":  getDescriptionSchema(),
 			"display_name": getDisplayNameSchema(),
 			"tag":          getTagsSchema(),
+			"node_id": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				Description:   "Unique Id of the fabric node",
+				ConflictsWith: []string{"ip_addresses", "fqdn", "deployment_config", "external_id"},
+			},
 			"failure_domain": {
 				Type:        schema.TypeString,
 				Optional:    true,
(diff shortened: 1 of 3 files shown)


nsxt/resource_nsxt_edge_transport_node.go
```diff
@@ -97,6 +97,12 @@ func resourceNsxtEdgeTransportNode() *schema.Resource {
 			"description":  getDescriptionSchema(),
 			"display_name": getDisplayNameSchema(),
 			"tag":          getTagsSchema(),
+			"node_id": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
+				Description: "Unique Id of the fabric node",
+			},
```
salv-orlando, line 105: This is probably a bad idea, but I'll suggest it anyway.
As this resource is not generated, we do not have to fully respect the NSX API schema.

Perhaps we can move node_id in deployment_config and specify that it conflictsWith vm_deployment_config?

nsxt/resource_nsxt_edge_transport_node.go
```diff
@@ -671,43 +677,55 @@ func getTransportNodeFromSchema(d *schema.ResourceData, m interface{}) (*mpmodel
...
+	var nodeDeploymentInfo *data.StructValue
+	if nodeID == "" {
+		/*
+			node_id attribute conflicts with node_deployment_info. As node_deployment_info is a complex object which has
+			attributes with default values, this schema property will always have values - therefore there is no simple way
+			to enforce this conflict within the provider (e.g check if node_id and node_deployment_info have values, then
+			fail.
+			So the provider will ignore node_deployment_info properties when node_id has a value - which would mean that
+			this edge appliance was created externally.
+		*/
```
salv-orlando, line 696: It might be helpful to also add a log statement here for debugging purposes.

e.g: "nodeID not specified, will deploy edge using values in deploymentConfig"

nsxt/resource_nsxt_edge_transport_node.go
```diff
@@ -671,43 +677,55 @@ func getTransportNodeFromSchema(d *schema.ResourceData, m interface{}) (*mpmodel
...
 		return nil, fmt.Errorf("failed to create Transport Node: %v", err)
 	}
 
-	converter := bindings.NewTypeConverter()
-	var dataValue data.DataValue
-	var errs []error
+	var nodeDeploymentInfo *data.StructValue
+	if nodeID == "" {
+		/*
+			node_id attribute conflicts with node_deployment_info. As node_deployment_info is a complex object which has
```
annakhm, line 690: Could you please clarify - even when specifying `node_id`, user still needs to populate `node_deployment_info`? What properties are needed?
I don't think `node_deployment_info` is `Computed`

nsxt/resource_nsxt_edge_transport_node.go
ksamoray: Done

nsxt/resource_nsxt_edge_transport_node.go
ksamoray: The resource has no `node_deployment_info` attribute as it's not implementing the entire API, as host nodes are created with policy API. Relevant `node_deployment_info` attributes are placed under `deployment_config`, and directly under the resource root.
I've set these attributes as conflicting with `node_id`.

nsxt/resource_nsxt_edge_transport_node.go
```diff
@@ -725,6 +745,16 @@ func resourceNsxtEdgeTransportNodeCreate(d *schema.ResourceData, m interface{})
 	connector := getPolicyConnector(m)
 	client := nsx.NewTransportNodesClient(connector)
 
+	nodeID := d.Get("node_id").(string)
+	if nodeID != "" {
+		_, err := client.Get(nodeID)
```
annakhm, line 750: We don't need to set revision here?

salv-orlando:
Keeping this PR in 3.8.0 milestone as we're almost done with the review process.

nsxt/resource_nsxt_edge_transport_node.go
```diff
@@ -725,6 +745,29 @@ func resourceNsxtEdgeTransportNodeCreate(d *schema.ResourceData, m interface{})
...
+		converter := bindings.NewTypeConverter()
+		base, errs := converter.ConvertToGolang(obj.NodeDeploymentInfo, mpmodel.EdgeNodeBindingType())
+		if errs != nil {
+			return handleCreateError("TransportNode", nodeID, errs[0])
+		}
+		node := base.(mpmodel.EdgeNode)
+		d.Set("external_id", node.ExternalId)
+		d.Set("fqdn", node.Fqdn)
+		d.Set("ip_addresses", node.IpAddresses)
+
```
annakhm, line 766: The revision is still missing - is that on purpose?

ksamoray: Nope, missed it...

nsxt/resource_nsxt_edge_transport_node.go
```diff
@@ -725,6 +745,32 @@ func resourceNsxtEdgeTransportNodeCreate(d *schema.ResourceData, m interface{})
...
+		node := base.(mpmodel.EdgeNode)
+		d.Set("external_id", node.ExternalId)
+		d.Set("fqdn", node.Fqdn)
+		d.Set("ip_addresses", node.IpAddresses)
+		if obj.Revision != nil {
+			d.Set("revision", obj.Revision)
+		}
+
+		d.SetId(nodeID)
+		return resourceNsxtEdgeTransportNodeUpdate(d, m)
```
annakhm, line 771: In case update fails, there might be a state problem here. Since we already performed `d.SetId(nodeId)`, the resource will be already present in state, if I'm not mistaken. 
If would be safer to catch error here, and in case Update fails, nullify the Id.

salv-orlando (approved):

Issue #696, opened 2021-12-02
Policy Datasources for leveraging built-in services


akhomyakov:
Hi folks,
Looking at the nsxt_ns_groups and nsxt_ns_services data resources (currently in Beta) they appear like exactly what is needed to make use of built in services and so on. 
What I can't figure out is how to refer to items imported. Is there a way to output the content of what is imported?
```
<snip for brevity>
data "nsxt_ns_groups" "NS-GROUPS" {
}
data "nsxt_ns_services" "NS-SERVICES" {
}
resource "nsxt_policy_security_policy" "SHARED-INFRASTRUCTURE" {
  display_name = "SHARED-INFRASTRUCTURE"
  description  = "Service access policy shared by all tenants"
  category     = "Infrastructure"
  locked       = false
  stateful     = true
  tcp_strict   = false

  rule {
    display_name       = "ICMP"
    destination_groups = [data.nsxt_ns_groups.NS-GROUPS.items["INTDMZ"]]
    action             = "ALLOW"
    services           = [data.nsxt_ns_services.NS-SERVICES.items["ICMP-ALL"]]
    logged             = true
  }
<snip for brevity>
```
```
<snip for brevity>
╷
│ Error: Invalid index
│
│   on firewall-policy-SHARED-INFRASTRUCTURE.tf line 12, in resource "nsxt_policy_security_policy" "SHARED-INFRASTRUCTURE":
│   12:     destination_groups = [data.nsxt_ns_groups.NS-GROUPS.items["INTDMZ"]]
│     ├────────────────
│     │ data.nsxt_ns_groups.NS-GROUPS.items is map of string with 9 elements
│
│ The given key does not identify an element in this collection value.
╵
╷
│ Error: Invalid index
│
│   on firewall-policy-SHARED-INFRASTRUCTURE.tf line 14, in resource "nsxt_policy_security_policy" "SHARED-INFRASTRUCTURE":
│   14:     services           = [data.nsxt_ns_services.NS-SERVICES.items["ICMP-ALL"]]
│     ├────────────────
│     │ data.nsxt_ns_services.NS-SERVICES.items is map of string with 373 elements
│
│ The given key does not identify an element in this collection value.
╵
<snip for brevity>
```

Just using the API doesn't seem to yeild the right results
```
// GET https://{{nsxmanager}}/policy/api/v1/infra/domains/default/groups
{
    "results": [
        {
            "expression": [
                {
                    "ip_addresses": [
[...]

PR #1476, opened 2024-12-03
Implement map data sources for policy services and groups

ksamoray:
These data sources allows the retrieval of the entire tables from NSX to save multiple calls to the backend.

Fixes: #696

### DIFF
--- a/nsxt/data_source_nsxt_policy_groups.go
+++ b/nsxt/data_source_nsxt_policy_groups.go
@@ -1,0 +1,50 @@
+// © Broadcom. All Rights Reserved.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MPL-2.0
+
+package nsxt
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+	"github.com/vmware/terraform-provider-nsxt/api/infra/domains"
+)
+
+func dataSourceNsxtPolicyGroups() *schema.Resource {
+	return &schema.Resource{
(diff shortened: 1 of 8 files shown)


nsxt/data_source_nsxt_policy_groups.go
```diff
@@ -0,0 +1,61 @@
...
+
+	client := domains.NewGroupsClient(getSessionContext(d, m), connector)
+
+	groupsMap := make(map[string]interface{})
+	results, err := client.List(domainName, nil, nil, nil, nil, nil, nil, nil)
+	if err != nil {
+		return err
+	}
+	for _, r := range results.Results {
+		groupMap := make(map[string]interface{})
```
annakhm, line 52: do we want to error out for duplicate display_name?

ksamoray: Yep done

website/docs/d/policy_groups.html.markdown
```diff
@@ -0,0 +1,69 @@
...
+
+  tag {
+    scope = "color"
+    tag   = "orange"
+  }
+
+  rule {
+    display_name = "allow_icmp"
+    destination_groups = [
+      data.nsxt_policy_groups.test.items[index(data.nsxt_policy_groups.test.items.*.display_name, "Cats")].path,
```
annakhm, line 30: I feel this syntax might be too complex to be widely adopted.. perhaps it would be more useful to offer just display_name=>path map.

annakhm (approved):
Perhaps in the services map owner filter could be useful to identify predefined services, but we can wait with this until someone asks for it

PR #1480, opened 2024-12-10
Add nat type to nat rule resource

annakhm:


annakhm:
PASS: TestAccResourceNsxtPolicyNATRuleT1_natType (48.34s)
PASS: TestAccResourceNsxtPolicyNATRule_minimalT0 (26.83s)
PASS: TestAccResourceNsxtPolicyNATRule_basicT1Import (34.93s)
PASS: TestAccResourceNsxtPolicyNATRuleNoSnatWithoutTNet (41.86s)
PASS: TestAccResourceNsxtPolicyNATRule_nat64T1 (46.48s)
PASS: TestAccResourceNsxtPolicyNATRule_withPolicyBasedVpnMode (46.61s)
PASS: TestAccResourceNsxtPolicyNATRule_basicT0 (52.45s)
PASS: TestAccResourceNsxtPolicyNATRuleT1_natType (60.76s)
PASS: TestAccResourceNsxtPolicyNATRule_basic_T1 (65.04s)
PASS: TestAccResourceNsxtPolicyNATRule_basicT1Import_multitenancy (27.56s)
PASS: TestAccResourceNsxtPolicyNATRule_basicT1_multitenancy (55.67s)

nsxt/resource_nsxt_policy_nat_rule.go
```diff
@@ -231,12 +246,15 @@ func patchNsxtPolicyNATRule(sessionContext utl.SessionContext, connector client.
 	return client.Patch(gwID, natType, *rule.Id, rule)
 }
 
-func getNatTypeByAction(action string) string {
+func getNatTypeByAction(natType string, action string) string {
 	if action == model.PolicyNatRule_ACTION_NAT64 {
 		return model.PolicyNat_NAT_TYPE_NAT64
 	}
```
salv-orlando, line 252: Do we still want the check above if nat_type != ""?

If nat_type is not Policy_NAT_TYPE_NAT64, The current code will "fix" an invalid configuration. It might be ok, but maybe we should let the call fail so users know they have an invalid configuration?

nsxt/resource_nsxt_policy_nat_rule.go
```diff
@@ -394,13 +422,15 @@ func resourceNsxtPolicyNATRuleCreate(d *schema.ResourceData, m interface{}) erro
...
-	err := patchNsxtPolicyNATRule(getSessionContext(d, m), connector, gwID, ruleStruct, isT0)
+	err := patchNsxtPolicyNATRule(getSessionContext(d, m), connector, gwID, ruleStruct, isT0, natType)
 	if err != nil {
 		return handleCreateError("NAT Rule", id, err)
 	}
 
 	d.SetId(id)
 	d.Set("nsx_id", id)
+	// In case nat type was not specified or got overridden by action
+	d.Set("type", natType)
```
salv-orlando, line 433: Is this to restore any change that might have been done in line 311?

nsxt/resource_nsxt_policy_nat_rule.go
```diff
@@ -284,8 +302,14 @@ func resourceNsxtPolicyNATRuleRead(d *schema.ResourceData, m interface{}) error
...
 
-	action := d.Get("action").(string)
-	natType := getNatTypeByAction(action)
+	natType := d.Get("type").(string)
+	if natType == "" {
+		// This can happen when provider was upgraded and we're refreshing an existing resource
+		// This is not an import case, so action should be set
+		action := d.Get("action").(string)
+		natType = getNatTypeByAction(natType, action)
+		d.Set("type", natType)
```
salv-orlando, line 311: Upon refresh, this will update the Terraform state with Type, and will avoid having a diff when doing the planning, is that correct?

nsxt/resource_nsxt_policy_nat_rule.go
annakhm: Good point, I have added the validation to make sure NAT64 action can only go under NAT64 type, and vise versa.
Regarding the `natType == ""` check - this one is merely to replace default value functionality, since default can not be configured together with `Computed` marker.

nsxt/resource_nsxt_policy_nat_rule.go
annakhm: Correct

nsxt/resource_nsxt_policy_nat_rule.go
annakhm: No, this is create flow - if nat type was left blank by the user, we need to set the `Computed` nat type that was determined in line 372

salv-orlando (approved):
lgtm!

PR #1483, opened 2024-12-17
Add `vc_folder` flag to project definition.

annakhm:
This is relevant for NSX 9.0.0 onwards.
In addition, mark `default_security_profile` as Computed to avoid non-empty diff.


annakhm:
Will repost

salv-orlando (approved):

PR #1484, opened 2024-12-17
Add VPC subnet realization data source

annakhm:
This is an alias to segment realization data source, but with naming that is consistent with VPC.


annakhm:
Will repost

Issue #1481, opened 2024-12-11
Resouere import runs into endless loop


martinrohrbach:
### Describe the bug

We're trying to specifically import a nsxt_logical_tier1_router resource with
NSX-T: 4.1.2.5
Provider: 3.7.1

But the import runs into an endless loop and never comes back. Note that the import works with provider version 3.7.0 so it must be a recent change.

I can see the resource just fine using this code here so communication and authentication works fine and the resource exists:
```
data "nsxt_logical_tier1_router" "trf-test" {
  display_name = "import-test-delete-me-if-seen-in-2025"
}

$ terraform plan
data.nsxt_logical_tier1_router.trf-test: Reading...
data.nsxt_logical_tier1_router.trf-test: Read complete after 0s [id=7790f680-bf9c-4835-9f35-eda70573fc42]

No changes. Your infrastructure matches the configuration.
```
But when I try to import the resource, the plan never comes back:
```
import {
  to = nsxt_logical_tier1_router.trf-test
  id = "7790f680-bf9c-4835-9f35-eda70573fc42"
}

resource "nsxt_logical_tier1_router" "trf-test" {
  description     = "trf-test"
  display_name    = "trf-test"
}

$ terraform plan
nsxt_logical_tier1_router.trf-test: Preparing import... [id=7790f680-bf9c-4835-9f35-eda70573fc42]
(hangs forever)
```

With TF_LOG=TRACE, we get the following:
```
$ TF_DEBUG="TRACE" terraform plan
... skipping to the actual import ...
nsxt_logical_tier1_router.trf-test: Preparing import... [id=7790f680-bf9c-4835-9f35-eda70573fc42]
2024-12-11T11:10:05.863+0100 [TRACE] GRPCProvider: ImportResourceState
2024-12-11T11:10:05.863+0100 [TRACE] GRPCProvider: GetProviderSchema
2024-12-11T11:10:05.863+0100 [TRACE] GRPCProvider: returning cached schema: EXTRA_VALUE_AT_END=[registry.terraform.io/vmware/nsxt](http://registry.terraform.io/vmware/nsxt)
[...]

PR #1486, opened 2024-12-17
ImporterWrapper: Allocate a new Importer object (#1481)

salv-orlando:
The wrapper function was updating the existing importer, but the replaced function ended up having the same pointer as the wrapper function, meaning that import would enter an infinite recursion loop.

Tested by doing import, read, update, and delete for
1) nsxt_logical_tier1_router
2) nsxt_ip_pool

### DIFF
--- a/nsxt/removed_resource_wrapper.go
+++ b/nsxt/removed_resource_wrapper.go
@@ -66,15 +66,18 @@ func deleteWrapper(originalFunc schema.DeleteFunc, name string) schema.DeleteFun
 }
 
 func importerWrapper(originalImporter *schema.ResourceImporter, name string) *schema.ResourceImporter {
+	newImporter := new(schema.ResourceImporter)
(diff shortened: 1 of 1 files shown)


nsxt/removed_resource_wrapper.go
```diff
@@ -66,15 +66,18 @@ func deleteWrapper(originalFunc schema.DeleteFunc, name string) schema.DeleteFun
 }
 
 func importerWrapper(originalImporter *schema.ResourceImporter, name string) *schema.ResourceImporter {
+	newImporter := new(schema.ResourceImporter)
```
annakhm, line 69: would it be more straightforward to return `originalImporter` here, and let the `Read` fail?

salv-orlando: that was the previous code.
Probably you meant do not wrapping the importer at all and let import fail at read?

annakhm (approved):

PR #1489, opened 2024-12-23
Retrieve also system-defined hostswitch profile while creating transport nodes

ksamoray:
User may use system-defined hostswitch profiles while creating transport nodes, and should be retrieved as well as user defined hostswitch profiles.


salv-orlando (approved):
LGMT

PR #1490, opened 2024-12-24
Fix edge transport node documentation

ksamoray:
Edge transport node transport_zone and host_switch_id attributes can handle path parameters but it is better to avoid using it instead of id values. For now we only document this and will handle that within the code later on.


salv-orlando (approved):
LGTM

PR #1495, opened 2024-12-31
Provide informative import error message for path

annakhm:
Whenever bad path is provided for import, we need to communicate what the expected format is.

### DIFF
--- a/nsxt/policy_utils.go
+++ b/nsxt/policy_utils.go
@@ -27,6 +27,7 @@ import (
 // ErrNotAPolicyPath - Define an ignorable error for  policy path importer - to indicate that the given path is not a
 // policy path and may be processed as an id - which is handy for legacy import method
 var ErrNotAPolicyPath = errors.New("specified import identifier is not a policy path")
+var ErrUnexpectedPolicyPath = errors.New("unexpected policy path")
 var ErrEmptyImportID = errors.New("import identifier cannot be empty")
 
 func isSpaceString(s string) bool {
@@ -410,7 +411,33 @@ func validateImportPolicyPath(policyPath string) error {
 	return nil
 }
 
-// This importer function accepts policy path and resource ID
+func isValidResourceID(id string) bool {
+	if isSpaceString(id) {
+		return false
+	}
+	if len(id) > 1024 {
+		return false
+	}
+	if strings.Contains(id, "/") {
+		return false
+	}
+	return true
+}
+
+// This importer function accepts policy path only as import ID
+func getFriendlyPolicyPathResourceImporter(pathExample string) func(d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
+	return func(d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
+		rd, err := nsxtPolicyPathResourceImporterHelper(d, m)
+		if errors.Is(err, ErrNotAPolicyPath) || errors.Is(err, ErrUnexpectedPolicyPath) {
+			return rd, fmt.Errorf("Invalid policy path %s; expected format: %s", d.Id(), pathExample)
+		} else if err != nil {
+			return rd, err
+		}
+		return rd, nil
+	}
+}
+
+// This importer function accepts policy path or resource ID
 func nsxtPolicyPathResourceImporter(d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
 	rd, err := nsxtPolicyPathResourceImporterHelper(d, m)
 	if errors.Is(err, ErrNotAPolicyPath) {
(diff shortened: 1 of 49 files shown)


nsxt/policy_utils.go
```diff
@@ -421,21 +448,47 @@ func nsxtPolicyPathResourceImporter(d *schema.ResourceData, m interface{}) ([]*s
...
+func getMultitenancyPathExample(pathExample string) string {
+	return fmt.Sprintf("%s or /orgs/[org]/projects/[project]%s", pathExample, pathExample)
 }
 
-func nsxtVPCPathResourceImporter(d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
-	rd, err := nsxtPolicyPathResourceImporterHelper(d, m)
-	if err != nil {
-		return rd, err
+// This importer function accepts policy path or resource ID
+func getFriendlyPolicyPathOrIDResourceImporter(pathExample string) func(d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
```
salv-orlando, line 456: I have nothing against friendship, but since the old importers are removed perhaps we can just call this function getPolicyPathOrIdResourceImporter?

Similarly for the one above

nsxt/policy_utils.go
```diff
@@ -410,7 +411,33 @@ func validateImportPolicyPath(policyPath string) error {
...
+		return false
+	}
+	if strings.Contains(id, "/") {
+		return false
+	}
+	return true
+}
+
+// This importer function accepts policy path only as import ID
+func getFriendlyPolicyPathResourceImporter(pathExample string) func(d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
```
salv-orlando, line 428: I have no problem with the approach you followed, but why did you choose to create a function that returns an importer rather than defining directly the importer function?

nsxt/policy_utils.go
```diff
@@ -421,21 +448,47 @@ func nsxtPolicyPathResourceImporter(d *schema.ResourceData, m interface{}) ([]*s
...
+// This importer function accepts policy path or resource ID
+func getFriendlyPolicyPathOrIDResourceImporter(pathExample string) func(d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
+	return func(d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
+		rd, err := nsxtPolicyPathResourceImporterHelper(d, m)
+		if errors.Is(err, ErrUnexpectedPolicyPath) {
+			return rd, fmt.Errorf("Invalid policy path %s; expected format: %s", d.Id(), pathExample)
+		}
+		if errors.Is(err, ErrNotAPolicyPath) {
+			if !isValidResourceID(d.Id()) {
+				return rd, fmt.Errorf("Invalid policy path or id %s; expected path format: %s", d.Id(), pathExample)
```
salv-orlando, line 464: Do you think this error message makes sense in the case we fail to validate the resource id? The user did not specify a path.

nsxt/policy_utils.go
```diff
@@ -421,21 +448,47 @@ func nsxtPolicyPathResourceImporter(d *schema.ResourceData, m interface{}) ([]*s
...
+		rd, err := nsxtPolicyPathResourceImporterHelper(d, m)
+		if errors.Is(err, ErrNotAPolicyPath) || errors.Is(err, ErrUnexpectedPolicyPath) {
+			return rd, fmt.Errorf("Invalid policy path %s; expected format: %s", d.Id(), pathExample)
+		}
+
+		if err != nil {
+			return rd, err
+		}
+
+		projectID, vpcID := getContextDataFromSchema(d)
```
salv-orlando, line 486: For code readability, can you add a comment stating that we check the context here because it's supposed to be filled by nsxtPolicyPathResourceImporterHelper

nsxt/resource_nsxt_policy_dhcp_relay.go
```diff
@@ -22,7 +22,7 @@ func resourceNsxtPolicyDhcpRelayConfig() *schema.Resource {
 		Update: resourceNsxtPolicyDhcpRelayConfigUpdate,
 		Delete: resourceNsxtPolicyDhcpRelayConfigDelete,
 		Importer: &schema.ResourceImporter{
-			State: nsxtPolicyPathResourceImporter,
+			State: getFriendlyPolicyPathOrIDResourceImporter(getMultitenancyPathExample("/infra/dhcp-relays/[dhcp-relay]")),
```
salv-orlando, line 25: I recall we discussed you would look into extracting the example path from the API spec. Was that for generation?

nsxt/policy_utils.go
salv-orlando: Clarified live with Anna.

nsxt/policy_utils.go
annakhm: Makes sense, done

nsxt/policy_utils.go
annakhm: There is little sense to give an example of id.. but I rephrased it to avoid confusion, does it look better now?

nsxt/resource_nsxt_policy_dhcp_relay.go
annakhm: Correct, generator will fill this up

nsxt/resource_nsxt_vpc_static_routes.go
```diff
@@ -82,7 +82,7 @@ func resourceNsxtVpcStaticRoutes() *schema.Resource {
 		Update: resourceNsxtVpcStaticRoutesUpdate,
 		Delete: resourceNsxtVpcStaticRoutesDelete,
 		Importer: &schema.ResourceImporter{
-			State: nsxtVPCPathResourceImporter,
+			State: getVpcPathResourceImporter("/orgs/[org]/projects/[project]/vpcs/[vpc]/static-routes/[route]"),
```
ksamoray, line 85: Having these long constant strings within the function call is quite terrible.
I wouldn't ask you to change all of those as it's a pain but maybe it's worthwhile to embed these in the metadata struct somehow or use a constant (especially as we plan to generate this as well).

ksamoray:
LGTM, other than the comment above.

nsxt/resource_nsxt_vpc_static_routes.go
annakhm: My personal preference is how the code is written today, I would argue its more readable this way, since you don't need to scroll up to see the example. However, I'm fine with rewriting this with a constant. I think we should have same approach for current PR and future generated code.

salv-orlando: I think that generated code is not supposed to be readable. It will be good to have constants, but if it takes time to implement that, there are worse thing that a long literal - like 500+ files in the same directory :)

salv-orlando (approved):
LGTM

nsxt/resource_nsxt_vpc_static_routes.go
annakhm: Done

ksamoray:
@annakhm please rebase this so tests can run

nsxt/resource_nsxt_policy_transport_zone.go
```diff
@@ -19,14 +19,16 @@ import (
...
+
 func resourceNsxtPolicyTransportZone() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceNsxtPolicyTransportZoneCreate,
 		Read:   resourceNsxtPolicyTransportZoneRead,
 		Update: resourceNsxtPolicyTransportZoneUpdate,
 		Delete: resourceNsxtPolicyTransportZoneDelete,
 		Importer: &schema.ResourceImporter{
-			State: resourceNsxtPolicyTransportZoneImporter,
+			State: getPolicyPathOrIDResourceImporter(transportZonePathSample),
```
ksamoray, line 31: This importer won't work here, need that one which gathers `site_id`

ksamoray (commented):
Can we also remove `nsxtPolicyPathResourceImporter` as it isn't used anymore?

nsxt/resource_nsxt_policy_transport_zone.go
annakhm: correct, my bad! thanks Kobi

PR #1498, opened 2025-01-08
Support IPv6 in TNP and edge node resources

annakhm:


nsxt/resource_nsxt_edge_transport_node.go
```diff
@@ -1474,7 +1699,12 @@ func setHostSwitchSpecInSchema(d *schema.ResourceData, spec *data.StructValue, n
...
 			}
 			var err error
-			elem["ip_assignment"], err = setIPAssignmentInSchema(sw.IpAssignmentSpec)
+			if sw.IpAssignmentSpec != nil {
+				elem["ip_assignment"], err = setIPAssignmentInSchema(sw.IpAssignmentSpec)
+			}
+			if sw.Ipv6AssignmentSpec != nil {
+				elem["ipv6_assignment"], err = setIPv6AssignmentInSchema(sw.Ipv6AssignmentSpec)
+			}
 			if err != nil {
```
ksamoray, line 1708: I'd rather check err after 1073, 1076 lines - as err from `setIPAssignmentInSchema` could be overwritten by `setIPv6AssignmentInSchema`

nsxt/resource_nsxt_edge_transport_node.go
```diff
@@ -1140,6 +1368,7 @@ func getHostSwitchSpecFromSchema(d *schema.ResourceData, m interface{}, nodeType
 			return nil, err
 		}
 		iPAssignmentSpec, err := getIPAssignmentFromSchema(swData["ip_assignment"])
```
ksamoray, line 1370: Check err after `getIPAssignmentFromSchema` call

nsxt/resource_nsxt_edge_transport_node.go
annakhm: Thanks!

nsxt/resource_nsxt_edge_transport_node.go
```diff
@@ -1016,7 +1136,7 @@ func getIPAssignmentFromSchema(ipAssignmentList interface{}) (*data.StructValue,
 	converter := bindings.NewTypeConverter()
 
 	for _, ia := range ipAssignmentList.([]interface{}) {
-		iaType, iaData, err := getIPAssignmentData(ia.(map[string]interface{}))
+		iaType, iaData, err := getIPAssignmentData(ia.(map[string]interface{}), ipAssignmentTypes)
```
salv-orlando, line 1139: I realized that IPAssignmentSpec also supports StaticIpMacListSpec, now we're adding it to the v6 assignementSpec, but the v4 one does not have it. Probably we should add it.

annakhm: Good point, added `static_ip_mac` option to ipv4, and also the missing `no_ipv4` option

ksamoray (approved):

PR #1504, opened 2025-01-15
Support service entries in policy rules

annakhm:
This follows API change on NSX that allows to specify explicit service entries on rule, as an alternative to service path.


salv-orlando (approved):
LGTM!

PR #1508, opened 2025-01-17
Fix panic in compute manager

annakhm:
This change protects against nil value when empty credential block is provided.


salv-orlando (approved):
LGTM

PR #1512, opened 2025-01-22
Replace host switch profiles attr with typed ones

annakhm:
Deprecate list of host switch profiles, and introduce typed profile attributes instead - `uplink_profile` and `vtep_ha_profile`. This is because certain profile types are autoassigned by NSX, in which case permadiff is introduced, unless the user fixes it by adding the auto-assigned profile into resource intent.
Having type-specific computed attributes solves this problem.


ksamoray (approved):
Change LGTM, we gotta cover this with testing for the impacted resources though

ksamoray:
I've tested this PR, with `nsxt_edge_transport_node` and `nsxt_policy_host_transport_node_profile`, it works OK. I have a doubt whether we should support more than one profile of a type, in any of these resources.
I'll raise this question.

PR #1514, opened 2025-01-22
Address linter error

ksamoray:
Some old lint issue caught here:
nsxt/data_source_nsxt_policy_bridge_profile_test.go


ksamoray:
Addressed in #1513

PR #1517, opened 2025-01-25
Avoid recreating gateway when locale service added

annakhm:
This change removes ForceNew from nsx_id attribute in locale service schema, in order to avoid recreating the gateway when locale service is added.
Without this fix, adding locale_service within the gateway would trigger recreation even when nsx_id is not specified, since nsx_is is defined as Computed.


salv-orlando (approved):
LGTM

ksamoray (approved):

ksamoray (approved):

annakhm:
We need to add an update test with interfaces present - not merging this yet.

Issue #1474, opened 2024-11-29
L7 Access Profile creation


m-j-gardner:
### Is your feature request related to a problem? Please describe.

This is more of a question than a feature request at this point.

I need to create a L7 Access Profile for a T1 Gateway but can't find a suitable resource.

Is the capability already in the provider and I'm missing something?

### Describe the solution you'd like

I'm hoping there's already a solution and I'm just looking in the wrong places

### Describe alternatives you've considered

I thought nsxt_policy_context_profile might be dual purpose but there's no mention of L7 in it's documentation.

ksamoray:
Hi @m-j-gardner
Unfortunately, as of now we have no implementation for this object.

m-j-gardner:
@ksamoray  .. thank you for your prompt reply.

Do I need to do something to request that this is added to future releases?  It won't help me but hopefully will be there for the next person :-)

ksamoray:
@m-j-gardner all the required info is available above, thanks for reporting!

PR #1519, opened 2025-01-27
Implement L7AccessProfile resource

ksamoray:
Fixes: #1474


nsxt/resource_nsxt_policy_l7_access_profile_test.go
```diff
@@ -0,0 +1,369 @@
...
+	return fmt.Sprintf(`
+resource "nsxt_policy_l7_access_profile" "test" {
+%s
+  display_name = "%s"
+  description  = "%s"
+  default_action_logged = %s
+  default_action = "%s"
+
+  l7_access_entry {
+    nsx_id = "%s"
```
annakhm, line 300: for better coverage we need tests without nsx_id, and ideally with a mix (some entries with id, some without)

website/docs/r/policy_l7_access_profile.html.markdown
```diff
@@ -0,0 +1,116 @@
...
+* `l7_access_entry` - (Optional) Property containing L7 access entries for Policy L7 Access Profile.
+    * `nsx_id` - (Optional) The NSX ID of this L7 access entry. If set, this ID will be used to create the resource.
+    * `display_name` - (Required) Display name of the L7 access entry.
+    * `description` - (Optional) Description of the L7 access entry.
+    * `action` - (Optional) The action to be applied to all the services.  Valid values are `ALLOW`, `REJECT`, `REJECT_WITH_RESPONSE`.
+    * `attribute` - (Optional) Property containing attributes/sub-attributes for Policy L7 Access Profile.
+        * `attribute_source` - (Optional) Source of attribute value i.e whether system defined or custom value. Valid values are `SYSTEM`, `CUSTOM`.
+        * `custom_url_partial_match` - (Optional) True value for this flag will be treated as a partial match for custom url.
+        * `description` - (Optional) Description for attribute value.
+        * `key` - (Optional) URL_Reputation is currently not available. Please do not use it in Attribute Key while creating context profile.
```
salv-orlando, line 83: This is likely generated from the API spec, but it does not seem to make much sense.

Perhaps we might want to change it to something like

Key for attribute. Supported Attribute Keys are APP_ID, URL_CATEGORY, CUSTOM_URL

(we don't mention URL_REPUTATION since it's not supported)

salv-orlando (commented):
Just a minor comments about documentation.
Otherwise looks good.

annakhm (approved):

nsxt/resource_nsxt_policy_l7_access_profile_test.go
```diff
@@ -0,0 +1,362 @@
...
+
+func TestAccResourceNsxtPolicyL7AccessProfile_basic(t *testing.T) {
+	testAccResourceNsxtPolicyL7AccessProfileBasic(t, false, func() {
+		testAccPreCheck(t)
+	})
+}
+
+func TestAccResourceNsxtPolicyL7AccessProfile_multitenancy(t *testing.T) {
+	testAccResourceNsxtPolicyL7AccessProfileBasic(t, true, func() {
+		testAccPreCheck(t)
```
annakhm, line 79: We need to prevent this from running on Global Manager

PR #1527, opened 2025-02-05
Replace host switch profiles attr with typed ones

annakhm:
Deprecate list of host switch profiles, and introduce typed profile attributes instead - `uplink_profile` and `vtep_ha_profile`. This is because certain profile types are autoassigned by NSX, in which case permadiff is introduced, unless the user fixes it by adding the auto-assigned profile into resource intent.
Having type-specific computed attributes solves this problem.

This PR contains the fix for transport_node_profile_sub_config which was missing in the previous PR

### DIFF
--- a/nsxt/resource_nsxt_edge_transport_node.go
+++ b/nsxt/resource_nsxt_edge_transport_node.go
@@ -447,8 +447,20 @@ func getStandardHostSwitchSchema(nodeType string) *schema.Schema {
 					Computed:    true,
 				},
 				"host_switch_profile": getHostSwitchProfileIDsSchema(),
-				"ip_assignment":       getIPAssignmentSchema(false),
-				"ipv6_assignment":     getIPv6AssignmentSchema(),
+				"uplink_profile": {
+					Type:        schema.TypeString,
+					Description: "Host switch uplink profile",
+					Optional:    true,
+					Computed:    true,
+				},
+				"vtep_ha_profile": {
+					Type:        schema.TypeString,
+					Description: "Host switch high availability profile",
+					Optional:    true,
+					Computed:    true,
+				},
+				"ip_assignment":   getIPAssignmentSchema(false),
+				"ipv6_assignment": getIPv6AssignmentSchema(),
 				"pnic": {
 					Type:        schema.TypeList,
 					Optional:    true,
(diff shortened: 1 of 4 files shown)


nsxt/resource_nsxt_edge_transport_node.go
```diff
@@ -447,8 +447,20 @@ func getStandardHostSwitchSchema(nodeType string) *schema.Schema {
 					Computed:    true,
 				},
 				"host_switch_profile": getHostSwitchProfileIDsSchema(),
-				"ip_assignment":       getIPAssignmentSchema(false),
-				"ipv6_assignment":     getIPv6AssignmentSchema(),
+				"uplink_profile": {
```
salv-orlando, line 450: Should uplink_profile and vtep_ha_profile have a conflictsWith with hostSwitchProfile?

nsxt/resource_nsxt_edge_transport_node.go
```diff
@@ -2007,12 +2055,23 @@ func setIPv6AssignmentInSchema(spec *data.StructValue) (interface{}, error) {
...
+		}
+		if *hspID.Key == mpmodel.BaseHostSwitchProfile_RESOURCE_TYPE_VTEPHAHOSTSWITCHPROFILE {
+			parentMap["vtep_ha_profile"] = hspID.Value
+		}
 		hostSwitchProfileIDs = append(hostSwitchProfileIDs, hspID.Value)
 	}
-	return hostSwitchProfileIDs
+
+	// deprecated way of specifying profiles in single list
+	parentMap["host_switch_profile"] = hostSwitchProfileIDs
```
salv-orlando, line 2074: With this logic, the same profile is being added both to host_switch_profile and to the profile_specific_attributes.

This seems safe, but also made me think that we will keep adding the auto-assigned uplink profile. Why this does not cause a permadiff?

nsxt/resource_nsxt_edge_transport_node.go
```diff
@@ -1147,8 +1173,36 @@ func getHostSwitchProfileResourceType(m interface{}, id string) (string, error)
...
+		profileType := mpmodel.BaseHostSwitchProfile_RESOURCE_TYPE_VTEPHAHOSTSWITCHPROFILE
+		elem := mpmodel.HostSwitchProfileTypeIdEntry{
+			Key:   &profileType,
+			Value: &haProfile,
+		}
+		hswProfiles = append(hswProfiles, elem)
+	}
+
+	if len(hswProfiles) > 0 {
+		return hswProfiles, nil
```
salv-orlando, line 1201: I think this return is tantamount to ignoring host_switch_profiles is the profile specific attributes were set.
In that case why not use conflictsWith?

nsxt/resource_nsxt_edge_transport_node.go
annakhm: Unfortunately `ConflictsWith` [does not work](https://github.com/hashicorp/terraform-plugin-sdk/blob/4d5e3b9a3211f139b41e78ea9ee880c8bf9bb87f/helper/schema/schema.go#L262) with nested schema.
This should be solved in plugin framework.

nsxt/resource_nsxt_edge_transport_node.go
annakhm: `host_switch_profile` is Computed, so if the user switched to the new way of configuring profiles, there will be no permadiff. 
If the user keeps using `host_switch_profile` for configuring one profile type, while the other type is being auto-assigned, the permadiff will occur. Thats why we're deprecating `host_switch_profile` and recommending to use the new attributes.

nsxt/resource_nsxt_edge_transport_node.go
annakhm: Answered above

ksamoray (approved):

PR #1528, opened 2025-02-07
Support LB server ssl profile resource

annakhm:


nsxt/resource_nsxt_policy_lb_server_ssl_profile.go
```diff
@@ -0,0 +1,308 @@
...
+			SdkFieldName: "CipherGroupLabel",
+		},
+	},
+	"ciphers": {
+		Schema: schema.Schema{
+			Type: schema.TypeList,
+			Elem: &metadata.ExtendedSchema{
+				Schema: schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validation.StringInSlice(lbServerSslProfileCiphersValues, false),
```
salv-orlando, line 97: The list of ciphers varies with NSX release. It might be safer to remove the validation function.
But as we have the same for client_ssl_ciphers, perhaps it does not matter.

For consistency, we can also remove validation for cipher groups, even if that list is pretty static

website/docs/r/policy_lb_server_ssl_profile.html.markdown
```diff
@@ -0,0 +1,74 @@
...
+
+The following arguments are supported:
+
+* `display_name` - (Required) Display name of the resource.
+* `description` - (Optional) Description of the resource.
+* `tag` - (Optional) A list of scope + tag pairs to associate with this resource.
+* `nsx_id` - (Optional) The NSX ID of this resource. If set, this ID will be used to create the resource.
+* `cipher_group_label` - (Required) One of `BALANCED`, `HIGH_SECURITY`, `HIGH_COMPATIBILITY` or `CUSTOM`.
+* `ciphers` - (Optional) Supported SSL cipher list. Can only be specified if `cipher_group_label` is set to `CUSTOM`, otherwise the list is assigned by NSX.
+* `protocols` - (Optional) In newer versions of NSX, only TLS1.2 is supported.
```
salv-orlando, line 47: This was taken from the OpenAPI spec, but maybe for Terraform docs users would like to know what "newer" means. I can find that out.

nsxt/resource_nsxt_policy_lb_server_ssl_profile.go
annakhm: That's a good point, we should update the list with every release (hopefully the process will be automatic)
I'm also wondering whether we should include the whole list in the docs..

website/docs/r/policy_lb_server_ssl_profile.html.markdown
annakhm: I actually wrote that one :) Depending on how old ago non-TLS1.2 was removed from NSX, we might want to consider removing this attribute altogether.

nsxt/resource_nsxt_policy_lb_server_ssl_profile.go
annakhm: I don't think so

website/docs/r/policy_lb_server_ssl_profile.html.markdown
salv-orlando: The "newer versions of NSX" start with 4.2

nsxt/resource_nsxt_policy_lb_server_ssl_profile.go
annakhm: I removed the validation for this resource. Will remove the same validation in other resources in a separate PR.

ksamoray (approved):
LGTM, once @salv-orlando is happy with it we can merge

salv-orlando (approved):
LGTM

PR #1538, opened 2025-02-15
Remove Computed from gateway edge cluster path

annakhm:
When attribute is Computed, clearing the attribute is not flagged by terraform SDK as change in intent. This means user can not clear edge cluster path and convert gateway to Distributed Only mode.

In addition, fix log printouts.


salv-orlando (approved):
LGTM

PR #1539, opened 2025-02-15
Remove validations for TLS ciphers

annakhm:
Cipher list can change frequently, rather than keeping up in the provider, we let NSX fail the validation.


salv-orlando (approved):
LGTM

PR #1542, opened 2025-02-17
Document undocumented attributes of nsxt_upgrade_run

ksamoray:
The upgrade run resource has some attributes which control the wait for the upgrade execution. However these aren't documented, and could be handy.

Also, change default initial delay to 5 minutes.


salv-orlando (approved):
LGTM

annakhm (approved):

salv-orlando (approved):
LGTM

PR #1543, opened 2025-02-17
Handle upgrade run crash, log returned error

ksamoray:
Upgrade run crashes while waiting for state, we should return nil when an error is returned.
To improve troubleshooting, added a few debug logs, and return nil explicitly when an error is raised.

### DIFF
--- a/nsxt/resource_nsxt_upgrade_run.go
+++ b/nsxt/resource_nsxt_upgrade_run.go
@@ -546,6 +546,7 @@ type upgradeStatusAndDetail struct {
 func getUpgradeStatus(statusClient upgrade.StatusSummaryClient, component *string) (*upgradeStatusAndDetail, error) {
 	status, err := statusClient.Get(component, nil, nil)
 	if err != nil {
+		log.Printf("[DEBUG] Failed to get status: error is %v", err)
 		return nil, err
 	}
 	if component == nil {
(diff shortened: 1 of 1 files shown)


nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -546,6 +546,7 @@ type upgradeStatusAndDetail struct {
 func getUpgradeStatus(statusClient upgrade.StatusSummaryClient, component *string) (*upgradeStatusAndDetail, error) {
 	status, err := statusClient.Get(component, nil, nil)
 	if err != nil {
+		log.Printf("[DEBUG] Failed to get status: error is %v", err)
```
salv-orlando, line 549: Could this also be at INFO or WARNING level?

nsxt/resource_nsxt_upgrade_run.go
```diff
@@ -576,7 +577,7 @@ func waitUpgradeForStatus(upgradeClientSet *upgradeClientSet, component *string,
 				return nil, model.ComponentUpgradeStatus_STATUS_IN_PROGRESS, nil
 			}
 			if err != nil {
-				return status, model.ComponentUpgradeStatus_STATUS_FAILED, err
+				return nil, model.ComponentUpgradeStatus_STATUS_FAILED, err
```
salv-orlando, line 580: Are we sure the setting ComponentUpgradeStatus_STATUS_FAILED is the right thing to do here? Since we were not able to retrieve the status at all, we probably don't know what's the actual status

salv-orlando (approved):
LGTM

PR #1545, opened 2025-02-19
Fix build failures

ksamoray:
Build requires make fmt to pass, fix some issue with resource_nsxt_policy_lb_virtual_server.go


salv-orlando (approved):
LGTM

PR #1547, opened 2025-02-19
Fix invalid retry configuration in provider

annakhm:
If retry max interval is smaller than min interval, today provider crashes. In case of such invalid configuration, provider would assume max interval to be min interval + 1.


salv-orlando (approved):
LGTM!

ksamoray (approved):

PR #1553, opened 2025-02-25
Implement nsxt_policy_edge_cluster resource

ksamoray:


nsxt/resource_nsxt_policy_edge_cluster.go
```diff
@@ -0,0 +1,406 @@
...
+		Metadata: metadata.Metadata{
+			SchemaType:   "list",
+			SdkFieldName: "PolicyEdgeNodes",
+			ReflectType:  reflect.TypeOf(model.PolicyEdgeClusterMember{}),
+		},
+	},
+	"password_managed_by_vcf": {
+		Schema: schema.Schema{
+			Type:     schema.TypeBool,
+			Optional: true,
```
annakhm, line 141: API defines default `false` here

nsxt/resource_nsxt_policy_edge_cluster.go
```diff
@@ -0,0 +1,406 @@
...
+			Computed: true,
+		},
+		Metadata: metadata.Metadata{
+			SdkFieldName: "InterSiteForwardingEnabled",
+		},
+	},
+	"edge_cluster_profile": {
```
annakhm, line 66: `edge_cluster_profile_path` for consistency?

nsxt/resource_nsxt_policy_edge_cluster.go
```diff
@@ -0,0 +1,406 @@
...
+	"policy_edge_node": {
+		Schema: schema.Schema{
+			Type: schema.TypeList,
+			Elem: &metadata.ExtendedResource{
+				Schema: map[string]*metadata.ExtendedSchema{
+					"edge_transport_node_path": {
+						Schema: schema.Schema{
+							Type:         schema.TypeString,
+							ValidateFunc: validatePolicyPath(),
+							Required:     true,
```
annakhm, line 109: I see that `Edge transport node path can't be changed once member is added to cluster.` But I assume those can be added/removed? In which case `ForceNew` might not be applicable?

nsxt/resource_nsxt_policy_edge_cluster.go
```diff
@@ -0,0 +1,406 @@
...
+			Type:         schema.TypeString,
+			Optional:     true,
+			ValidateFunc: validatePolicyPath(),
+		},
+		Metadata: metadata.Metadata{
+			SchemaType:   "string",
+			SdkFieldName: "EdgeClusterProfile",
+		},
+	},
+	"allocation_rule": {
```
annakhm, line 77: If we are rewriting this part manually, maybe just offer the boolean flag? (I'm guessing multiple rules are not supported)

website/docs/r/policy_edge_cluster.html.markdown
```diff
@@ -0,0 +1,62 @@
+---
+subcategory: "FIXME"
```
annakhm, line 2: Beta

nsxt/resource_nsxt_policy_edge_cluster.go
ksamoray: IDK which implications `ForceNew` will imply - e.g, if I have a tier1 which is allocated on any node within the cluster, and `ForceNew` will attempt to remove the whole cluster, I'm guessing that this transaction will fail, right?

There is some API for Edge node replacement, `/policy/api/v1/infra/sites/{site-id}/enforcement-points/{enforcementpoint-id}/edge-clusters/{edge-cluster-id}/action/replace-edge-transport-node`

But IDK how we could leverage this in the provider...

annakhm: Probably best to leave as is

nsxt/resource_nsxt_policy_edge_cluster.go
ksamoray: My assumption was that if this complexity was added in the API, then they might had some other rule functionality further on, so I left it as is. If you don't expect any, we can replace this with a boolean.

annakhm (approved):

PR #1554, opened 2025-02-25
Implement nsxt_policy_edge_high_availability_profile resource

ksamoray:


website/docs/r/policy_edge_high_availability_profile.html.markdown
```diff
@@ -0,0 +1,61 @@
+---
+subcategory: "FIXME"
```
annakhm, line 2: `Beta` or `Fabric`

nsxt/resource_nsxt_policy_edge_high_availability_profile.go
```diff
@@ -0,0 +1,298 @@
...
+			SchemaType: "string",
+			Skip:       true,
+		},
+	},
+	"bfd_probe_interval": {
+		Schema: schema.Schema{
+			Type:     schema.TypeInt,
+			Optional: true,
+			Default:  500,
+		},
```
annakhm, line 59: MP resource has more refined validations - f.e, for this attribute we have `validation.IntBetween(50, 60000)`. Policy spec also specifies those - I think we should be consistent with MP resource here

website/docs/r/policy_edge_high_availability_profile.html.markdown
ksamoray: I think that all those policy edge should be beta, right?

nsxt/resource_nsxt_policy_edge_high_availability_profile.go
ksamoray: I'm a bit confused as so far we were reluctant to validate such things in the provider, and left those to NSX.
As this resource is auto-generated, maybe we should add these validations in the generator? Those ranges are available in the APi spec...

website/docs/r/policy_edge_high_availability_profile.html.markdown
annakhm: If we expect those to be fully tested by release, then we can go with `Fabric` right away

nsxt/resource_nsxt_policy_edge_high_availability_profile.go
annakhm: I was just pointing out inconsistency with MP resource, not, not insisting on the limits

annakhm (approved):

PR #1562, opened 2025-03-10
Improve validation for LB pool resource

annakhm:
IP addresses in SNAT can only be specified for certain snat type


salv-orlando (approved):
LGTM

PR #1567, opened 2025-03-13
Adding a data source for retrieving the tags based on a scope

aruntony005:
Adding a data source for retrieving the tags based on a scope

### DIFF
--- a/nsxt/data_source_nsxt_policy_tags.go
+++ b/nsxt/data_source_nsxt_policy_tags.go
@@ -1,0 +1,74 @@
+// © Broadcom. All Rights Reserved.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MPL-2.0
+
+package nsxt
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+)
+
+func dataSourceNsxtTags() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNsxtTagsRead,
+		Schema: map[string]*schema.Schema{
+			"scope": getRequiredStringSchema("The scope of the tags to retrieve."),
+			"items": {
+				Type:        schema.TypeList,
+				Description: "List of tags based on the scope.",
+				Computed:    true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+		},
+	}
+}
+
+func dataSourceNsxtTagsRead(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+	scope := d.Get("scope").(string)
+	tagsList, err := listTags(connector, scope)
+	if err != nil {
+		return err
+	}
+
+	var tags []string
+	for _, tag := range tagsList {
+		tags = append(tags, *tag.Tag)
+	}
+	d.Set("items", tags)
+
+	d.SetId(newUUID())
+
+	return nil
+}
+
+func listTags(connector client.Connector, scope string) ([]model.TagInfo, error) {
+	client := infra.NewTagsClient(connector)
+	if client == nil {
+		return nil, policyResourceNotSupportedError()
+	}
+	var cursor *string
+	var tagsListResults []model.TagInfo
+	total := 0
+	for {
+		tagsList, err := client.List(cursor, nil, nil, nil, nil, &scope, nil, nil, nil, nil)
+		if err != nil {
+			return nil, err
+		}
+		tagsListResults = append(tagsListResults, tagsList.Results...)
+		if total == 0 && tagsList.ResultCount != nil {
+			total = int(*tagsList.ResultCount)
+		}
+
+		cursor = tagsList.Cursor
+		if len(tagsListResults) >= total {
+			return tagsListResults, nil
+		}
+	}
+
+}
--- a/nsxt/data_source_nsxt_policy_tags_test.go
+++ b/nsxt/data_source_nsxt_policy_tags_test.go
@@ -1,0 +1,87 @@
+// © Broadcom. All Rights Reserved.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MPL-2.0
+
+package nsxt
+
+import (
+	"fmt"
+	"regexp"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccDataSourceNsxtPolicyTags_basic(t *testing.T) {
+	tagName := "testTag"
+	emptyScopeTag := "testEmptyScopeTag"
+	transportZone := getOverlayTransportZoneName()
+	re, _ := regexp.Compile(`.*EmptyScope.*`)
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccOnlyLocalManager(t)
+		},
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccNSXPolicyTagsReadTemplate(tagName, emptyScopeTag, transportZone),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckOutput("nsxt_tags", tagName),
+					resource.TestMatchOutput("empty_nsxt_tags", re),
+					resource.TestCheckOutput("wildcard_nsxt_tags", tagName),
+				),
+			},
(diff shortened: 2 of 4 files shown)


nsxt/data_source_nsxt_policy_tags.go
```diff
@@ -0,0 +1,50 @@
...
+	tagsList, err := client.List(nil, nil, nil, nil, nil, nil, nil, nil, nil, nil)
+	if err != nil {
+		return err
+	}
+
+	scopeFilter, isSet := d.GetOk("scope")
+
+	var tags []string
+	for _, tag := range tagsList.Results {
+		if !isSet || *tag.Scope == scopeFilter {
```
annakhm, line 41: I think tag.Scope can be nil if omitted, so lets protect against null pointer exception here

aruntony005: The tag.Scope value of the result has a non nil empty string if the scope is omitted. Also with the new change I am handling the filtering in the client.List() function.

nsxt/data_source_nsxt_policy_tags_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccDataSourceNsxtPolicyTags_basic(t *testing.T) {
+	testAccDataSourceNsxtPolicyTagsBasic(t, func() {
```
ksamoray, line 15: We use this pattern with an internal func when we reuse the test func for other purposes, e.g multitenancy.
For simplicity you can discard `testAccDataSourceNsxtPolicyTagsBasic` and use this pattern as seen [here](https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/data_source_nsxt_manager_cluster_node_test.go#L14).

nsxt/data_source_nsxt_policy_tags_test.go
```diff
@@ -0,0 +1,64 @@
...
+				Config: testAccNSXPolicyTagsReadTemplate(tagName, transportZone),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckOutput("nsxt_tags", tagName),
+				),
+			},
+		},
+	})
+}
+
+func testAccNSXPolicyTagsReadTemplate(tagName string, transportZone string) string {
```
ksamoray, line 37: Can we add a test where scope is not specified? As this is supported by the data source?

nsxt/data_source_nsxt_policy_tags.go
```diff
@@ -0,0 +1,47 @@
...
+				},
+			},
+		},
+	}
+}
+
+func dataSourceNsxtTagsRead(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+	client := infra.NewTagsClient(connector)
+	scope := d.Get("scope").(string)
```
ksamoray, line 32: `d.Get()` will return a null string in this case. I'm not sure how `clilent.List()` will behave in this case, which we do not cover with testing.
Maybe it's better to make `scope` a pointer and assign a value to it when `d.Get()` has a value.

nsxt/data_source_nsxt_policy_tags_test.go
aruntony005: Changed the pattern

nsxt/data_source_nsxt_policy_tags_test.go
aruntony005: Added a test for empty scope case.

nsxt/data_source_nsxt_policy_tags.go
```diff
@@ -0,0 +1,47 @@
...
+	tagsList, err := client.List(nil, nil, nil, nil, nil, &scope, nil, nil, nil, nil)
```
annakhm, line 33: I think we should support pagination here, especially since scope can be regex

annakhm:
I noticed there is [support for startsWith, endsWith, contains](https://github.com/vmware/vsphere-automation-sdk-go/blob/master/services/nsxt/infra/TagsClient.go#L23) in the API - lets include this in the docs

nsxt/data_source_nsxt_policy_tags.go
aruntony005: client.List() is handling this. Also I have made scope a "required" argument. So if scope is not defined it will fail in the validations.

nsxt/data_source_nsxt_policy_tags_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccDataSourceNsxtPolicyTags_basic(t *testing.T) {
+	tagName := "testTag"
+	emptyScopeTag := "testEmptyScopeTag"
+	transportZone := getOverlayTransportZoneName()
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
```
ksamoray, line 20: Can this run against a GM setup? If not we should check this here. @annakhm might know.

website/docs/d/policy_tags.html.markdown
```diff
@@ -0,0 +1,29 @@
+---
+subcategory: "Beta"
+layout: "nsxt"
+page_title: "NSXT: policy_tags"
+description: Tags data source.
+---
+
+# nsxt_policy_service
+
+This data source provides the list of tags with a particular scope.
```
ksamoray, line 10: Usually we list here which NSX platforms are supported, IDK if this is relevant here as tags are supported everywhere.

ksamoray (commented):
Code LGTM, a couple of things which @annakhm should maybe give feedback about.

nsxt/data_source_nsxt_policy_tags_test.go
annakhm: tags API is supported for GM

nsxt/data_source_nsxt_policy_tags.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	d.Set("items", tags)
+
+	d.SetId(newUUID())
+
+	return nil
+}
+
+func listTags(connector client.Connector, scope string) ([]model.TagInfo, error) {
+	client := infra.NewTagsClient(connector)
```
annakhm, line 51: This would limit support to Local Manager only. We have two options here:
1. Extend support for Global Manager as well, preferably by adding tags.go [to abstraction layer ](https://github.com/vmware/terraform-provider-nsxt/tree/master/api/infra)
2. Support Local Manager only, in which case we need to document this data source as only applicable to local manager, and also add test directive `testAccOnlyLocalManager(t)` as Kobi suggested

ksamoray:
@aruntony005 there are failures on the CI acceptance tests, e.g for 3.2.5:
```
=== CONT  TestAccDataSourceNsxtPolicyTags_basic
    data_source_nsxt_policy_tags_test.go:18: Step 1/1 error: Error running apply: exit status 1
        
        Error: Invalid index
        
          on terraform_plugin_test.tf line 27, in output "nsxt_tags":
          27:   value = data.nsxt_policy_tags.tags.items[0]
            â”œâ”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€
            â”‚ data.nsxt_policy_tags.tags.items is empty list of string
        
        The given key does not identify an element in this collection value: the
        collection has no elements.
        
        Error: Invalid index
        
          on terraform_plugin_test.tf line 36, in output "empty_nsxt_tags":
          36:   value = data.nsxt_policy_tags.emptytags.items[0]
            â”œâ”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€
            â”‚ data.nsxt_policy_tags.emptytags.items is empty list of string
        
        The given key does not identify an element in this collection value: the
        collection has no elements.
        
        Error: Invalid index
        
          on terraform_plugin_test.tf line 45, in output "wildcard_nsxt_tags":
          45:   value = data.nsxt_policy_tags.wildcardscope.items[0]
            â”œâ”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€
            â”‚ data.nsxt_policy_tags.wildcardscope.items is empty list of string
        
        The given key does not identify an element in this collection value: the
        collection has no elements.
--- FAIL: TestAccDataSourceNsxtPolicyTags_basic (4.51s)
```

LMK if you need help with this.

nsxt/data_source_nsxt_policy_tags_test.go
aruntony005: Added the local manager preCheck function to the test.

nsxt/data_source_nsxt_policy_tags.go
aruntony005: Added the local manager preCheck function to the test.

website/docs/d/policy_tags.html.markdown
aruntony005: Have updated the documentation.

annakhm (approved):

PR #1574, opened 2025-03-17
Update SDK to fix special char issue

annakhm:
The runtime update addresses the issue of data source search based on display name that contains the "&" character.


ksamoray (approved):
LGTM

PR #1590, opened 2025-04-02
Support ip block quota resource

annakhm:
Also allow specifying quotas in project and vpc resources


nsxt/resource_nsxt_policy_ip_block_quota_test.go
```diff
@@ -0,0 +1,254 @@
...
+  cidr         = "fc7e:0011:da43::0/48"
+}
+
+resource "nsxt_policy_ip_block_quota" "test" {
+  %s
+  display_name = "%s"
+  description  = "%s"
+
+  quota {
+    ip_block_paths        = [nsxt_policy_ip_block.%s.path]
```
salv-orlando, line 221: It's interesting that the API allows us to change the IP blocks a quota is applied to.

nsxt/resource_nsxt_policy_ip_block_quota_test.go
```diff
@@ -0,0 +1,254 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+var accTestPolicyIpBlockQuotaCreateAttributes = map[string]string{
+	"display_name":          getAccTestResourceName(),
+	"description":           "terraform created",
+	"single_ip_cidrs":       "2",
+	"ip_block_visibility":   "PRIVATE",
+	"ip_block_address_type": "IPV4",
+	"mask":                  "",
```
salv-orlando, line 21: It might be worth ensuring non-empty masks are also covered in tests

nsxt/resource_nsxt_policy_ip_block_quota_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+var accTestPolicyIpBlockQuotaCreateAttributes = map[string]string{
+	"display_name":          getAccTestResourceName(),
+	"description":           "terraform created",
+	"single_ip_cidrs":       "2",
+	"ip_block_visibility":   "PRIVATE",
```
salv-orlando, line 19: Nit: quotas are mostly used for external blocks, might be worth focusing on those for testing

nsxt/resource_nsxt_policy_ip_block_quota_test.go
```diff
@@ -0,0 +1,254 @@
...
+	} else {
+		attrMap = accTestPolicyIpBlockQuotaUpdateAttributes
+		ipblock = "test-ipv6"
+	}
+	context := ""
+	if withContext {
+		context = testAccNsxtPolicyMultitenancyContext()
+	}
+	return fmt.Sprintf(`
+resource "nsxt_policy_ip_block" "test-ipv4" {
```
salv-orlando, line 203: As commented above, it might be worth creating blocks with external visibility for these tests

nsxt/resource_nsxt_policy_ip_block_quota_test.go
```diff
@@ -0,0 +1,254 @@
...
+	return fmt.Sprintf(`
+resource "nsxt_policy_ip_block_quota" "test" {
+  %s
+  display_name = "%s"
+  quota {
+    ip_block_visibility   = "%s"
+    ip_block_address_type = "%s"
+    single_ip_cidrs       = %s
+
+    other_cidrs {
```
salv-orlando, line 250: Is it mandatory to specify the empty struct, or are you verifying that it won't throw a nil pointer exception when used?

nsxt/resource_nsxt_vpc.go
```diff
@@ -115,6 +115,25 @@ var vpcSchema = map[string]*metadata.ExtendedSchema{
 			OmitIfEmpty:  true,
 		},
 	},
+	"quotas": {
```
salv-orlando, line 118: In this case could it be worth to deviate from API naming and call this attribute "ip_block_quotas"?

website/docs/r/policy_ip_block_quota.html.markdown
```diff
@@ -0,0 +1,92 @@
...
+    * `project_id` - (Required) The ID of the project which the object belongs to
+* `display_name` - (Required) The display name for the IP Block.
+* `description` - (Optional) Description of the resource.
+* `quota` - (Required) Quota specification
+  * `ip_block_paths` - (Required) List of IP blocks that this quota applies to.
+  * `ip_block_address_type` - (Optional) One of `IPV4`, `IPV6`. A quota will be applied on block of same address type. One v4 block and another v6 block cannot be speficied within the same quota.
+  * `ip_block_visibility` - (Optional) One of `EXTERNAL`, `PRIVATE`. A quota will be applied on blocks with same visibility. Private and External blocks cannot be specified within the same block.
+  * `single_ip_cidrs` - (Optional) Quota for single IP CIDRs allowed. Default is -1 (unlimited). 
+  * `other_cidrs` - (Required) Quota for other cidrs
+    * `mask` - (Optional) Largest mask size that is allowed
```
salv-orlando, line 70: I think we need to specify how this is specified (e.g.: /28).
User might think this is a subnet size specification (integer)

website/docs/r/policy_project.html.markdown
```diff
@@ -19,7 +19,6 @@ resource "nsxt_policy_project" "test" {
   description         = "Terraform provisioned Project"
   short_id            = "test"
   tier0_gateway_paths = ["/infra/tier-0s/test"]
-
```
salv-orlando, line 22: Is this accidental or did you mean doing something else?

nsxt/resource_nsxt_vpc.go
annakhm: There could be other types of quotas in the future

nsxt/resource_nsxt_policy_ip_block_quota_test.go
annakhm: I'm not sure what the behavior is if quota is already enforced - I think it would still allow to change the quota, even to smaller number

nsxt/resource_nsxt_policy_ip_block_quota_test.go
annakhm: I've made `other_cidrs` required since its auto-assigned, hence must be either `Required` or `Computed`. The latter option causes non-intuitive behavior when block is deleted (instead of reverting to default, the quota would stay as its last value, since terraform would consider it computed by NSX)

website/docs/r/policy_project.html.markdown
annakhm: just cosmetics - removing unneeded rows

api/infra/limit.go
```diff
@@ -0,0 +1,117 @@
+//nolint:revive
```
ksamoray, line 1: Where do we keep the `api_list.yaml` file?

annakhm: In generator repo

ksamoray (approved):

PR #1591, opened 2025-04-02
Fix race condition in vpc profile tests

annakhm:
Two parallel tests created profile with the same name, which resulted in failure in data source retrieval.
Avoid the issue by randomizing name in import test.

Signed-off-by: Anna Khmelnitsky <anna.khmelnitsky@broadcom.com>


salv-orlando (approved):
LGTM

ksamoray (approved):

PR #1592, opened 2025-04-03
Ease validation on project and vpc ids

annakhm:
The provider will allow special characters other than those reserved for URI

### DIFF
--- a/nsxt/policy_utils.go
+++ b/nsxt/policy_utils.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"log"
 	"net/url"
-	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


salv-orlando (approved):
LGTM!

ksamoray:
I think that `?`, `=` and maybe some others are reserved within a URL. IDK how NSX will handle these - maybe it could escape those while placed within the id?
Does NSX have some validation for ids?

annakhm:
> I think that `?`, `=` and maybe some others are reserved within a URL. IDK how NSX will handle these - maybe it could escape those while placed within the id? Does NSX have some validation for ids?

NSX does not seem to specify any restriction for ids. I tried a bunch on non-ASCII characters as well as reserved ones, all work.
For the provider to work, we need to avoid failing on policy path parsing, so I think we only care about `/`. In future we should consider supporting everything and escaping non-ascii characters.

ksamoray (approved):

PR #1595, opened 2025-04-07
chore(gh): update tests workflow

tenthirtyam:
### Description

Updates the tests workflow.

- Renames the workflow from `golangci-ut.yml` to `test.yml`.
- Pins the used actions to their release commit hash.
- Limits the permissions for the workflow.
- Limits runs to pull requests affecting Go files in the `nsxt/metadata` directory.


tenthirtyam:
> We also have tests in `validator_test.go` (and hopefully more to come), so we shouldn't be limited to metadata library.

@annakhm Want to change it to `go test -v ./...` and limit the path to `nsxt/**.go'?

The previous was `go test -v ./nsxt/metadata/...`

ksamoray (approved):
LGTM

PR #1600, opened 2025-04-09
Remove VPC support for the 3.8.1 bugfix release

annakhm:
VPC features will be released with 3.9.0


nsxt/provider.go
```diff
@@ -329,19 +329,6 @@ func Provider() *schema.Provider {
 			"nsxt_policy_distributed_flood_protection_profile":       dataSourceNsxtPolicyDistributedFloodProtectionProfile(),
 			"nsxt_policy_gateway_flood_protection_profile":           dataSourceNsxtPolicyGatewayFloodProtectionProfile(),
 			"nsxt_manager_info":                                      dataSourceNsxtManagerInfo(),
-			"nsxt_vpc":                                               dataSourceNsxtVPC(),
```
salv-orlando, line 332: This was already released in v3.8.0

nsxt/provider.go
```diff
@@ -518,31 +505,9 @@ func Provider() *schema.Provider {
 			"nsxt_policy_gateway_flood_protection_profile_binding":     resourceNsxtPolicyGatewayFloodProtectionProfileBinding(),
 			"nsxt_policy_compute_sub_cluster":                          resourceNsxtPolicyComputeSubCluster(),
 			"nsxt_policy_tier0_inter_vrf_routing":                      resourceNsxtPolicyTier0InterVRFRouting(),
-			"nsxt_vpc_security_policy":                                 resourceNsxtVPCSecurityPolicy(),
-			"nsxt_vpc_group":                                           resourceNsxtVPCGroup(),
-			"nsxt_vpc_gateway_policy":                                  resourceNsxtVPCGatewayPolicy(),
```
salv-orlando, line 523: These 3 were also released in v3.8.0.
If we made significant changes to them it might be the case of removing them.

salv-orlando (commented):
LGMT, but we need to decide what to do with the resources that we already released as experimental in v380

PR #1607, opened 2025-04-15
chore(gh): update release workflow

tenthirtyam:
- Updates `actions/checkout` to v4.2.2.
- Updates `actions/setup-go` to v5.4.0.
- Updates `golangci/golangci-lint-action` to v7.0.0.
- Updates `golangci/golangci-lint-action` to use `latest` of `golangci-lint` to match `golangci-lint.yml`.
- Removes `skip-pkg-cache` and `skip-build-cache` options that were removed April 2024 in `golangci/golangci-lint-action` v5.0.0 asinc ethe cache management is handled in `actions/setup-go`. 

Ref: https://github.com/golangci/golangci-lint-action/releases/tag/v5.0.0

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,3 +1,4 @@
+---
 name: Release
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -9,32 +10,29 @@ permissions:
...
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: 0
       - name: Setup Go
-        uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # v5.0.2
+        uses: actions/setup-go@0aaccfd150d50ccaeb58ebd88d36e91967a5f35b # v5.4.0
         with:
-          go-version-file: 'go.mod'
+          go-version-file: go.mod
       - name: golangci-lint
```
ksamoray, line 24: I liked your suggestion [here](https://github.com/vmware/terraform-provider-nsxt/pull/1587#issuecomment-2766575146) to run the linter from `./.github/workflows/golangci-lint.yml` instead of having this duplication

tenthirtyam: Yes, we can still do this - I was going to take a couple steps for this to be replicated across projects.

.github/workflows/release.yml
```diff
@@ -9,32 +10,29 @@ permissions:
...
         uses: golangci/golangci-lint-action@55c2c1448f86e01eaae002a5a3a9624417608d84 # v6.5.2
         with:
-          version: v1.61
-          args: --issues-exit-code=1 --timeout=30m
-          skip-pkg-cache: true
-          skip-build-cache: true
+          version: v1.64.8
```
ksamoray, line 27: We use latest in the per-PR lint action. Maybe we should be consistent here (or just use the same action).

ksamoray (approved):

PR #1608, opened 2025-04-15
Adding version check for tags data source acceptance tests

aruntony005:


nsxt/data_source_nsxt_policy_tags_test.go
```diff
@@ -21,6 +21,7 @@ func TestAccDataSourceNsxtPolicyTags_basic(t *testing.T) {
 		PreCheck: func() {
 			testAccPreCheck(t)
 			testAccOnlyLocalManager(t)
+			testAccNSXVersion(t, "4.2.0")
```
salv-orlando, line 24: Can you try and limit this to min version 4.0.0?

salv-orlando (approved):

nsxt/data_source_nsxt_policy_tags_test.go
aruntony005: Done

ksamoray:
lm-9.0.0 failure is due to a fix in #1613 this one is good to go.

PR #1614, opened 2025-04-21
Adding data source for T0 & T1 gateway interface

aruntony005:


nsxt/data_source_nsxt_policy_tier0_gateway_interface.go
```diff
@@ -0,0 +1,130 @@
...
+
+func dataSourceNsxtPolicyTier0GatewayInterface() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNsxtPolicyTier0GatewayInterfaceRead,
+
+		Schema: map[string]*schema.Schema{
+			"id":           getDataSourceIDSchema(),
+			"display_name": getDataSourceDisplayNameSchema(),
+			"description":  getDataSourceDescriptionSchema(),
+			"t0_gateway_name": {
```
annakhm, line 24: I this we should pull by `gateway_path` rather than `gateway_name` for consistency

nsxt/generic_policy_search.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-gm/model"
+
+	utl "github.com/vmware/terraform-provider-nsxt/api/utl"
+)
+
+type GenericResourceData struct {
```
annakhm, line 19: This seem like refactoring of the code in `policy_search.go`, could you please explain the reason for it?

annakhm:
Let's consider having a single data source `nsxt_policy_gateway_interface`, that would support both T0 and T1

go.mod
```diff
@@ -50,6 +50,7 @@ require (
 	github.com/hashicorp/terraform-registry-address v0.2.4 // indirect
 	github.com/hashicorp/terraform-svchost v0.1.1 // indirect
 	github.com/hashicorp/yamux v0.1.1 // indirect
+	github.com/katbyte/terrafmt v0.5.5 // indirect
```
salv-orlando, line 53: Do we know why this new dependency is being added?

nsxt/data_source_nsxt_policy_tier0_gateway_interface.go
```diff
@@ -0,0 +1,130 @@
...
+			"path": getPathSchema(),
+			"edge_cluster_path": {
+				Type:        schema.TypeString,
+				Description: "The path of the edge cluster connected to the Tier0 gateway linked to this interface",
+				Optional:    true,
+				Computed:    true,
+			},
+			"segment_path": {
+				Type:        schema.TypeString,
+				Description: "The path of the edge cluster connected to the Tier0 gateway linked to this interface",
```
salv-orlando, line 38: This description is likely a copy/paste from edge_cluster path, it might need to be amended

nsxt/data_source_nsxt_policy_tier0_gateway_interface.go
```diff
@@ -0,0 +1,130 @@
...
+		return err
+	}
+	t0GwList := []model.Tier0{}
+	for _, obj := range gwObjList {
+		dataValue, errors := converter.ConvertToGolang(obj, model.Tier0BindingType())
+		if len(errors) > 0 {
+			return errors[0]
+		}
+		curGw := dataValue.(model.Tier0)
+		if *curGw.DisplayName == t0Gw {
```
salv-orlando, line 69: @annakhm already asked to fetch the interface using the gateway path. This will also help here, because the display name is not unique.

Therefore you might end up having more than item in `t0GwList`, but if you use the path you will always have exactly one item and probably you can also directly fetch the desired T0 instead of fetching all of them

nsxt/data_source_nsxt_policy_tier0_gateway_interface.go
```diff
@@ -0,0 +1,130 @@
...
+
+		if isT0 && gwID == *t0GwList[0].Id {
+			isOp = true
+			if currInt.Path != nil {
+				err := d.Set("path", *currInt.Path)
+				if err != nil {
+					return fmt.Errorf("Error while setting interface path : %v", err)
+				}
+			}
+			if isPolicyGlobalManager(m) {
```
salv-orlando, line 101: Unlike T0, maybe for gateway interfaces even on GM we might be able to set the edge cluster path.
Can we verify this? I don't think a given interface will span multiple LM instances

nsxt/data_source_nsxt_policy_tier1_gateway_interface.go
```diff
@@ -0,0 +1,113 @@
...
+			"description":  getDataSourceDescriptionSchema(),
+			"t1_gateway_name": {
+				Type:        schema.TypeString,
+				Description: "The name of the Tier1 gateway where the interface is linked",
+				Required:    true,
+			},
+			"path": getPathSchema(),
+			"segment_path": {
+				Type:        schema.TypeString,
+				Description: "The path of the edge cluster connected to the Tier1 gateway linked to this interface",
```
salv-orlando, line 30: Description does not seem to be right here as well (same as the T0 interface data source)

nsxt/data_source_nsxt_policy_tier0_gateway_interface.go
salv-orlando: @annakhm can you please comment on this?

nsxt/data_source_nsxt_policy_tier0_gateway_interface.go
aruntony005: Have updated it to receive gateway_path

nsxt/generic_policy_search.go
aruntony005: I added the generic search because the existing one fails if are there with the same name pattern multiple resources (like same interface name in different gateways). But as I have changed the logic to get gateway path instead of the name this will not be needed any more. So I have removed it

docs/data-sources/policy_tier0_gateway_interface.md
```diff
@@ -0,0 +1,35 @@
...
+subcategory: "Gateways and Routing"
+page_title: "NSXT: policy_tier0_gateway_interface"
+description: A policy Tier-0 gateway interface data source.
+---
+
+# nsxt_policy_tier0_gateway_interface
+
+This data source provides information about policy Tier-0 gateway interface configured on NSX.
+
+This data source is applicable to NSX Policy Manager, NSX Global Manager and VMC.
```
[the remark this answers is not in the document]
aruntony005: Have removed it.

docs/data-sources/policy_tier0_gateway_interface.md
```diff
@@ -0,0 +1,35 @@
...
+* `id` - (Optional) The ID of Tier-0 gateway to retrieve.
+* `display_name` - (Optional) The Display Name prefix of the Tier-0 gateway interface to retrieve.
+* `t0_gateway_name` - (Optional) The Tier-0 gateway to retrieve to which the interface should be linked to.
+
+## Attributes Reference
+
+In addition to arguments listed above, the following attributes are exported:
+
+* `description` - The description of the resource.
+* `edge_cluster_path` - The path of the Edge cluster where this Tier-0 gateway is placed. This attribute is not set for NSX Global Manager, where gateway can spawn across multiple sites.
```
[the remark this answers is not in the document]
aruntony005: Have updated it.

go.mod
aruntony005: Mistakenly added while running website-lint. Have removed it and updated the go.sum as well.

nsxt/data_source_nsxt_policy_tier0_gateway_interface.go
aruntony005: Have added the gateway_path logic.

nsxt/data_source_nsxt_policy_tier1_gateway_interface.go
aruntony005: Updated it.

nsxt/data_source_nsxt_policy_tier0_gateway_interface.go
aruntony005: Updated the description.

nsxt/data_source_nsxt_policy_gateway_interface.go
```diff
@@ -0,0 +1,130 @@
...
+				Required:    true,
+			},
+			"path": getPathSchema(),
+			"edge_cluster_path": {
+				Type:        schema.TypeString,
+				Description: "The path of the edge cluster connected to the gateway linked to this interface. This is exported only for Tier0 gateways",
+				Optional:    true,
+				Computed:    false,
+			},
+			"segment_path": {
```
annakhm, line 36: I think we should allow this to be an input - namely, pull the interface by gateway path and segment path.
The original PR also suggested to allow locale service as an input

docs/data-sources/policy_gateway_interface.md
```diff
@@ -0,0 +1,52 @@
...
+* `display_name` - (Required) The Display Name prefix of the gateway interface to retrieve.
+* `gateway_path` - (Required) The path of the gateway to retrieve, where the interface should be linked to.
+
+## Attributes Reference
+
+In addition to arguments listed above, the following attributes are exported:
+
+* `id` - ID of the interface.
+* `description` - The description of the resource.
+* `edge_cluster_path` - The path of the Edge cluster where this gateway is placed. This attribute is not set for NSX Global Manager, where gateway can span across multiple sites. This attribute is set only for Tier0 gateways.
```
annakhm, line 50: Why are we only assigning this for T0 gateways?

nsxt/data_source_nsxt_policy_gateway_interface.go
```diff
@@ -0,0 +1,130 @@
...
+		description = currGwInterface.Description
+		segmentPath = currGwInterface.SegmentPath
+		id = currGwInterface.Id
+	}
+	if len(errors) > 0 {
+		return errors[0]
+	}
+
+	if path != nil {
+		err := d.Set("path", *path)
```
annakhm, line 93: nil check is already included in d.Set, so its safe to go with `d.Set("path", path)`

nsxt/data_source_nsxt_policy_gateway_interface_test.go
```diff
@@ -0,0 +1,177 @@
...
+	interfaceDescription := "Acceptance Test"
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		// CheckDestroy: func(state *terraform.State) error {
+		// 	return testAccNsxtPolicyTier0InterfaceCheckDestroy(state, t0InterfaceName)
+		// },
+		Steps: []resource.TestStep{
+			{
+				Config: testAccNsxtPolicyTier0InterfaceDataSourceTemplate(t0InterfaceName, t1InterfaceName, t0GatewayName, t1GatewayName, transportZoneName, interfaceDescription),
```
annakhm, line 27: I would suggest to reuse the test for the resource interface that we already have. One option is to add the data source test to resource test, another option would be to reuse the terraform code that creates GW + segment + interface (`testAccNsxtPolicyTier0InterfaceServiceTemplate`), and create a separate data source on top of that.

nsxt/data_source_nsxt_policy_gateway_interface.go
salv-orlando: @annakhm I think you meant that in the issue report the user would like to be able to search the interface by locale service too.
1) I think we should use locale_service_path (ie: the parent_path) instead of local service name
2) What about gateway_path? I think users should specify either service_path or gateway_path, but specifying both would probably not make sense.

docs/data-sources/policy_gateway_interface.md
aruntony005: T1 spec doesn't contain edge_cluster_path.

nsxt/data_source_nsxt_policy_gateway_interface.go
aruntony005: Have updated the Set functions.

nsxt/data_source_nsxt_policy_gateway_interface_test.go
aruntony005: Have updated the acceptance tests to reuse the tests for resource interface.

nsxt/data_source_nsxt_policy_gateway_interface.go
aruntony005: Adding service_path argument to the interface data source. Added acceptance tests for the same and updated the documentation.
Added a validation to check if either service_path or gateway_path is provided.

nsxt/data_source_nsxt_policy_gateway_interface.go
```diff
@@ -0,0 +1,133 @@
...
+		return errors[0]
+	}
+	d.Set("path", *path)
+	if isT0 && edgePath != nil {
+		err = d.Set("edge_cluster_path", *edgePath)
+		if err != nil {
+			return fmt.Errorf("Error while setting the interface edge cluster path : %v", err)
+		}
+	}
+	d.Set("description", *description)
```
annakhm, line 117: I believe description can be null, so dereferencing it here may crash.
`d.Set` can accept pointer and it performs a null check, so the following will work.
`d.Set("description", description)`

f.e. see here https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/data_source_nsxt_policy_ip_block.go#L84

same goes for other attributes above and below

salv-orlando (approved):
LGTM

nsxt/data_source_nsxt_policy_gateway_interface.go
aruntony005: Have updated the attributes.

ksamoray (approved):

PR #1617, opened 2025-04-22
Handle allocation_rules = nil case in policy_edge_cluster

ksamoray:
This case causes a provider crash, and should be checked

### DIFF
--- a/nsxt/resource_nsxt_policy_edge_cluster.go
+++ b/nsxt/resource_nsxt_policy_edge_cluster.go
@@ -324,25 +324,29 @@ func resourceNsxtPolicyEdgeClusterUpdate(d *schema.ResourceData, m interface{}) 
 
 	converter := bindings.NewTypeConverter()
 	var allocationRules []model.AllocationRule
-	allocRules := d.Get("allocation_rules").([]interface{})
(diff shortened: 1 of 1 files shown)


salv-orlando (approved):
LGTM

nsxt/resource_nsxt_policy_edge_cluster.go
```diff
@@ -324,25 +324,27 @@ func resourceNsxtPolicyEdgeClusterUpdate(d *schema.ResourceData, m interface{})
...
-			dataValue, errs := converter.ConvertToVapi(fdRule, model.AllocationBasedOnFailureDomainBindingType())
-			if errs != nil {
-				return handleCreateError("PolicyEdgeCluster", id, errs[0])
-			}
-			allocRule = model.AllocationRule{
-				Action: dataValue.(*data.StructValue),
+	allocRules := d.Get("allocation_rules")
+	if allocRules != nil {
+		for _, r := range allocRules.([]interface{}) {
+			rule := r.(map[string]interface{})
```
annakhm, line 330: This can be nil too in case of empty block, in which case casting will crash

PR #1619, opened 2025-04-22
Wait for ip address allocation on create

annakhm:
When ip address allocation resource is used without specific IP, platform allocates the address from the pool asyncronously. Previously, wait for the allocation happened in Read flow. This is problematic, since if allocation fails, resource cannot be deleted unless -refresh=false option is used.
This change fixes the code to wait for allocation in create flow. If allocation fails, create fails as well.

### DIFF
--- a/nsxt/resource_nsxt_policy_ip_address_allocation.go
+++ b/nsxt/resource_nsxt_policy_ip_address_allocation.go
@@ -125,6 +125,37 @@ func resourceNsxtPolicyIPAddressAllocationCreate(d *schema.ResourceData, m inter
 		return handleCreateError("IPAddressAllocation", id, err)
 	}
 
+	if allocationIP == "" {
+		// Wait for actual IP allocation
+		log.Printf("[DEBUG] Waiting for realization of IP Address for IP Allocation with ID %s", id)
+
+		obj, err := client.Get(poolID, id)
+		if err != nil {
+			return err
+		}
+
+		timeout := d.Get("timeout").(int)
+		stateConf := nsxtPolicyWaitForRealizationStateConf(connector, d, *obj.Path, timeout)
+		entity, err := stateConf.WaitForState()
+		if err != nil {
+			// Clean up NSX allocation
+			resourceNsxtPolicyIPAddressAllocationDelete(d, m)
+			return err
+		}
+		realizedResource := entity.(model.GenericPolicyRealizedResource)
+		for _, attr := range realizedResource.ExtendedAttributes {
+			if *attr.Key == "allocation_ip" {
+				d.SetId(id)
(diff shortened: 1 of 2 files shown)


nsxt/resource_nsxt_policy_ip_address_allocation.go
```diff
@@ -125,6 +125,37 @@ func resourceNsxtPolicyIPAddressAllocationCreate(d *schema.ResourceData, m inter
 		return handleCreateError("IPAddressAllocation", id, err)
 	}
 
+	if allocationIP == "" {
+		// Wait for actual IP allocation
```
ksamoray, line 129: Would be nicer to arrange this in a func. Otherwise LGTM.

annakhm: I could either create it as nested function (which is less readable in my opinion) or pass a lot of parameters (d, m, client, connector, etc..)

ksamoray: I agree about nested func - it has no advantage here. I thought that putting this long `if` in a func is more readable, but as you prefer, it's not really important.

ksamoray (approved):

nsxt/resource_nsxt_policy_ip_address_allocation.go
salv-orlando: I think it's ok to not use a function for this logic, I don't see a potential use case for code re-use, the reason would be readability only.

And this code change, after all, is pretty much moving the realization check from read to create.

nsxt/resource_nsxt_policy_ip_address_allocation.go
```diff
@@ -125,6 +125,37 @@ func resourceNsxtPolicyIPAddressAllocationCreate(d *schema.ResourceData, m inter
...
+		for _, attr := range realizedResource.ExtendedAttributes {
+			if *attr.Key == "allocation_ip" {
+				d.SetId(id)
+				d.Set("nsx_id", id)
+				d.Set("allocation_ip", attr.Values[0])
+				return resourceNsxtPolicyIPAddressAllocationRead(d, m)
+			}
+		}
+		// Clean up NSX allocation
+		resourceNsxtPolicyIPAddressAllocationDelete(d, m)
```
salv-orlando, line 155: Is this "just for safety"? I don't think there is a case where realisation would succeed but no IP was allocated.

salv-orlando (approved):
LGTM

nsxt/resource_nsxt_policy_ip_address_allocation.go
annakhm: Correct, this code should not be reachable

PR #1627, opened 2025-04-28
Update date of release in CHANGELOG

annakhm:


ksamoray:
LGTM, let's ignore the golangci-lint failure (or cherry-pick #1626 to fix it?).

ksamoray (approved):

ksamoray:
Maybe it's better if we cherry-pick #1626 as the linter runs also in the release process.

ksamoray:
Actually #1626 will address nothing of these... :/

PR #1633, opened 2025-04-30
Remove code generation scripts

annakhm:
Those scripts have moved to internal repository


ksamoray:
Upstream api_list.yaml contains entries for `Share`, `SharedResource` and `L7AccessProfile` which aren't in the downstream repo.
Let's keep these before we delete?

annakhm:
I've added the missing `SharedResource` to downstream

PR #1650, opened 2025-05-07
Cherry-pick v3.8.2 final changes into branch_390

ksamoray:


annakhm:
We don't need to remove the doc - this resource will be released with 3.9.0. I think only changelog needs cherry-picking

ksamoray:
> We don't need to remove the doc - this resource will be released with 3.9.0. I think only changelog needs cherry-picking

Fixed. Thanks!

salv-orlando (approved):
LGTM

PR #1653, opened 2025-05-07
Edge transport node docfix

ksamoray:
Data source doc filename is incorrectly suffixed.
Also add a comment for non-updatable attributes.


salv-orlando (approved):
lgtm

PR #1658, opened 2025-05-13
Fix doc for IP block quota resource

annakhm:
Visibility and IP address type are required attributes, and should appear in the example.


tenthirtyam (approved):

ksamoray (approved):

salv-orlando (approved):
lgtm

PR #1663, opened 2025-05-15
fix: staticcheck QF1003

tenthirtyam:
Refactors to use a tagged switch instead of an `if/else-if` chain.

Ref: `staticcheck` [QF1003](https://staticcheck.dev/docs/checks#QF1003)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -52,7 +52,7 @@ linters:
       # TODO: Setting temporary exclusions.
       - linters:
           - staticcheck
-        text: QF1001|QF1003
+        text: QF1001
       - linters:
           - staticcheck
         text: ST1005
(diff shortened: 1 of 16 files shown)


.golangci.yml
```diff
@@ -49,9 +49,9 @@ linters:
       - linters:
           - staticcheck
         text: QF1001
-      - linters:
-          - staticcheck
-        text: QF1003
+      # - linters:
```
ksamoray, line 52: Any reason to comment this out instead of removing?

ksamoray (commented):
/test-all

.golangci.yml
tenthirtyam: Should have been removed. Was commented during testing.

tenthirtyam: Will update this today whilst resolving the conflict post-rebase.

tenthirtyam: Resolved, but I did need to drop QF1001 back in since two additional items for that have popped up.

ksamoray (approved):

PR #1664, opened 2025-05-15
fix: staticcheck QF1004

tenthirtyam:
Refactors to use `strings.ReplaceAll` instead of `strings.Replace` with `n == -1`.

Ref: `staticcheck` [QF1004](https://staticcheck.dev/docs/checks#QF1004)


ksamoray (approved):

salv-orlando (approved):
LGTM

replace with n<0 is the same as replaceall

PR #1665, opened 2025-05-15
fix: staticcheck QF1007

tenthirtyam:
Merges conditional assignment into variable declaration.

Ref: `staticcheck` [QF1007](https://staticcheck.dev/docs/checks#QF1007)

### DIFF
--- a/nsxt/gateway_common.go
+++ b/nsxt/gateway_common.go
@@ -656,10 +656,7 @@ func parseGatewayInterfacePolicyPath(path string) (bool, string, string, string)
 		return false, "", "", ""
 	}
 
-	isT0 := true
-	if segs[2] != "tier-0s" {
-		isT0 = false
-	}
+	isT0 := (segs[2] == "tier-0s")
 
 	gwID := segs[3]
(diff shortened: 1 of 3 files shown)


nsxt/gateway_common.go
```diff
@@ -656,10 +656,7 @@ func parseGatewayInterfacePolicyPath(path string) (bool, string, string, string)
 		return false, "", "", ""
 	}
 
-	isT0 := true
-	if segs[2] != "tier-0s" {
-		isT0 = false
-	}
+	isT0 := !(segs[2] != "tier-0s")
```
salv-orlando, line 659: possibly dumb question: would it be ok to rewrite this to:

`isT0 :=  (segs[2] == "tier-0s")`

This would be just for readability

tenthirtyam: Yes, it should be! The above was auto-generated with `golangci-lint run --fix` based on the definitions to correct QF1007.

salv-orlando (approved):
LGTM

PR #1668, opened 2025-05-16
chore: apply hcl formatting to tests

tenthirtyam:
Applies `terraform fmt` formatting to `_test.go` files as needed.

```shell
terrafmt fmt ./nsxt
```

### DIFF
--- a/nsxt/resource_nsxt_policy_gateway_dns_forwarder_test.go
+++ b/nsxt/resource_nsxt_policy_gateway_dns_forwarder_test.go
@@ -268,7 +268,7 @@ data "nsxt_policy_gateway_dns_forwarder" "test" {
   display_name = "%s"
 
   gateway_path = nsxt_policy_tier%d_gateway.test.path
-  depends_on = [nsxt_policy_gateway_dns_forwarder.test]
+  depends_on   = [nsxt_policy_gateway_dns_forwarder.test]
 }
 `, context, attrMap["display_name"], attrMap["description"], whyDoesGoNeedToBeSoComplicated[isT0], attrMap["listener_ip"], attrMap["enabled"], attrMap["log_level"], attrMap["cache_size"], context, attrMap["display_name"], whyDoesGoNeedToBeSoComplicated[isT0])
 }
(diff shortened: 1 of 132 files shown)


ksamoray:
Can we implement some action to validate that this effort won't be lost with future PRs?

tenthirtyam:
Absolutely! I'll add that to my list, Kobi.

tenthirtyam:
@ksamoray - I've marked this one as draft whilst I add the check into an action for the team. In the meantime, I've updated the `GNUmakefile` with a `test-hcl` and `test-hcl-fix` similar to what is used for the documenation; however, there are a couple tests that need a review before we enable this in an action.

tenthirtyam:
> @ksamoray - I've marked this one as draft whilst I add the check into an action for the team. In the meantime, I've updated the `GNUmakefile` with a `test-hcl` and `test-hcl-fix` similar to what is used for the documenation; however, there are a couple tests that need a review before we enable this in an action.

Actually, I'll need to revist the exlusion and action in a seperate PR. In the meantime, the update to the makefile will fix any existing issues other than some of the specific tests that return issues when terrafmt is run against them.

ksamoray:
> > @ksamoray - I've marked this one as draft whilst I add the check into an action for the team. In the meantime, I've updated the `GNUmakefile` with a `test-hcl` and `test-hcl-fix` similar to what is used for the documenation; however, there are a couple tests that need a review before we enable this in an action.
> 
> Actually, I'll need to revist the exlusion and action in a seperate PR. In the meantime, the update to the makefile will fix any existing issues other than some of the specific tests that return issues when terrafmt is run against them.

So for now we'll run the make manually to make sure that we don't break too much of this formatting effort? Yeah that's doable.

nsxt/resource_nsxt_policy_gateway_dns_forwarder_test.go
```diff
@@ -248,7 +248,7 @@ resource "nsxt_policy_gateway_dns_forwarder" "test" {
 %s
   display_name = "%s"
   description  = "%s"
-  gateway_path = nsxt_policy_tier%d_gateway.test.path
+  gateway_path = nsxt_policy_tier % d_gateway.test.path
```
ksamoray, line 251: Spaces here break this test - this is not an HCL % operator but a sprintf %d.
There are a few of these below.

tenthirtyam: Resolved in https://github.com/vmware/terraform-provider-nsxt/commit/c4d727ee3f2865e81de7c333d72ac570948b9abc.

tenthirtyam:
@ksamoray - rebased in https://github.com/vmware/terraform-provider-nsxt/commit/107dc86131ed24b4241de6c1944afa9ba4ac265e. I removed `test-hcl-fix` for now because that would break some of the test formatting until we have a way to exclude those in a future update.

PR #1673, opened 2025-05-20
Fix LB UDP profile test failures with v9.1.0

ksamoray:
nsxt_policy_lb_udp_monitor_profile send and receive attributes became mandatory with v9.1.0.

### DIFF
--- a/nsxt/resource_nsxt_policy_lb_udp_monitor_profile_test.go
+++ b/nsxt/resource_nsxt_policy_lb_udp_monitor_profile_test.go
@@ -202,6 +202,8 @@ func testAccNsxtPolicyLBUdpMonitorProfileMinimalistic() string {
 	return fmt.Sprintf(`
 resource "nsxt_policy_lb_udp_monitor_profile" "test" {
   display_name = "%s"
+  receive = "%s"
+  send = "%s"
 
-}`, accTestPolicyLBUdpMonitorProfileUpdateAttributes["display_name"])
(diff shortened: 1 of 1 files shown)


ksamoray:
@annakhm @salv-orlando 
The v9.1.0 failure is due to some other test breakage, which should be addressed in NSX:
```
        Error:  Failed to create Project 67c69a93-03b7-4870-94be-e01c8bdcf364: Gateway connection path=[/infra/gateway-connections/57bc275c-aeca-4622-9da4-d23242e7d3d5] edge cluster does not match with Project path=[/infra/gateway-connections/57bc275c-aeca-4622-9da4-d23242e7d3d5] edge cluster. (code 612821)
        
          with nsxt_policy_project.test,
          on terraform_plugin_test.tf line 19, in resource "nsxt_policy_project" "test":
          19: resource "nsxt_policy_project" "test" {
```

It's safe to merge this in the meantime.

annakhm:
As changing `Optional` to` Required` is not backward compatible change - is NSX 9.1 mandating both those values? What about previous NSX versions?

salv-orlando (approved):
This looks good to me, but technically this is a backward-incompatible change.
However, the thing is that we have realized that these monitors could not possibly be working if these two attributes were not specified. Therefore, it is reasonable to expect that all users that have implemented them specified those parameters.

ksamoray:
As a user could have a working plan which would break with setting these to mandatory (OK the LB profile wouldn't be working but the plan would apply), it's better to keep these as optional and fix only the tests.

PR #1674, opened 2025-05-21
Ignore edge transport node order in cluster

ksamoray:
NSX may not retain the order of the nodes within the cluster when index is not specified by the user.
So using order-less set is better in this case.


salv-orlando (approved):
LGTM

PR #1675, opened 2025-05-21
Implement policy edge transport node realization datasource

ksamoray:
Also update nsxt_policy_edge_transport_node documentatation to use new DS.

### DIFF
--- a/docs/data-sources/policy_edge_transport_node_realization.md
+++ b/docs/data-sources/policy_edge_transport_node_realization.md
@@ -1,0 +1,77 @@
+---
+subcategory: "Realization"
+page_title: "NSXT: policy_edge_transport_node_realization"
+description: Policy Edge transport node resource realization information.
+---
+
+# nsxt_policy_edge_transport_node_realization
+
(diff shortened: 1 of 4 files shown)


docs/data-sources/policy_edge_transport_node_realization.md
```diff
@@ -0,0 +1,82 @@
...
+
+* `path` - (Required) The path for the Policy Edge transport node.
+* `delay` - (Optional) Delay (in seconds) before realization polling is started. Default is set to 1.
+* `timeout` - (Optional) Timeout (in seconds) for realization polling. Default is set to 1200.
+
+## Attributes Reference
+
+In addition to the arguments listed above, the following attributes are exported:
+
+* `state` - The realization state of the resource. Transitional states are: "pending", "in_progress", "in_sync", "unknown". Target states are: "success", "failed", "partial_success", "orphaned", "error".
```
annakhm, line 82: I understand that this data source fails in case of error state? If so, do we need `state` attribute? It can only be success for resolved data source

salv-orlando (approved):
LGTM

PR #1680, opened 2025-05-25
Add network_span_paths attr to HostTransportNodeCollection

ksamoray:
This is a new addition for NSX v9.1.0

### DIFF
--- a/nsxt/resource_nsxt_policy_host_transport_node_collection.go
+++ b/nsxt/resource_nsxt_policy_host_transport_node_collection.go
@@ -11,12 +11,13 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"github.com/vmware/terraform-provider-nsxt/nsxt/util"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/fabric/compute_collections"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/sites/enforcement_points"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+
+	"github.com/vmware/terraform-provider-nsxt/nsxt/util"
 )
 
 const removeOnDestroyDefault = true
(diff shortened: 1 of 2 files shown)


nsxt/resource_nsxt_policy_host_transport_node_collection.go
```diff
@@ -256,6 +270,11 @@ func resourceNsxtPolicyHostTransportNodeCollectionRead(d *schema.ResourceData, m
 		d.Set("sub_cluster_config", sccList)
 	}
 	d.Set("transport_node_profile_path", obj.TransportNodeProfileId)
+	if util.NsxVersionHigherOrEqual("9.1.0") {
+		if len(d.Get("network_span_paths").([]interface{})) > 0 {
```
annakhm, line 274: I don't think we need the null check here, since it will prevent nullifying paths if those were deleted on backend

nsxt/resource_nsxt_policy_host_transport_node_collection.go
```diff
@@ -165,6 +176,10 @@ func policyHostTransportNodeCollectionUpdate(siteID, epID, id string, isCreate b
 		TransportNodeProfileId: &transportNodeProfileID,
 		SubClusterConfig:       subClusterConfigs,
 	}
+	networkSitePaths := getStringListFromSchemaList(d, "network_span_paths")
+	if util.NsxVersionHigherOrEqual("9.1.0") && len(networkSitePaths) > 0 {
```
salv-orlando, line 180: the usual question... will this work of if user provides an empty list, e.g:

```
network_span_paths {
}
```

nsxt/resource_nsxt_policy_host_transport_node_collection.go
salv-orlando: I actually do not understand what @ksamoray meant do here... it seems we are setting network_span_paths only if we already have some paths in the schema?

ksamoray: I was trying to avoid setting an empty list, when the value returned from NSX is nil. IDK if this is required or not.

nsxt/resource_nsxt_policy_host_transport_node_collection.go
ksamoray: We have two different types of lists:
Those which have
```
    Elem: &schema.Resource{
```

Which are complex objects, and appear in the TF plan like:
```
   network_span_path {
   }
```

And for those we use singular naming, e.g `network_span_path`.

There's also these when have
```
    Elem: &schema.Schema{
        Type:         schema.TypeString,
```

Which are simple objects, and appear in the TF plan as
```
  network_span_paths = [...]
```

Those have plural names, e.g `network_span_paths`.

We should be able to handle empty list in this case.

PR #1682, opened 2025-05-26
Address test failures with v9.1.0

ksamoray:
Tests in v9.1.0 fail due to platform bug. To allow test execution against v9.1.0, fix this until the bug is fixed in NSX. Later we can revert this.


salv-orlando (approved):
LGTM

PR #1695, opened 2025-05-28
refactor: use nil slice declaration

tenthirtyam:
Replaces empty slice declaration with nil slice declaration.

An empty slice can be represented by nil or an empty slice literal. They are functionally equivalent — their len and cap are both zero — but the nil slice is the preferred style.


ksamoray (approved):

tenthirtyam:
JetBrains IDEs can detect this with code inspection.

PR #1697, opened 2025-05-28
In resource test, check data source only on create

annakhm:
When resource is updated, data source that is dependant on that resource would not wait for resource update to complete unless its waiting for specific computed attribute.

In case of the VtepHAHostSwitchProfile test, dependency is enforced with depends_on directive. This would work for create, but during update the data source would be resolved in parallel to resource update. To avoid the race condition, we perform data source validation only on create.


annakhm:
More test might suffer from this issue

salv-orlando (approved):
LGTM!

PR #1708, opened 2025-05-30
Support import for bgp config resource

annakhm:


nsxt/resource_nsxt_policy_bgp_config.go
```diff
@@ -28,11 +31,34 @@ func resourceNsxtPolicyBgpConfig() *schema.Resource {
...
 }
 
+func resourceNsxtPolicyBgpConfigImporter(d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
+	importID := d.Id()
+	parents, pathErr := parseStandardPolicyPathVerifySize(importID, 2, bgpConfigPathExample)
+	if pathErr != nil {
+		return nil, pathErr
+	}
+
+	tokens := strings.Split(importID, "/locale-services/")
```
salv-orlando, line 49: @salv-orlando address a nit here: if there is nothing specified after /locale-services/ it will still return 2 tokens.
You also need to make sure the 2nd token is non-empty.

Also verify if the 2nd token really needs to be in the form "local-service-id/bgp" or if we just need to amek sure the local service if specified.

ksamoray: @salv-orlando maybe we need some utility func that will do those validations - I tried `fmt.Sscanf()` but it's unusable here (it requires spaces). As this issue exists everywhere.

salv-orlando (approved):
LGTM

PR #1710, opened 2025-06-04
Transit gateway span attributes

ksamoray:
Add the span attributes to the transit gateway resource.

### DIFF
--- a/nsxt/resource_nsxt_policy_transit_gateway.go
+++ b/nsxt/resource_nsxt_policy_transit_gateway.go
@@ -6,6 +6,10 @@ package nsxt
 
 import (
 	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/vmware/terraform-provider-nsxt/nsxt/util"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
 	"log"
 	"reflect"
 
@@ -17,6 +21,11 @@ import (
 	utl "github.com/vmware/terraform-provider-nsxt/api/utl"
 	"github.com/vmware/terraform-provider-nsxt/nsxt/metadata"
 )
+
+var transitGatewayHaModeValues = []string{
+	model.TransitGatewayHighAvailabilityConfig_HA_MODE_ACTIVE,
+	model.TransitGatewayHighAvailabilityConfig_HA_MODE_STANDBY,
+}
 
 var transitGatewaySchema = map[string]*metadata.ExtendedSchema{
 	"nsx_id":       metadata.GetExtendedSchema(getNsxIDSchema()),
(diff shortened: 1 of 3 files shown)


nsxt/resource_nsxt_policy_transit_gateway.go
```diff
@@ -46,6 +55,136 @@ var transitGatewaySchema = map[string]*metadata.ExtendedSchema{
...
+										},
+									},
+								},
+							},
+						},
+						Metadata: metadata.Metadata{
+							Skip: true,
+						},
+					},
+					"zone_based_span": {
```
salv-orlando, line 152: the schema is missing use_all_zones

nsxt/resource_nsxt_policy_transit_gateway.go
```diff
@@ -46,6 +55,136 @@ var transitGatewaySchema = map[string]*metadata.ExtendedSchema{
...
+						Schema: schema.Schema{
```
salv-orlando, line 153: does this also need the ExactlyOneOf specified as cluster_based_span?

ksamoray: We use this in the same way in compute_manager (which I've coded too...). Also examples which I found elsewhere have this set only for one of the attributes.

nsxt/resource_nsxt_policy_transit_gateway.go
ksamoray: `use_all_zones` is marked as readonly, we do not include those normally.

nsxt/resource_nsxt_policy_transit_gateway.go
```diff
@@ -46,6 +55,141 @@ var transitGatewaySchema = map[string]*metadata.ExtendedSchema{
...
+			MaxItems: 1,
+			Computed: true,
+			Elem: &metadata.ExtendedResource{
+				Schema: map[string]*metadata.ExtendedSchema{
+					"cluster_based_span": {
+						Schema: schema.Schema{
+							Type:     schema.TypeList,
+							MaxItems: 1,
+							Computed: true,
+							ExactlyOneOf: []string{
```
salv-orlando, line 127: I never remember this: does exactlyOneOf require users to specify an element so that they will not be able to provide in input an empty lvalue for span?

nsxt/resource_nsxt_policy_transit_gateway.go
```diff
@@ -77,6 +221,44 @@ func resourceNsxtPolicyTransitGatewayExists(sessionContext utl.SessionContext, i
...
 
+func getSpanFromSchema(iSpan interface{}) (*data.StructValue, error) {
+	if len(iSpan.([]interface{})) == 0 {
+		return nil, nil
+	}
+	converter := bindings.NewTypeConverter()
+
+	// We're limiting to one span of any kind in the schema
+	span := iSpan.([]interface{})[0].(map[string]interface{})
+	if span["cluster_based_span"] != nil {
```
salv-orlando, line 232: Regarding switch vs if, does the linter check we recently added only catch cases where we use nested if then else blocks?

nsxt/resource_nsxt_policy_transit_gateway_test.go
```diff
@@ -28,11 +28,11 @@ func TestAccResourceNsxtPolicyTransitGateway_basic(t *testing.T) {
 	testResourceName := "nsxt_policy_transit_gateway.test"
 	testDataSourceName := "nsxt_policy_transit_gateway.test"
 
-	resource.ParallelTest(t, resource.TestCase{
+	resource.Test(t, resource.TestCase{
 		PreCheck: func() {
 			testAccPreCheck(t)
 			testAccOnlyVPC(t)
-			testAccNSXVersion(t, "9.1.0")
+			testAccNSXVersion(t, "9.0.0")
```
salv-orlando: ops, were we not running this test so far?

salv-orlando (commented):
Some minor comments

nsxt/resource_nsxt_policy_transit_gateway.go
ksamoray: Yep - if you add a 
```
  span{}
 ```
 
 to your TGW you'll get
 ```
 │ "span.0.cluster_based_span": one of `span.0.cluster_based_span,span.0.zone_based_span` must be specified
```

nsxt/resource_nsxt_policy_transit_gateway.go
ksamoray: I thought that Ryan added some check for this [here](https://github.com/vmware/terraform-provider-nsxt/pull/1663).
Maybe it only catches `if-else` which test for the same var (in this case it's checking a different map value).

nsxt/resource_nsxt_policy_transit_gateway_test.go
ksamoray: Yeah it's one of those which we've discussed - that are skipped by the LM job as they're `testAccOnlyVPC(t)` and then skipped by the VPC job which runs only those which match with `^TestAcc.*V[Pp][Cc].*$`.
Hopefully we'll address this soon and run VPC stuff in the LM job.

In the meantime I ran these TGW tests manually, e.g 
```
=== RUN   TestAccResourceNsxtPolicyTransitGateway_basic
--- PASS: TestAccResourceNsxtPolicyTransitGateway_basic (52.92s)
=== RUN   TestAccResourceNsxtPolicyTransitGateway_withSpan
--- PASS: TestAccResourceNsxtPolicyTransitGateway_withSpan (38.92s)
=== RUN   TestAccResourceNsxtPolicyTransitGateway_importBasic
=== PAUSE TestAccResourceNsxtPolicyTransitGateway_importBasic
=== CONT  TestAccResourceNsxtPolicyTransitGateway_importBasic
--- PASS: TestAccResourceNsxtPolicyTransitGateway_importBasic (23.73s)
PASS
```

ksamoray: And anyway this test should run only on 9.1.0 as 9.0 has no creation for TGW (default only)

nsxt/resource_nsxt_policy_transit_gateway.go
salv-orlando: Thanks

nsxt/resource_nsxt_policy_transit_gateway.go
salv-orlando: Sounds good, I have no strong opinion, we can leave the code as it is.

salv-orlando (approved):
LGTM

PR #1724, opened 2025-06-20
Tier1/Segment DS: Allow fetching global resources, move is_global into context

salv-orlando:
This change moves the is_global attribute for the T0 GW data source in the resource context.

To this aim this change:
- adds the context schema to the T0 GW data source
- adds the abiltiy of specifying a from_global boolean setting on the resource context
- changes TF session context generation to set the from_global setting. When this setting is true, project_id and vpc_id will be ignored as resources coming from global manager cannot have a multi-tenancy scope

### DIFF
--- a/docs/data-sources/policy_tier0_gateway.md
+++ b/docs/data-sources/policy_tier0_gateway.md
@@ -18,10 +18,23 @@ data "nsxt_policy_tier0_gateway" "tier0_gw_gateway" {
 }
 ```
 
+## Example Usage - Global infra
+
+```hcl
+data "nsxt_policy_tier0_gateway" "tier0_gw_gateway_global" {
+  context {
+    from_global = true
+  }
+  display_name = "tier0-gw"
+}
+```
+
 ## Argument Reference
 
 * `id` - (Optional) The ID of Tier-0 gateway to retrieve.
 * `display_name` - (Optional) The Display Name prefix of the Tier-0 gateway to retrieve.
+* `context` - (Optional) The context which the object belongs to
+  * `from_global` - (Optional) Set to True if the data source will need to search Tier-0 gateway created in a global manager instance (/global-infra)
 
 ## Attributes Reference
 
--- a/docs/data-sources/policy_tier1_gateway.md
+++ b/docs/data-sources/policy_tier1_gateway.md
@@ -33,12 +33,24 @@ data "nsxt_policy_tier1_gateway" "demotier1" {
 }
 ```
 
+## Example Usage - Global infra
+
+```hcl
+data "nsxt_policy_tier1_gateway" "tier1_router_global" {
+  context {
+    from_global = true
+  }
+  display_name = "tier1_gw"
+}
+```
+
 ## Argument Reference
 
 * `id` - (Optional) The ID of Tier-1 gateway to retrieve.
 * `display_name` - (Optional) The Display Name prefix of the Tier-1 gateway to retrieve.
 * `context` - (Optional) The context which the object belongs to
-  * `project_id` - (Required) The ID of the project which the object belongs to
+  * `project_id` - (Optional) The ID of the project which the object belongs to
+  * `from_global` - (Optional) Set to True if the data source will need to search Tier-1 gateway created in a global manager instance (/global-infra)
 
 ## Attributes Reference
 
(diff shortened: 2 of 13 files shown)


salv-orlando:
Updated PR description (I added the support for fetching global T1s, as I noticed it was not done earlier)

This change adds the ability to fetch globally-created Tier-1 gateways using the nsxt_policy_tier1_gateway data source, similarly to what is already available for nsxt_policy_tier0_gateway data source.
    
This change also moves the is_global attribute into the context, renaming it to "from_global". To this aim this change:

- adds the context schema to the T0 GW data source
- expands the context schema of the T1 GW data source to include from_global
- adds a from_global boolean setting in the context fro both T0 and T1 data sources
- changes TF session context generation to set the from_global setting. When this setting is true, project_id and vpc_id will be ignored as resources coming from global manager cannot have a  multi-tenancy scope

docs/data-sources/policy_tier1_gateway.md
```diff
@@ -38,7 +38,8 @@ data "nsxt_policy_tier1_gateway" "demotier1" {
 * `id` - (Optional) The ID of Tier-1 gateway to retrieve.
 * `display_name` - (Optional) The Display Name prefix of the Tier-1 gateway to retrieve.
 * `context` - (Optional) The context which the object belongs to
-  * `project_id` - (Required) The ID of the project which the object belongs to
+  * `project_id` - (Optional) The ID of the project which the object belongs to
+  * `from_global` - (Optinal) Set to True if the data source will need to search Tier-1 gateway created in a global manager instance (/global-infra)
```
ksamoray: s/Optinal/Optional/

docs/data-sources/policy_tier0_gateway.md
```diff
@@ -22,6 +22,8 @@ data "nsxt_policy_tier0_gateway" "tier0_gw_gateway" {
 
 * `id` - (Optional) The ID of Tier-0 gateway to retrieve.
 * `display_name` - (Optional) The Display Name prefix of the Tier-0 gateway to retrieve.
+* `context` - (Optional) The context which the object belongs to
+  * `from_global` - (Optinal) Set to True if the data source will need to search Tier-0 gateway created in a global manager instance (/global-infra)
```
ksamoray: s/Optinal/Optional/

ksamoray (commented):
Some nits. Are there other DS which from_global should apply to?

docs/data-sources/policy_tier1_gateway.md
```diff
@@ -38,7 +38,8 @@ data "nsxt_policy_tier1_gateway" "demotier1" {
 * `id` - (Optional) The ID of Tier-1 gateway to retrieve.
 * `display_name` - (Optional) The Display Name prefix of the Tier-1 gateway to retrieve.
 * `context` - (Optional) The context which the object belongs to
-  * `project_id` - (Required) The ID of the project which the object belongs to
+  * `project_id` - (Optional) The ID of the project which the object belongs to
```
ksamoray, line 52: As both `project_id` and `from_global` are optional, we might have this issue with empty context block, e.g
```
   context {}
```
Do we handle this case?

nsxt/data_source_nsxt_policy_tier1_gateway.go
```diff
@@ -8,6 +8,7 @@ import (
 	"fmt"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	utl "github.com/vmware/terraform-provider-nsxt/api/utl"
```
ksamoray: Shouldn't the linter complain about this? In-project imports should be separate, right?

salv-orlando: It did not complain, but I can surely move it.

docs/data-sources/policy_tier1_gateway.md
salv-orlando: I checked that the provider does not crash, because from_global defaults to False:

For
```
data "nsxt_policy_tier1_gateway" "t1_local_3" {
  context {}
  display_name = "TestTier1"
}
```
We get the following:
```
2025-06-27T10:12:39.065+0200 [INFO]  provider.terraform-provider-nsxt: 2025/06/27 10:12:39 [WARNING]: THIS IS YOUR CONTEXT: [map[from_global:false project_id:]]: timestamp="2025-06-27T10:12:39.065+0200"
```

ksamoray (approved):

PR #1726, opened 2025-06-23
Fix policy_edge_transport_node doc

ksamoray:
Using nsxt_policy_edge_transport_node data source in this use case is better than using nsxt_policy_edge_node.

While nsxt_policy_edge_node is a legacy DS which retrieves edge position with the cluster, nsxt_policy_edge_transport_node is used to retrieve the edge itself.


salv-orlando (approved):
lgtm

Issue #1716, opened 2025-06-11
Corrections to nsxt_policy_host_transport_node_collection


smoonen:
### Provider Version

3.8.2

### Motivation

Errors in doc

### Suggestion

Two issues with [https://registry.terraform.io/providers/vmware/nsxt/latest/docs/resources/policy_host_transport_node_collection](this page) -

First, the `sub_cluster_id` is described as taking a sub-cluster ID value. In practice, I have found that this causes an error. My plan is only successful if I enter the sub-cluster's _path_ rather than its id.

Second, the organization of the `sub_cluster_config` is incorrect. The `host_switch_config_source` is described as a required list, but with no objects nested underneath it. In practice I have found that the `host_switch_id` and `transport_node_profile_sub_config_name` parameters must be included inside of it.

Thank you!

PR #1733, opened 2025-06-24
Deprecate sub_cluster_id in host_transport_node_collection

ksamoray:
Name is misleading, as this attribute is actually a policy path. Replaced this with an attribute with a correct name and marked the old one as deprecated.

Partially fixes: #1716


salv-orlando (approved):
LGTM!

Issue #1730, opened 2025-06-23
Error on update after MP2P migration and import of r/nsxt_policy_tier0_gateway_interface


martinrohrbach:
### Terraform

1.10.5

### Terraform Provider

3.8.0

### VMware NSX

4.2.2.0

### Description

We've finally started our MP2P migration (see also #854). We directly use the migration API and have extensively tested this, looks good so far.

After we've migrated an object (or a specific tree of objects really), we create the matching terraform ressources in policy API and reimport those objects so we can continue using them with IaC.

We've now found the first issue with this. After we've imported a nsxt_policy_tier0_gateway_interface that is bound to a particular edge, we can't modify the interface with terraform at all, instead we get the error:

`Edge cluster index cannot be modified after tier-0 interface path=[/infra/tier-0s/<t0-id>/locale-services/<locale-services-id>/interfaces/<interface-id>] deployment."`

We've tried updating the interface using the API and had no problems whatsoever so It's not an NSX issue per se. 

So we also had a look at what the provider is doing and it looks like it does not preserve the "edge_cluster_member_index" property which is MP2P migrated over from manager API and then triggers the error within NSX.

This is how the interface looks during the plan (ommitting everything that doesn't start with "edge"):
```
{
  "edge_cluster_member_index" : 2,
  "edge_path" : "/infra/sites/default/enforcement-points/default/edge-clusters/44c5ee35-3898-409e-a2c5-d921e5904d7b/edge-nodes/d22f8498-a253-11e9-a6c5-00505696031f",
...
  "_revision" : 31
}
```

And that's what is sent during apply:
```
{
  "edge_path": "/infra/sites/default/enforcement-points/default/edge-clusters/44c5ee35-3898-409e-a2c5-d921e5904d7b/edge-nodes/d22f8498-a253-11e9-a6c5-00505696031f",
...
  "_revision": 31,
}
```

If I use the API to PUT the exact same update from the apply and preserve the edge_cluster_member_index then everything behaves as expected. 

Could this be fixed by e.g. keeping the "edge_cluster_member_index" if it is initially set? 

### Affected Resources or Data Sources

nsxt_policy_tier0_gateway_interface

### Terraform Configuration

See above

### Debug Output

See above

### Expected Behavior

nsxt_policy_tier0_gateway_interface can be updated after import from MP2P migration.

### Actual Behavior

Update failes after import from MP2P migration. 

### Steps to Reproduce

- Create Tier0 in manager API
[...]

PR #1737, opened 2025-06-30
Add edge_cluster_member_index to T0 gw interface

ksamoray:
While importing this resource and executing an update, there are failures as this Attribute is missing.
We add it as deprecated as it's deprecated in the API spec.

Fixes: #1730


salv-orlando (approved):
LGTM

PR #1738, opened 2025-07-01
Updating the policy VM data sources to fix the usability issues raised

aruntony005:


ksamoray:
The fact that display_name can be a regular expression is not documented. It's worthwhile to add an example with a regexp as well.

nsxt/data_source_nsxt_policy_vms.go
```diff
@@ -32,7 +33,11 @@ func dataSourceNsxtPolicyVMs() *schema.Resource {
 		Read: dataSourceNsxtPolicyVMsRead,
 
 		Schema: map[string]*schema.Schema{
-			// TODO: add option to filter by display name regex
+			"display_name": {
+				Type:        schema.TypeString,
+				Description: "The display name regex of the VMs",
```
salv-orlando: nit: consider "display name for the VM. Supports regular expressions"

nsxt/data_source_nsxt_policy_vms.go
```diff
@@ -76,6 +82,13 @@ func dataSourceNsxtPolicyVMsRead(d *schema.ResourceData, m interface{}) error {
 	if err != nil {
 		return fmt.Errorf("Error reading Virtual Machines: %v", err)
 	}
+	var re *regexp.Regexp
+	if d.HasChange("display_name") {
```
salv-orlando: Why did you use HasChange rather than GetOK?

nsxt/data_source_nsxt_policy_vms.go
```diff
@@ -76,6 +82,13 @@ func dataSourceNsxtPolicyVMsRead(d *schema.ResourceData, m interface{}) error {
...
+		re, err = regexp.Compile(displayNameRegex)
```
salv-orlando: Did you verify this also works if you specify an exact display name rather than a regex?

nsxt/data_source_nsxt_policy_vms.go
```diff
@@ -92,7 +105,7 @@ func dataSourceNsxtPolicyVMsRead(d *schema.ResourceData, m interface{}) error {
 				}
 			}
 		}
-		if vm.DisplayName == nil {
+		if vm.DisplayName == nil || (d.HasChange("display_name") && !re.MatchString(*vm.DisplayName)) {
```
salv-orlando: you can consider adding an else statement for the if block at line 86 and compile the regex to `.*` if display_name is not provided, and simplify this line in the following way:

`if vm.DisplayName == nil || !!re.MatchString(*vm.DisplayName))`

nsxt/data_source_nsxt_policy_vms.go
aruntony005: Have updated the description

nsxt/data_source_nsxt_policy_vms.go
aruntony005: Yes. It works

nsxt/data_source_nsxt_policy_vms.go
aruntony005: Because I just needed if a value is set. 
I've updated it now, since display_name wasn't used anywhere else — just within the if block.

nsxt/data_source_nsxt_policy_vms.go
aruntony005: Have made the change

docs/data-sources/policy_vm.md
```diff
@@ -18,6 +18,14 @@ data "nsxt_policy_vm" "nsxt_vm1" {
 }
 ```
 
+## Example Usage - Filtering by UID (external_id, bios_id or instance_id)
```
ksamoray, line 21: I'm a bit confused: you update the `policy_vm` doc while changing the `policy_vms` data source. Is that on purpose?
Also, `policy_vms` doc doesn't say that `display_name` has regex capability. I'd add an example as well.

aruntony005: There was a request for the policy_vm. Hence the change in the policy_vm doc.
Also I have updated the policy_vms documentation with the new regex support and an example.

ksamoray (approved):

PR #1741, opened 2025-07-02
Use latest dev SDK and apply required changes

ksamoray:


nsxt/resource_nsxt_policy_project.go
```diff
@@ -112,23 +112,33 @@ func resourceNsxtPolicyProject() *schema.Resource {
...
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
-						"default_span": {
-							Type:         schema.TypeString,
-							Description:  "Policy path of the Cluster based Span object of type NetworkSpan",
-							ValidateFunc: validatePolicyPath(),
-							Optional:     true,
-						},
-						"non_default_spans": {
+						"span_reference": {
```
salv-orlando, line 120: Did you check if span_reference and zone_external_ids need to be mutually exclusive?

nsxt/resource_nsxt_policy_project.go
```diff
@@ -112,23 +112,33 @@ func resourceNsxtPolicyProject() *schema.Resource {
 			"vpc_deployment_scope": {
 				Type:        schema.TypeList,
 				Optional:    true,
+				Computed:    true,
```
salv-orlando, line 115: What do we get if we don't specify anything? Is it automatically populated with the default span?

ksamoray: It's populated with default span (at least for now, that's why all these attributes are computed actually).

nsxt/resource_nsxt_policy_project.go
ksamoray: IDK - where can we check that?

PR #1742, opened 2025-07-02
Handle crash in policy transit gateway

ksamoray:
nsxt_policy_transit_gateway resource crashed when called with empty span list.


salv-orlando (approved):
LGTM

PR #1751, opened 2025-07-11
Adding default span reference to handle empty span issue

aruntony005:


nsxt/resource_nsxt_policy_project.go
```diff
@@ -112,16 +112,24 @@ func resourceNsxtPolicyProject() *schema.Resource {
...
 				MaxItems:    1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						"default_span_path": {
+							Type:         schema.TypeString,
+							Description:  "Policy path of the Cluster based default Span object of type NetworkSpan",
+							ValidateFunc: validatePolicyPath(),
+							Default:      "/infra/network-spans/default",
+							Optional:     true,
+							Computed:     false,
```
salv-orlando: Since NSX API may populate the default span, should this be computed?

nsxt/resource_nsxt_policy_project.go
```diff
@@ -232,12 +234,18 @@ func resourceNsxtPolicyProjectPatch(connector client.Connector, d *schema.Resour
 		// There should be just one object here
 		vdScope := vdsIntface.([]interface{})[0].(map[string]interface{})
 		var spanReferences []model.SpanReference
+		defaultSpanPath := vdScope["default_span_path"].(string)
```
salv-orlando: This is optional, it might be missing in the resource spec.

nsxt/resource_nsxt_policy_project.go
ksamoray: Also, `Computed` is false by default, you could instead drop the `Computed` attribute from these schema attrs.

nsxt/resource_nsxt_policy_project.go
aruntony005: But there is a default value in the schema "/infra/network-spans/default"

nsxt/resource_nsxt_policy_project.go
salv-orlando: As we discussed we probably want to keep this computed: true, and remove the default value. What do you reckon?

aruntony005: Have updated it

nsxt/resource_nsxt_policy_project.go
```diff
@@ -112,16 +112,23 @@ func resourceNsxtPolicyProject() *schema.Resource {
...
+							ValidateFunc: validatePolicyPath(),
+							Optional:     true,
+							Computed:     true,
+						},
 						"span_reference": {
 							Type:        schema.TypeList,
 							Description: "List of Span object references available with the project for TGW consumption",
 							MaxItems:    10,
-							Computed:    true,
+							Computed:    false,
```
ksamoray: nit: `Computed: false` is redundant

nsxt/resource_nsxt_policy_project.go
```diff
@@ -130,13 +137,7 @@ func resourceNsxtPolicyProject() *schema.Resource {
...
 										ValidateFunc: validatePolicyPath(),
 										Optional:     true,
-										Computed:     true,
-									},
-									"is_default": {
-										Type:        schema.TypeBool,
-										Description: "Default span indicator",
-										Optional:    true,
-										Computed:    true,
+										Computed:     false,
```
ksamoray: nit: `Computed: false` is redundant

nsxt/resource_nsxt_policy_project.go
```diff
@@ -389,14 +398,20 @@ func resourceNsxtPolicyProjectRead(d *schema.ResourceData, m interface{}) error
 	if util.NsxVersionHigherOrEqual("9.1.0") && obj.VpcDeploymentScope != nil {
 		deploymentScope := make(map[string]interface{})
```
ksamoray, line 399:
```suggestion
		deploymentScope := make(map[string]interface{})
		var spanRefs []interface{}
		var defaultSpanRefs *string
		for _, spanRef := range obj.VpcDeploymentScope.SpanReferences {
			if *spanRef.IsDefault {
				defaultSpanRefs = spanRef.SpanPath
			} else {
				sr := make(map[string]interface{})
				sr["span_path"] = spanRef.SpanPath
				spanRefs = append(spanRefs, sr)
			}
		}
		deploymentScope["default_span_path"] = defaultSpanRefs
		if len(spanRefs) > 0 {
			deploymentScope["span_reference"] = spanRefs
		}
```

nsxt/resource_nsxt_policy_project.go
aruntony005: Removed the flag.

nsxt/resource_nsxt_policy_project.go
aruntony005: Removed the flag.

nsxt/resource_nsxt_policy_transit_gateway_static_route_test.go
```diff
@@ -239,6 +239,8 @@ resource "nsxt_policy_project" "test" {
   site_info {
     edge_cluster_paths = [data.nsxt_policy_edge_cluster.test.path]
   }
+  vpc_deployment_scope {
```
ksamoray: Why having
```
  vpc_deployment_scope {
  }
```
here and in the tests above? This isn't a mandatory block. Maybe keep it somewhere to test this case but remove from others?

aruntony005: Few acceptance tests were failing, because while patching a resource the default span added by NSX creates an unexpected diff. So added this part as a temporary fix. 
I have removed it now. Making vpc_deployment_scope computed fixed this issue.

docs/resources/policy_project.md
```diff
@@ -43,9 +43,8 @@ The following arguments are supported:
 * `vc_folder` - (Optional) Flag to specify whether the DVPGs created for project segments are grouped under a folder on the VC. `vc_folder` should be set to true when `vpc_deployment_scope` is set.
 * `quotas` - (Optional) List of policy paths for quota resources that are applicable to this project.
 * `vpc_deployment_scope` - (Optional) Project VPC network Deployment Scope.
-    * `span_reference` - (Optional) List of Span object references available with the project for TGW consumption.
-          * `span_path` - (Required) Policy path of the Cluster based Span object of type NetworkSpan.
-          * `is_default` - (Optional) Default span indicator.
+    * `default_span_path` - (Optional) Policy path of the Cluster based default Span object of type NetworkSpan.
+    * `non_default_span_paths` - (Optional) List of non default policy path of the Span objects of type NetworkSpan
```
ksamoray: nit: policy paths

nsxt/resource_nsxt_policy_project.go
```diff
@@ -117,28 +117,21 @@ func resourceNsxtPolicyProject() *schema.Resource {
...
+						"default_span_path": {
+							Type:         schema.TypeString,
+							Description:  "Policy path of the Cluster based default Span object of type NetworkSpan",
+							ValidateFunc: validatePolicyPath(),
+							Optional:     true,
+							Computed:     true,
+						},
+						"non_default_span_paths": {
 							Type:        schema.TypeList,
 							Description: "List of Span object references available with the project for TGW consumption",
```
ksamoray: This description is not correct anymore (you can just use the same description from the doc).

ksamoray (commented):
A couple of nits. Other than these, we're good!

docs/resources/policy_project.md
aruntony005: Updated

nsxt/resource_nsxt_policy_project.go
aruntony005: Updated

PR #1757, opened 2025-07-15
Updating the gateway connection documentation to provide ip block paths instead of CIDR

aruntony005:


salv-orlando (commented):
LGTM, but before submitting you need to resolve conflicts.

PR #1758, opened 2025-07-15
Handle upgrade run test failures

ksamoray:
When the user pauses after each upgrade unit, the upgrade_run resource might exit before the upgrade is reaching a permanent state. This fix should correct this.


salv-orlando (approved):
LGTM

PR #1759, opened 2025-07-17
Add is_default field to TGW data_source

poojav25:


ksamoray:
Let's validate this in a test as well. Also this might be handy as an argument for looking up the default TGW.

nsxt/data_source_nsxt_policy_transit_gateway.go
```diff
@@ -18,14 +20,37 @@ func dataSourceNsxtPolicyTransitGateway() *schema.Resource {
...
 func dataSourceNsxtPolicyTransitGatewayRead(d *schema.ResourceData, m interface{}) error {
-	_, err := policyDataSourceResourceRead(d, getPolicyConnector(m), getSessionContext(d, m), "TransitGateway", nil)
+	// Using deprecated API because GetOk is not behaving as expected when is_default = "false".
+	// It does not return true for a key that's explicitly set to false.
+	_, defaultOK := d.GetOkExists("is_default")
+	_, dpOk := d.GetOk("display_name")
+	_, idOk := d.GetOk("id")
+
+	if defaultOK && !dpOk && !idOk {
+		_, err := policyDataSourceReadWithFlag(d, getPolicyConnector(m), getSessionContext(d, m), "TransitGateway", "is_default", nil)
```
ksamoray: Instead of adding `policyDataSourceReadWithFlag` you could build a query map as done [here](https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/data_source_nsxt_policy_transport_zone.go#L85) and use `policyDataSourceResourceReadWithValidation`. That would simplify the change IMO.

poojav25: okay will do

poojav25: okay will do

poojav25:
> Seems like the context isn't processed or validated properly in this datasource, and the test which you've added just exposes this bug.

--> this is due to project env not being set , I have faced it similarly locally , once i explicitly set this "NSXT_PROJECT_ID="default" , it would resolve this crash

ksamoray:
> > Seems like the context isn't processed or validated properly in this datasource, and the test which you've added just exposes this bug.
> 
> --> This is due to project env not being set , I have faced it similarly locally , once i explicitly set this "NSXT_PROJECT_ID="default" , it would resolve this crash I will check how to make sure env is not null

Hi, no matter what input the provider is given, it shouldn't crash. Anyway this bug is not due to your change but should be addressed in any case.
Regarding NSXT_PROJECT_ID being set - in the resource tests it's being handled by creating a project within the test scope.

nsxt/data_source_nsxt_policy_transit_gateway.go
poojav25: I modified the code to use the additionalQueryMap to pass the new fields and construct the query with default strings (like 'resource type' and 'mark for delete=false'). I also validated the policyDataSourceResourceReadWithValidation function for reuse in this use case. However, the function is currently hardcoded to check for display name or ID, and the callback for the query build function also uses hardcoded display name and ID fields

poojav25: updated the ProjectID validator  func to check for the empty projectID string

nsxt/data_source_nsxt_policy_transit_gateway.go
```diff
@@ -18,14 +20,39 @@ func dataSourceNsxtPolicyTransitGateway() *schema.Resource {
 			"description":  getDataSourceDescriptionSchema(),
 			"path":         getPathSchema(),
 			"context":      getContextSchemaExtended(true, false, false, true),
+			"is_default": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: false,
```
ksamoray: nit: `Computed: false` is redundant as false is the default.

nsxt/data_source_nsxt_policy_transit_gateway.go
```diff
@@ -18,14 +20,39 @@ func dataSourceNsxtPolicyTransitGateway() *schema.Resource {
...
 
 func dataSourceNsxtPolicyTransitGatewayRead(d *schema.ResourceData, m interface{}) error {
-	_, err := policyDataSourceResourceRead(d, getPolicyConnector(m), getSessionContext(d, m), "TransitGateway", nil)
+	// Using deprecated API because GetOk is not behaving as expected when is_default = "false".
+	// It does not return true for a key that's explicitly set to false.
+	_, defaultOK := d.GetOkExists("is_default")
+	_, dpOk := d.GetOk("display_name")
+	_, idOk := d.GetOk("id")
+
+	if defaultOK && !dpOk && !idOk {
```
ksamoray, line 40: This behavior should be documented, that is_default and the other attributes are mutually exclusive.

nsxt/policy_search.go
```diff
@@ -115,9 +115,36 @@ func policyDataSourceResourceReadWithValidation(d *schema.ResourceData, connecto
 	return policyDataSourceResourceFilterAndSet(d, resultValues, resourceType)
 }
 
-// globalFlag argument is an optional
-func listPolicyResourcesByNameAndType(connector client.Connector, context utl.SessionContext, displayName string, resourceType string, additionalQuery *string, isGlobal bool) ([]*data.StructValue, error) {
-	query := fmt.Sprintf("resource_type:%s AND display_name:%s* AND marked_for_delete:false", resourceType, escapeSpecialCharacters(displayName))
+func policyDataSourceReadWithCustomField(d *schema.ResourceData, connector client.Connector, context utl.SessionContext, resourceType string, additionalQuery map[string]string) (*data.StructValue, error) {
```
ksamoray, line 118: So we need `policyDataSourceReadWithCustomField()` to avoid the regular processing of `display_name` and `id`?
It sounds reasonable but maybe we should document this - at some point someone might want to replicate this behavior and reuse this.

ksamoray (commented):
LGTM, some minor things

nsxt/data_source_nsxt_policy_transit_gateway_test.go
```diff
@@ -0,0 +1,51 @@
...
+	testAccDataSourceNsxtPolicyTransitGatewayBasic(t, true, func() {
+		testAccPreCheck(t)
+		testAccOnlyVPC(t)
+		testAccNSXVersion(t, "9.1.0")
+		testAccEnvDefined(t, "NSXT_PROJECT_ID")
+
+	})
+}
+
+func testAccDataSourceNsxtPolicyTransitGatewayBasic(t *testing.T, withContext bool, preCheck func()) {
```
ksamoray: No reason to split out `testAccDataSourceNsxtPolicyTransitGatewayBasic` from `TestAccDataSourceNsxtPolicyTransitGateway_basic`. As testAccDataSourceNsxtPolicyTransitGatewayBasic isn't reused anywhere (e.g in a multitenancy test), no need to have this function, the whole code should be in the `TestAccDataSourceNsxtPolicyTransitGateway_basic` func.

poojav25: got it

nsxt/policy_search.go
poojav25: okay!

nsxt/data_source_nsxt_policy_transit_gateway.go
poojav25: okay

ksamoray (approved):

PR #1769, opened 2025-07-28
Handle 30023 error correctly in upgrade-run

ksamoray:
Status check may result with BAD_REQUEST + 30023 error code. This case should be ignored.


salv-orlando (approved):
LGTM

PR #1775, opened 2025-08-04
Make sure that Edge is realized, before allowing creation of the RTEP

ksamoray:
RTEP creation fails when Policy Edge is not realized successfully.


salv-orlando (approved):
lgtm

PR #1780, opened 2025-08-10
Fix nsxt_policy_tier*_gateways failures on GM

ksamoray:
GM tests fails due to test issues with newly introduced datasources.


salv-orlando (approved):
LGTM

Issue #1534, opened 2025-02-10
TF Provider tries to change order auf VTI Default Rules in nsxt_policy_predefined_gateway_policy


benzander:
### Describe the bug

I use nsxt_policy_predefined_gateway_policy to change the default rule action from ALLOW to Drop in all gateway Firewalls.
For every Route Based VPN the NSX Manager creates a default rule. So I use the same ressource to set those rules to DROP as well.

Unfortunately I run into an issue in case there are multiple Route Based VPNs in the Same T1 or T0.
The TF Providers tries to change the order of those default rules. That is how the Change looks like with 4 Route Based VPNs:

```
Terraform will perform the following actions:

  # module.vrf_On_CustomerT0_t0V-001.nsxt_policy_predefined_gateway_policy.gwFW-predefinedGatewayPolicy will be updated in-place
  ~ resource "nsxt_policy_predefined_gateway_policy" "gwFW-predefinedGatewayPolicy" {
        id          = "Policy_Default_Infra-tier0-df1f3954-0683-43db-89a0-f5bfe52fe142"
        # (3 unchanged attributes hidden)

      ~ default_rule {
          ~ scope           = "/infra/tier-0s/df1f3954-0683-43db-89a0-f5bfe52fe142/ipsec-vpn-services/8241df91-878f-45f9-8902-8d16a39d511e/sessions/84f05f4a-e89c-4d1e-a1bc-106a08a8e724" -> "/infra/tier-0s/df1f3954-0683-43db-89a0-f5bfe52fe142/ipsec-vpn-services/8241df91-878f-45f9-8902-8d16a39d511e/sessions/50c5e8d2-42b9-48d4-b705-640ffbe224b7"
            # (8 unchanged attributes hidden)
        }
      ~ default_rule {
          ~ scope           = "/infra/tier-0s/df1f3954-0683-43db-89a0-f5bfe52fe142/ipsec-vpn-services/8241df91-878f-45f9-8902-8d16a39d511e/sessions/8ed8467a-deed-4d01-84eb-7623d6cd8814" -> "/infra/tier-0s/df1f3954-0683-43db-89a0-f5bfe52fe142/ipsec-vpn-services/8241df91-878f-45f9-8902-8d16a39d511e/sessions/84f05f4a-e89c-4d1e-a1bc-106a08a8e724"
            # (8 unchanged attributes hidden)
        }
      ~ default_rule {
[...]

PR #1781, opened 2025-08-11
Ignore gateway policy default rule order

ksamoray:
When there are multiple default rules, NSX may return the rules in an incorrect order and cause permadiff.

Fixes: #1534

### DIFF
--- a/nsxt/resource_nsxt_policy_predefined_gateway_policy.go
+++ b/nsxt/resource_nsxt_policy_predefined_gateway_policy.go
@@ -50,7 +50,7 @@ func getPolicyPredefinedGatewayPolicySchema() map[string]*schema.Schema {
 
 func getGatewayPolicyDefaultRulesSchema() *schema.Schema {
 	return &schema.Schema{
-		Type:          schema.TypeList,
+		Type:          schema.TypeSet,
 		Description:   "List of default rules",
 		Optional:      true,
 		Computed:      true,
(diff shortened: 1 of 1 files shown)


nsxt/resource_nsxt_policy_predefined_gateway_policy.go
```diff
@@ -50,7 +50,7 @@ func getPolicyPredefinedGatewayPolicySchema() map[string]*schema.Schema {
 
 func getGatewayPolicyDefaultRulesSchema() *schema.Schema {
 	return &schema.Schema{
-		Type:          schema.TypeList,
+		Type:          schema.TypeSet,
```
salv-orlando, line 53: When using Typeset, do we still have a guarantee that the order in which the rules are specified in Terraform is the same in which they are created in NSX? For firewall rules, we must respect the order.

ksamoray: @salv-orlando 
Rule order wouldn't be impacted here - just default_rule order. Does this order has any significance? If so - why would NSX change it?

poojav25: Since TypeSet does not maintain order and removes duplicates, if the firewall rules must be in a specific order, would using the lifecycle block to ignore changes to the rule order help in this case?

ksamoray: Hi @poojav25, as these aren't rule but a per-scope default rules, their order is not necessarily important as each of these is evaluated in different scope. I've asked PM to clarify this, as otherwise this has to be handled by NSX (they're changing the order, which is why we have this issue).

Issue #1782, opened 2025-08-12
Provider crashed and left some changes untracked in state file


tamere-allo-peter:
### Terraform

Tofu 1.10.0

### Terraform Provider

3.9.0

### VMware NSX

4.1.1.0.0.22224312

### Description

OpenTofu v1.10.0
on linux_amd64
+ provider registry.opentofu.org/hashicorp/local v2.5.3
+ provider registry.opentofu.org/vmware/nsxt v3.9.0

While trying to create a huge bunch of groups (>900) and a security policy with a set of distributed firewall rules the provider crashed. When I launched a destroy, only a subset of what was created was destroyed, but a few NSX groups that were created remain, so they are probably not tracked anymore in the state file.

When creating only the huge number of groups, all is fine. All is fine as well if I try to create a different security policy, and so I believe I've traced the problem to be related to an empty service_entries value for a rule.

The crash report is in de debug section.

### Affected Resources or Data Sources

resource/nsxt_policy_security_policy

### Terraform Configuration

I've reduced the code to the meaningful parts. The goal is to load a YAML backup of NSX security policies and rules, and apply them elsewhere using terraform/tofu.

Here's a subset of the YAML input backup : 

```yaml
---
#
# Security policies with their firewall rules
nsx_security_policies:
  - id: "nsx-mgmt"
    display_name: "nsx-mgmt"
    category: "Infrastructure"
    description: "Blah"
    stateful: true
    tcp_strict: true
    scope: ['ANY']
    rules:
      - id: "icmp"
        display_name: "icmp"
        description: "Allows ALL ICMP"
        tag: "mgmt-icmp"
        logged: false
        source_groups: ['ANY']
        sources_excluded: false
        destination_groups: ['ANY']
        destinations_excluded: false
        services: ['/infra/services/ICMP-ALL']
        direction: "IN_OUT"
        action: "ALLOW"
        disabled: false
        scope: ['ANY']
        ip_protocol: "IPV4_IPV6"
```

The above YAML will produce a `service_entries {}` block with the HCL code below because this particular rule has no `service_entries` attribute, and I believe this is what causes the provider to crash at apply time, although at plan time there's no error reported : 

```hcl
data "local_file" "nsx_source_configuration_file" {
  filename = "${path.module}/conf/nsx-config-backup.yml"
}

locals {
  nsx_source_config = yamldecode(data.local_file.nsx_source_configuration_file.content)
[...]

PR #1783, opened 2025-08-12
Handle empty service_entries block case in nsxt_policy_security_policy

ksamoray:
When a rule was created with an empty service_entries block, the provider crashed. This change addresses that bug.

Fixes: #1782


salv-orlando (approved):
LGTM

PR #1786, opened 2025-08-13
Remove Computed attribute from Project's default_span_path attribute

ksamoray:
Terraform fails to detect a change in this attribute - this seems to be a known TF issue:
[https://github.com/pulumi/pulumi-terraform-bridge/issues/2152](https://github.com/pulumi/pulumi-terraform-bridge/issues/2152)

To work around this, removed the Computed attribute from this attribute.


nsxt/resource_nsxt_policy_project.go
```diff
@@ -109,41 +109,30 @@ func resourceNsxtPolicyProject() *schema.Resource {
 					},
 				},
 			},
-			"vpc_deployment_scope": {
+			"default_span_path": {
+				Type:         schema.TypeString,
+				Description:  "Policy path of the Cluster based default Span object of type NetworkSpan",
+				ValidateFunc: validatePolicyPath(),
+				Optional:     true,
+				Default:      "/infra/network-spans/default",
```
salv-orlando, line 117: Ok, so we're back at this - which was the initial solution from @aruntony005 
It's probably not ideal as we will be in a bad position if NSX API behaviour changes in the future.

But if there's nothing better we can do, I can live with it

nsxt/resource_nsxt_policy_project.go
```diff
@@ -220,31 +209,28 @@ func resourceNsxtPolicyProjectPatch(connector client.Connector, d *schema.Resour
...
-	vdsIntface, ok := d.GetOk("vpc_deployment_scope")
-	if util.NsxVersionHigherOrEqual("9.1.0") && ok && len(vdsIntface.([]interface{})) > 0 && vdsIntface.([]interface{})[0] != nil {
+	if util.NsxVersionHigherOrEqual("9.1.0") {
 		// There should be just one object here
-		vdScope := vdsIntface.([]interface{})[0].(map[string]interface{})
 		var spanReferences []model.SpanReference
-		if _, ok := vdScope["default_span_path"]; ok && vdScope["default_span_path"] != "" {
-			defaultSpanPath := vdScope["default_span_path"].(string)
+		defaultSpanPath := d.Get("default_span_path").(string)
+		if defaultSpanPath != "" {
```
salv-orlando: If the user provides an empty string for default_span_path we will not send a default span in the payload to NSX.
Then NSX will set automatically /infra/network-spans/default.

Will this cause a permadiff?

ksamoray: The user can't provide an empty string as it'll fail validation (we validate that it's a policy path).
So defaultSpanPath can never be empty actually, I should fix that `if()` statement.

PR #1816, opened 2025-08-31
Use singular naming for range attribute

ksamoray:
When the API has a list of complex objects, we use singular naming, e.g

```
range {
  start = ...
  end = ...
}
```
instead of:
```
ranges {
  start = ...
  end = ...
}
```
which is confusing - as each instance is represented by a separate block.


poojav25 (approved):

poojav25:
LGTM

PR #1818, opened 2025-09-02
[DOC] Fix example usages in policy_up_address_allocation resource

salv-orlando:
This change:
- Removes context project reference from 1st example as it should not be leveraging NSX multi-tenancy
- Fixes reference to project data source in Multi-Tenancy example to consistently refer to the "demoproj" data source


poojav25:
LGTM

PR #1819, opened 2025-09-04
Added a retry and timeout feature to the policy search used in the data sources. Also updated the share data source to filter by id.

aruntony005:


nsxt/policy_search.go
```diff
@@ -108,28 +113,59 @@ func policyDataSourceResourceRead(d *schema.ResourceData, connector client.Conne
...
+	readTimeoutStr, retryCheck := os.LookupEnv("NSXT_READ_RETRY_TIMEOUT_SECONDS")
+	if retryCheck {
+		readTimeoutSeconds, err = strconv.Atoi(readTimeoutStr)
+		if err != nil {
+			return nil, fmt.Errorf("Error reading NSXT_READ_RETRY_TIMEOUT_SECONDS: %v", err)
 		}
 	} else {
-		resultValues, err = listPolicyResourcesByNameAndType(connector, context, objName, resourceType, &additionalQueryString, isGlobal)
+		// This timeout is for the non retryable case. So it fails with the first error.
+		readTimeoutSeconds = 60
```
salv-orlando: Maybe it's better to set a smaller default timeout, for instance 10 seconds.
This will ensure we don't wait for a minute in cases where the object that we try to fetch really does not exist.

docs/data-sources/policy_share.md
```diff
@@ -8,6 +8,7 @@ description: A Share data source.
 
 This data source provides information about a Share configured on NSX.
 This data source is applicable to NSX Policy Manager.
+Note: The env variable NSXT_READ_RETRY_TIMEOUT_SECONDS can be set to add a retry while reading the resources.
```
ksamoray, line 11: This comment is general and not specific to the policy_share DS.
Maybe it belongs elsewhere, e.g index.md?

PR #1820, opened 2025-09-05
[3.10.0]  Tier1/Segment DS: Allow fetching global resources, move is_global into context

salv-orlando:
This change adds the ability to fetch globally created Tier-1 gateways and segments, using the relvant data sources, similarly to what is already available for the nsxt_policy_tier0_gateway data source.

This change also moves the is_global attribute into the context, renaming it to "from_global"

To this aim this change:
- adds the context schema to the T0 GW data source
- expands the context schema of the T1 GW data source to include from_global
- adds a from_global boolean setting in the context fro both T0 and T1 data sources
- changes TF session context generation to set the from_global setting. When this setting is true, project_id and vpc_id will be ignored as global resournces cannot have a multitenancy scope

### DIFF
--- a/api/utl/api_util.go
+++ b/api/utl/api_util.go
@@ -18,12 +18,21 @@ type SessionContext struct {
 	ClientType ClientType
 	ProjectID  string
 	VPCID      string
+	FromGlobal bool
 }
 type ClientContext struct {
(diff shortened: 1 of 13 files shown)


poojav25:
LGTM

PR #1821, opened 2025-09-05
Adding the property dns_server_preference which was added to the NSX API

aruntony005:


ksamoray:
@aruntony005 consider using `IntroducedInVersion` in the metadata block - I'm not sure that this will resolve the issue where this attribute is marked as required but isn't available before v9.1.
If that won't help, maybe mark this as optional and validate within the code that it's set for 9.1 and beyond.

docs/resources/vpc_subnet.md
```diff
@@ -81,6 +81,7 @@ The following arguments are supported:
             * `value` - (Required) value for vendor-specific configuration
     * `enable_vlan_extension` - (Optional) Enabling VLAN connection for the subnet. The user must configure the exclusive IP block for this subnet. This IP block must be leveraged by a distributed VLAN connection. The default value for this will be false. This attribute is supported with NSX 9.1.0 onwards.
 * `dhcp_config` - (Optional) DHCP configuration block
+    * `dns_server_preference` - (Required) DNS server IP preference. Select the preference between the DNS server IPs (from the DHCP config in VPC service profile), and VPC DNS forwarder IP. The preferred DNS server IP config will be attempted first when the system selects the DNS server to forward DNS requests. This can be one of `PROFILE_DNS_SERVERS_PREFERRED_OVER_DNS_FORWARDER`, `DNS_FORWARDER_PREFERRED_OVER_PROFILE_DNS_SERVERS`.
```
salv-orlando: Please also indicate PROFILE_DNS_SERVERS_PREFERRED_OVER_DNS_FORWARDER is the default value

aruntony005: Have updated it

PR #1822, opened 2025-09-07
Correct network span doc title

ksamoray:
Seems like due to copy paste error this was titled wrongly.
Also reorder datasource list to keep it sorted.


poojav25:
LGTM

PR #1824, opened 2025-09-09
Check object sorting in provider.go

ksamoray:
The list of resources and datasources in provider.go is sorted. Added a script and a github action to check that this sorting is maintained.


salv-orlando (approved):
LGTM!

PR #1831, opened 2025-09-10
[DNM] test share and uplink datasources

ksamoray:


nsxt/data_source_nsxt_policy_share.go
```diff
@@ -15,7 +17,7 @@ func dataSourceNsxtPolicyShare() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"id":           getDataSourceIDSchema(),
 			"path":         getPathSchema(),
-			"display_name": getDataSourceDisplayNameSchema(),
+			"display_name": getDisplayNameSchema(),
```
salv-orlando: It seems you are reverting the change @aruntony005 just made in PR #1832, I don't think you're doing this intentionally?

nsxt/data_source_nsxt_policy_share.go
```diff
@@ -25,9 +27,10 @@ func dataSourceNsxtPolicyShareRead(d *schema.ResourceData, m interface{}) error
 	connector := getPolicyConnector(m)
 
 	_, err := policyDataSourceResourceRead(d, connector, getSessionContext(d, m), "Share", nil)
-	if err == nil {
-		return nil
+	if isNotFoundError(err) {
+		return fmt.Errorf("Share with name '%s' was not found", d.Get("display_name").(string))
+	} else if err != nil {
+		return fmt.Errorf("encountered an error while searching Share with name '%s', error is %v", d.Get("display_name").(string), err)
```
salv-orlando, line 33: nit: For error messages we usually start the string with a capital letter

ksamoray: Actually it's a common pattern not to start with a capital (or end with a punctuation):
https://go.dev/wiki/CodeReviewComments#error-strings

nsxt/data_source_nsxt_policy_share.go
ksamoray: Yeah that's a mistake, I'll fix this.

nsxt/data_source_nsxt_policy_share.go
salv-orlando: Yes, that why I wrote "we" :)

But it's ok, np

Issue #1827, opened 2025-09-10
Interface conversion error when applying config to VLAN Backed Segments


TuaEsque:
### Terraform

3.9.0

### Terraform Provider

vmware/nsxt

### VMware NSX

9.0.0.0.24733063

### Description

When trying to apply configuration to VLAN backed segments in NSX, I get the below panic.

### Affected Resources or Data Sources

resource/nsxt_policy_vlan_segment

### Terraform Configuration

main.tf

```
data "nsxt_policy_transport_zone" "vlan_transport_zone" {
  display_name   = var.vlan_segment.transport_zone_path
  transport_type = "VLAN_BACKED"
}

data "nsxt_policy_spoofguard_profile" "spoof_prof1" {
  count        = var.vlan_segment.security_profile != null ? 1 : 0
  display_name = var.vlan_segment.security_profile.spoofguard_profile_path != null ? var.vlan_segment.security_profile.spoofguard_profile_path : null
}

data "nsxt_policy_segment_security_profile" "seg_sec_prof1" {
  count        = var.vlan_segment.security_profile != null ? 1 : 0
  display_name = var.vlan_segment.security_profile.security_profile_path != null ? var.vlan_segment.security_profile.security_profile_path : null
}

data "nsxt_policy_ip_discovery_profile" "ip_disco_prof1" {
  count        = var.vlan_segment.discovery_profile != null ? (var.vlan_segment.discovery_profile.ip_discovery_profile_path != null ? 1 : 0) : 0
  display_name = var.vlan_segment.discovery_profile.ip_discovery_profile_path
}

data "nsxt_policy_mac_discovery_profile" "mac_disco_prof1" {
  count        = var.vlan_segment.discovery_profile != null ? (var.vlan_segment.discovery_profile.mac_discovery_profile_path != null ? 1: 0) : 0
  display_name = var.vlan_segment.discovery_profile.mac_discovery_profile_path
}

data "nsxt_policy_qos_profile" "qos_prof1" {
  count        = var.vlan_segment.qos_profile != null ? 1 : 0
  display_name = var.vlan_segment.qos_profile.qos_profile_path != null ? var.vlan_segment.qos_profile.qos_profile_path : null
}


resource "nsxt_policy_vlan_segment" "vlan_segment" {
  display_name        = var.vlan_segment.display_name
  description         = var.vlan_segment.description
  domain_name         = var.vlan_segment.domain_name
[...]

PR #1837, opened 2025-09-15
Interface conversion error when applying config to VLAN Backed

ksamoray:
This is caused by adding an empty "l2_extesion" block to the segment.

Fixes: #1827


poojav25:
LGTM

salv-orlando (approved):
LGTM

PR #1839, opened 2025-09-17
Added a function to fetch and set the default span in the project resource, if not provided in resource spec

aruntony005:


nsxt/resource_nsxt_policy_project.go
```diff
@@ -254,6 +263,25 @@ func resourceNsxtPolicyProjectPatch(connector client.Connector, d *schema.Resour
...
+		return "", err
+	}
+	defaultSpanPath := []string{}
+	for _, spanObj := range spanList.Results {
+		if *spanObj.IsDefault {
+			defaultSpanPath = append(defaultSpanPath, *spanObj.Path)
+		}
+	}
+	if len(defaultSpanPath) > 1 {
+		return "", fmt.Errorf("Found more than one span marked as default")
```
salv-orlando: Did you find any case where this occurred?

nsxt/resource_nsxt_policy_project.go
```diff
@@ -212,8 +212,17 @@ func resourceNsxtPolicyProjectPatch(connector client.Connector, d *schema.Resour
 	if util.NsxVersionHigherOrEqual("9.1.0") {
 		// There should be just one object here
 		var spanReferences []model.SpanReference
-		defaultSpanPath := d.Get("default_span_path").(string)
-
+		defaultSpanPathinfce, isDefaultSet := d.GetOkExists("default_span_path")
```
salv-orlando: Sorry for being pedant, can you pick a variable name which is more readable than `defaultSpanPathinfce`?
Perhaps `defaultSpanPathFromSchema`?

nsxt/resource_nsxt_policy_project.go
aruntony005: No, for now

nsxt/resource_nsxt_policy_project.go
aruntony005: Have updated the variable

nsxt/resource_nsxt_policy_project.go
aruntony005: Have removed the check

Issue #1815, opened 2025-08-29
Add support for a segment port data source


benzander:
### Description

In order to use a segment port profile resource as requested here: https://github.com/vmware/terraform-provider-nsxt/issues/1814 I also need a way to find the correct segment port ID.
So we need a segment port data source to identify the id of a vms segment port.

### Use Case(s)

see https://github.com/vmware/terraform-provider-nsxt/issues/1814

### Potential Configuration

.

PR #1854, opened 2025-10-10
Adding a data source for segment port

aruntony005:
Fixes: #1815


ksamoray (approved):
LGTM. I understand that we can't cover this with acceptance tests - but did we test this in any way?

Issue #1814, opened 2025-08-29
Add Segment Port Profile Ressource


benzander:
### Description

I need a ressource to overwrite the inherited segment profiles for specific ports.
I'm sure that such a ressource would need the segment port id as input value.
So, I also need a data source to find a VMs port IDs

<img width="1395" height="1175" alt="Image" src="https://github.com/user-attachments/assets/0aa01955-921a-4b86-95de-31053fd0d2b0" />

### Use Case(s)

We use our preferred segment profiles in all segments. Those include e.g. duplicate IP detection and spoofguard.
But there are some use cases, e.g. clusters with VMs that share a virtual IP Address, where I need to disable duplicate IP detection.
Today we need to move those VMs into a different segment with other segment profiles. It would be a lot easier and more straight forward when I had the possibility to overwrite the profiles per port.

### Potential Configuration

.

PR #1856, opened 2025-10-14
Added nsxt_policy_segment_port resource with the CRUD functionalities.

aruntony005:
Added nsxt_policy_segment_port resource with the CRUD functionalities.

Updated the nsxt_policy_vm data source to return vif attachment IDs of the VMs.

Added nsxt_policy_segment_port data source to get the segment ports with the display name or the vif id from the VMs data source.

Updated the T1 segment port client code.

Added acceptance tests for segment port resource/data source and the changes made in the vms data source.

Updated the documentations for the above changes.

Fixes: #1814


nsxt/resource_nsxt_policy_segment_port.go
```diff
@@ -0,0 +1,223 @@
...
+			"hyperbus_mode": {
+				Type:         schema.TypeString,
+				Description:  "ID used to identify/look up a child attachment behind a parent attachment",
+				Optional:     true,
+				Default:      "DISABLE",
+				ValidateFunc: validation.StringInSlice([]string{"ENABLE", "DISABLE"}, false),
+			},
+			"type": {
+				Type:         schema.TypeString,
+				Description:  "Type of port attachment. PARENT type is automatically set if evpn_vlans or hyperbus_mode is configured. INDEPENDENT type is automatically set for ports that belong to Segment of type DVPortgroup. STATIC type is deprecated.",
```
salv-orlando: Probably this one should be computed too, since in some cases is set by the backend.

nsxt/segment_port_common.go
```diff
@@ -0,0 +1,692 @@
...
+	if child != nil {
+		children = append(children, child)
+	}
+
+	obj.Children = children
+	return nil
+
+}
+
+func nsxtPolicyPortDiscoveryProfileSetInStruct(d *schema.ResourceData) (*data.StructValue, error) {
```
salv-orlando: Question: how feasible would be to generalize the logic in segment_common.go so that we can use it for setting profiles either in ports or segments?

nsxt/data_source_nsxt_policy_vm.go
```diff
@@ -29,6 +29,14 @@ func dataSourceNsxtPolicyVM() *schema.Resource {
 			"instance_id":  getDataSourceStringSchema("Instance UUID of the Virtual Machine"),
 			"tag":          getTagsSchema(),
 			"context":      getContextSchema(false, false, false),
+			"vif_ids": {
```
ksamoray, line 32: Update here is to the `nsxt_policy_vm` data source, yet documentation and test changes are to `nsxt_policy_vms`. Is that intentional?

docs/resources/policy_segment_port.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `display_name` - (Required) Display name of the resource.
+* `description` - (Optional) Description of the resource.
+* `tag` - (Optional) A list of scope + tag pairs to associate with this policy.
+* `nsx_id` - (Optional) The NSX ID of this resource. If set, this ID will be used to create the resource.
+* `segment_path` - (Optional) Path of the segment port.
```
ksamoray: This is required in the code. Doc is inconsistent.

docs/data-sources/policy_segment_port.md
```diff
@@ -0,0 +1,31 @@
...
+
+```hcl
+data "nsxt_policy_segment_port" "segmentport1" {
+  display_name = "segport1"
+}
+```
+
+## Argument Reference
+
+* `id` - (Optional) The ID of Segment Port to retrieve.
```
ksamoray, line 36: Code has context for multitenancy, it's missing from doc. We also include a multitenancy example wherever it's supported.

docs/resources/policy_segment_port.md
```diff
@@ -0,0 +1,68 @@
...
+    security_profile_path   = data.nsxt_policy_segment_security_profile.segprofile.path
+  }
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `display_name` - (Required) Display name of the resource.
```
ksamoray, line 58: Should we have a context here?

nsxt/resource_nsxt_policy_segment_port.go
```diff
@@ -0,0 +1,223 @@
...
+func resourceNsxtPolicySegmentPort() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNsxtPolicySegmentPortCreate,
+		Read:   resourceNsxtPolicySegmentPortRead,
+		Update: resourceNsxtPolicySegmentPortUpdate,
+		Delete: resourceNsxtPolicySegmentPortDelete,
+		Importer: &schema.ResourceImporter{
+			State: getSegmentPortPathOrIDResourceImporter,
+		},
+		Schema: map[string]*schema.Schema{
```
ksamoray, line 20: Should we have a context here?

docs/resources/policy_segment_port.md
aruntony005: Have updated the doc

docs/data-sources/policy_segment_port.md
aruntony005: Have removed the context from the code as segments and segment ports can be only created in the default project.

docs/resources/policy_segment_port.md
aruntony005: No, because segments and segment ports can be only created in the default project.

nsxt/resource_nsxt_policy_segment_port.go
aruntony005: No, because segments and segment ports can be only created in the default project.

nsxt/data_source_nsxt_policy_vm.go
aruntony005: No, I have updated the docs and tests

nsxt/data_source_nsxt_policy_segment_port.go
```diff
@@ -0,0 +1,82 @@
...
+		},
+	}
+}
+
+func dataSourceNsxtPolicySegmentPortRead(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+	converter := bindings.NewTypeConverter()
+	var segmentPortList []*data.StructValue
+	var err error
+	if vifId, exists := d.GetOkExists("vif_id"); exists {
```
ksamoray, line 41: You forgot to implement the id argument

go.mod
```diff
@@ -77,4 +77,4 @@ require (
 	google.golang.org/grpc v1.75.1 // indirect
 	google.golang.org/protobuf v1.36.9 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
-)
+)
```
ksamoray: Change to go.mod, go.sum is not required here. Please revert.

nsxt/data_source_nsxt_policy_segment_port.go
aruntony005: Have addressed this.

go.mod
aruntony005: Have added the empty line to avoid diff

nsxt/data_source_nsxt_policy_segment_port.go
```diff
@@ -69,6 +69,7 @@ func dataSourceNsxtPolicySegmentPortRead(d *schema.ResourceData, m interface{})
 	d.Set("display_name", segmentPort.DisplayName)
 	d.Set("description", segmentPort.Description)
 	d.Set("path", segmentPort.Path)
+	d.Set("vif_id", segmentPort.Attachment.Id)
```
ksamoray: Attachment is a pointer and can be null. Add a check here that `segmentPort.Attachment` is not null here.

PR #1859, opened 2025-10-19
Handle non-triggered upgrade prechecks

ksamoray:
A user specifies a list of prechecks to be acknowledged. Each of these prechecks is attempted to be acknowledged, but the acknowledge resource does not verify that the precheck actually failed. This change will verify that the precheck actually failed before acknowledging it.
Also verify that all precheck ids provided are valid.

### DIFF
--- a/nsxt/resource_nsxt_upgrade_precheck_acknowledge.go
+++ b/nsxt/resource_nsxt_upgrade_precheck_acknowledge.go
@@ -6,6 +6,8 @@ package nsxt
 
 import (
 	"fmt"
+	"log"
+	"slices"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	nsxModel "github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/model"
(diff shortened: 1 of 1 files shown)


nsxt/resource_nsxt_upgrade_precheck_acknowledge.go
```diff
@@ -78,12 +107,27 @@ func resourceNsxtUpgradePrecheckAcknowledgeCreate(d *schema.ResourceData, m inte
...
-			msg := fmt.Sprintf("Failed to acknowledge precheck warning with ID %s", precheckID)
-			return logAPIError(msg, err)
+		if slices.Contains(warns, precheckID) {
+			err := client.Acknowledge(precheckID)
+			if err != nil {
+				msg := fmt.Sprintf("Failed to acknowledge precheck warning with ID %s", precheckID)
+				return logAPIError(msg, err)
+			}
+		} else {
+			log.Printf("Precheck warning with ID %s has not been triggered by NSX", precheckID)
```
salv-orlando: Do we need to prepend [INFO] (or [DEBUG]) to the string being logged?

nsxt/resource_nsxt_upgrade_precheck_acknowledge.go
```diff
@@ -118,7 +162,11 @@ func setPrecheckWarningsInSchema(d *schema.ResourceData, precheckWarnings []nsxM
 func resourceNsxtUpgradePrecheckAcknowledgeUpdate(d *schema.ResourceData, m interface{}) error {
 	id := d.Id()
 	precheckIDs := interface2StringList(d.Get("precheck_ids").([]interface{}))
-	err := acknowledgePrecheckWarnings(m, precheckIDs)
+	err := validatePrecheckIDs(m, precheckIDs)
```
salv-orlando: Can we call this only if there been a schema change in prechecks to ack?

salv-orlando (approved):
LGTM

PR #1860, opened 2025-10-20
Remove default audit username from policy edge TN

ksamoray:
NSX requires a password when this attribute contains a value. As Terraform will include this always - as it has a default, we shouldn't include it in the payload unless specified by the user.


salv-orlando (commented):
Kobi, please also update the docs

salv-orlando:
@ksamoray does this mean that audit_username should be also optional & computed?

salv-orlando (approved):
LGTM

PR #1862, opened 2025-10-21
Fixing the default_span_path dynamic diff issue using CustomizeDiff

aruntony005:


nsxt/resource_nsxt_policy_project_test.go
```diff
@@ -91,6 +91,7 @@ func runChecksNsx910(testResourceName string, expectedValues map[string]string)
 	return resource.ComposeTestCheckFunc(
 		resource.TestCheckResourceAttr(testResourceName, "non_default_span_paths.#", expectedValues["span_reference_count"]),
 		resource.TestCheckResourceAttrSet(testResourceName, "non_default_span_paths.0"),
+		resource.TestCheckResourceAttrSet(testResourceName, "default_span_path"),
```
salv-orlando: Please add an acceptance test where we explicitly update a project resource to remove the default span path, so we won't regress anymore on this thing

aruntony005: Have added separate acceptance tests for CustomizeDiff solution

nsxt/provider.go
```diff
@@ -1168,6 +1169,11 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	if err != nil {
 		return nil, err
 	}
+	connector := getPolicyConnector(clients)
+	clients.DefaultSpanPath, err = getDefaultSpan(connector)
```
ksamoray: @aruntony005 this should only run for v9.1 and above - we have no spans before this.
This is why you get that notfound error (that API doesn't exist in 9.0)

aruntony005: Have added version check wherever needed.

salv-orlando (approved):
LGTM

PR #1871, opened 2025-11-04
Skipping 404 response while getting default span as a temporary fix

aruntony005:
Skipping 404 response while getting default span as a temporary fix to make the tests pass until the relevant FSS is enabled on main


salv-orlando (approved):
LGTM

Issue #1874, opened 2025-11-10
Data Source nsxt_policy_groups doesn't follow cursor


derOtto:
### Terraform

OpenTofu v1.10.6

### Terraform Provider

3.10.0

### VMware NSX

4.2.1.3.0.24533884

### Description

Hi,
as I was suddenly missing some groups in the datasource nsxt_policy_groups.
I found out, that the datasource may not return all groups, when the page size of 1000 was hit.

It seems, as it won't follow the cursor: [nsxt/data_source_nsxt_policy_groups.go#L32-L50](https://github.com/vmware/terraform-provider-nsxt/blob/master/nsxt/data_source_nsxt_policy_groups.go#L32-L50)‽

### Affected Resources or Data Sources

datasource/nsxt_policy_groups

### Terraform Configuration

data "nsxt_policy_groups" "all" {}

### Expected Behavior

The Data Source should return a map with all groups.

### Actual Behavior

The Data Source returns only the groups fitting in the first page.

### Steps to Reproduce

1. Create more than 1000 groups within one domain.
2. Request / output the data source.
3. Some groups are missing.

PR #1876, opened 2025-11-10
Use cursor in nsxt_policy_groups data source

ksamoray:
When number of groups in a domain is over 1000 the results will be truncated.

Fixes: #1874

### DIFF
--- a/nsxt/data_source_nsxt_policy_groups.go
+++ b/nsxt/data_source_nsxt_policy_groups.go
@@ -35,15 +35,28 @@ func dataSourceNsxtPolicyGroupsRead(d *schema.ResourceData, m interface{}) error
 
 	client := domains.NewGroupsClient(getSessionContext(d, m), connector)
 
+	boolFalse := false
+	var cursor *string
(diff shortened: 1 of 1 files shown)


nsxt/data_source_nsxt_policy_groups.go
```diff
@@ -35,15 +35,28 @@ func dataSourceNsxtPolicyGroupsRead(d *schema.ResourceData, m interface{}) error
...
+			total = int(*results.ResultCount)
+		}
+		for _, r := range results.Results {
+			groupsMap[*r.DisplayName] = *r.Path
+		}
+		cursor = results.Cursor
+		if len(groupsMap) >= total {
+			break
+		}
+	}
```
salv-orlando, line 59: didn't we have a utility function for handling paginated responses?

ksamoray: AFAIK only for search API data sources, which isn't applicable here

salv-orlando (approved):
LGTM

PR #1878, opened 2025-11-11
Ignore gateway policy default rule order (#1781)

ksamoray:
When there are multiple default rules, NSX may return the rules in an incorrect order and cause permadiff.

Fixes: #1534


salv-orlando (approved):
LGTM

PR #1886, opened 2025-11-14
Adding parent gateway policy resource without rule to fix the rule mismatch issue when created using gateway_policy_rule resource

aruntony005:
When a gateway policy and a gateway policy rule is created seperately, the terraform state won't be in sync with the NSX backend. Because of this we will see a config diff even though no change is done to the terraform resource config. 
For such cases we are adding a new parent_gateway_policy resource which can be used if we plan to create the rules using the gateway_policy_rule resource.


salv-orlando:
> Should we use this resource in `resource_nsxt_policy_gateway_policy_test.go` instead of using `nsxt_policy_gateway_policy`?

I don't have a strong opinion, but this PR adds a test module for covering nsxt_parent_gateway_policy including addition of rules as separate resources, where as resource_nsxt_policy_gateway_policy_test has tests that cover creation of rules inside the policy.

So I'm leaning towards keeping tests in `resource_nsxt_policy_gateway_policy_test.go` as they are.

salv-orlando (approved):
LGTM

PR #1897, opened 2025-11-24
Updating the IP block, Distributed Vlan Connection and VPC subnet resources with the TEP-less 9.1.0 changes

aruntony005:


ksamoray:
@aruntony005 please change the PR title to reflect the objects which were changed, as there were many design changes in v9.1.0 :)

PR #1908, opened 2025-12-01
Address failures due to span behavior changes

ksamoray:
Assignment of a network span to a transit gateway requires that a span will have compute resources attached.


aruntony005:
LGTM

aruntony005 (approved):

PR #1925, opened 2025-12-10
[WIP] POC implementation of caching mechanism

poojav25:


nsxt/cache.go
```diff
@@ -0,0 +1,119 @@
+package nsxt
```
ksamoray, line 1: Can we place this in another package? `nsxt` is already overloaded

cache_benchmark/vpc_cache.tfbck
```diff
@@ -0,0 +1,139 @@
+terraform {
```
ksamoray, line 1: This file shouldn't be part of the PR

cache_benchmark/vpc_benchmark.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
```
ksamoray, line 1: This file shouldn't be part of the PR

cache_benchmark/vpc_bench_results.txt
```diff
@@ -0,0 +1,7 @@
+TIMESTAMP            VPC_COUNT  SUBNET_COUNT    PLAN_SECONDS    APPLY_SECONDS   STATUS     PLAN       CACHE_ENABLED
```
ksamoray, line 1: This file shouldn't be part of the PR

nsxt/cache.go
```diff
@@ -0,0 +1,119 @@
...
+	utl "github.com/vmware/terraform-provider-nsxt/api/utl"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+)
+
+type cache struct {
+	mu       sync.RWMutex
+	data     map[string]map[string]*data.StructValue //key: query , key DisplayName and objects
```
ksamoray, line 20: The mutex here is for the entire cache - I think that this is inefficient.
IMO data here can be a `map[string]<struct>
Where struct contains the map and a mutex - we have no issue with access to the cache of VpcSubnet while accessing the cache of some other type.
I guess that this will require some changes though as data is keyed by query string - IDK if this is doable.

nsxt/cache.go
```diff
@@ -0,0 +1,119 @@
...
+	c.data[query] = tmp
+	return nil
+}
+
+func isRefreshPhase(d *schema.ResourceData) bool {
+	return d.Id() != "" && !d.HasChangesExcept()
+}
+
+// SetCacheEnabled allows enabling/disabling cache for benchmarking
+var cacheEnabled = os.Getenv("NSXT_ENABLE_CACHE") == "true"
```
ksamoray, line 115: This should be also enabled/disabled by a provider-level attribute - you can use `allow_unverified_ssl` as an example for how to use both environment var and a provider attribute for activation

PR #1940, opened 2026-01-05
Fixing the attachment state sync issue and adding support for CHILD segment port

aruntony005:
* Fixing the attachment state sync issue.
* Adding support for CHILD segment port.
* Adding acceptance tests for the above.


salv-orlando (approved):
LGTM

PR #1945, opened 2026-01-09
Introduce in-memory cache for resource lookup optimization

poojav25:


nsxt/cache.go
```diff
@@ -0,0 +1,837 @@
...
+	postWriteByKey.Store(postWriteKey(resourceType, resourceID), struct{}{})
+}
+
+// MarkPostWriteForResourceTypeFromData marks a resource instance as just written.
+// Uses d.Get("path") when present, otherwise falls back to d.Id().
+func MarkPostWriteForResourceTypeFromData(resourceType string, d *schema.ResourceData) {
+	if d == nil {
+		return
+	}
+	if p, ok := d.GetOk("path"); ok {
```
salv-orlando: Post-write bypass is keyed by `path` here, but most Reads pass `d.Id()` as `resourceID` to `TryCacheRead`/`CacheAwareResourceRead` (e.g. `resourceNsxtPolicyGroupGeneralRead`, `IpAddressPoolStaticSubnet`/`BlockSubnet` Reads all use `id := d.Id()`). `path` is only empty on the very first Create, so every subsequent Update/Delete marks `type::path` while the lookup key is `type::id` — they never match, and the one-shot post-write bypass meant to guard against NSX Search's eventual consistency (see the comment near line 585) silently never fires after the first write.

nsxt/resource_nsxt_policy_security_policy_rule.go
```diff
@@ -185,6 +187,28 @@ func resourceNsxtPolicySecurityPolicyRuleRead(d *schema.ResourceData, m interfac
 	if client == nil {
 		return policyResourceNotSupportedError()
 	}
+
+	if isCacheEnabledForRead(d) {
+		sp, cacheUsed, _, cacheErr := TryCacheRead[model.SecurityPolicy](
```
salv-orlando, line 192: `policyID` here is the parent SecurityPolicy's ID, but `MarkPostWriteAndInvalidateCacheForResourceType("securitypolicy", d)` (lines 64/282/306) marks the post-write bypass keyed by this rule's own path/ID. The keys can never match, so after create/update/delete of a rule this always repopulates the just-invalidated `securitypolicy` bucket straight from NSX Search with no read-your-writes guard — risking a stale/incomplete rule list being cached for the rest of the run if Search hasn't indexed the write yet.

nsxt/resource_nsxt_policy_gateway_policy_rule.go
```diff
@@ -115,6 +118,28 @@ func resourceNsxtPolicyGatewayPolicyRuleRead(d *schema.ResourceData, m interface
 	if client == nil {
 		return policyResourceNotSupportedError()
 	}
+
+	if isCacheEnabledForRead(d) {
+		gp, cacheUsed, _, cacheErr := TryCacheRead[model.GatewayPolicy](
```
salv-orlando, line 123: Same key-mismatch as in `resource_nsxt_policy_security_policy_rule.go`: `MarkPostWriteAndInvalidateCacheForResourceType("gatewaypolicy", d)` (lines 66/176/200) marks the bypass by this rule's own path/ID, while `policyID` (the parent's ID) is what's looked up here — the post-write bypass never fires for GatewayPolicyRule writes either.

nsxt/cache.go
```diff
@@ -0,0 +1,837 @@
...
+	}
+	return ret
+}
+
+func getQueryString(resourceType string, context utl.SessionContext) string {
+	switch context.ClientType {
+	case utl.Global:
+		return fmt.Sprintf("resource_type:%s AND marked_for_delete:false AND context:Global", resourceType)
+	case utl.Local:
+		return fmt.Sprintf("resource_type:%s AND marked_for_delete:false AND context:Local", resourceType)
```
salv-orlando, line 298: This query has no notion of which NSX Manager/connector the result came from, and `gcache` (line 33) is a single process-wide singleton. Two aliased `nsxt` provider blocks pointing at different managers with the same `context_id` (defaults to `""` for both) produce an identical query string here, so a lookup for manager B's resource could be served stale data cached for manager A. Consider folding a manager/connector identity into the cache key.

nsxt/cache.go
```diff
@@ -0,0 +1,837 @@
...
+	if additionalQuery == "" {
+		return query
+	}
+	return fmt.Sprintf("%s AND %s ", query, additionalQuery)
+}
+
+func (c *typeScopedCache) readCache(resourceID string, resourceType string, d *schema.ResourceData, m interface{}, connector client.Connector) (interface{}, error) {
+	tc := c.getTypeCache(resourceType)
+
+	tc.mu.Lock()
```
salv-orlando, line 398: `tc.mu` is a `sync.RWMutex`, but this always takes the exclusive `Lock()`, even for pure cache-hit lookups, so every read of a given resource type fully serializes across goroutines. On a miss, the same lock is held through `writeCache`'s network round-trip to NSX Search, blocking all other reads (hits included) of that type until it returns. Under Terraform's default parallelism this can serialize resource reads more than the uncached path did, working against this PR's own performance goal. Consider `RLock()` for the hit path with a double-checked lock only around population.

nsxt/data_source_nsxt_policy_group.go
```diff
@@ -24,10 +27,41 @@ func dataSourceNsxtPolicyGroup() *schema.Resource {
...
 
 func dataSourceNsxtPolicyGroupRead(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+	objID := d.Get("id").(string)
+	displayName := d.Get("display_name").(string)
+	lookupKey := objID
+	if lookupKey == "" {
+		lookupKey = displayName
+	}
+
```
salv-orlando, line 30: This cache lookup is keyed only on `resourceType="Group"` plus the generic tag/context query built by `getCacheQueryKey` — it never folds in `domain` the way the non-cached path below does (`query["parent_path"] = "*/" + domain`). Two domains can each contain a Group with the same `display_name` (or even the same short `id` in MP-API-style setups); this cache-hit branch will happily `d.SetId`/`d.Set(...)` and `return nil` on the first match found in the process-wide `gcache`, regardless of which domain it actually belongs to. Consider folding `domain` into the cache query key here, or at minimum validating `obj.ParentPath`/`obj.Path` contains the expected domain segment before accepting the cache hit.

nsxt/data_source_nsxt_policy_service.go
```diff
@@ -23,7 +26,38 @@ func dataSourceNsxtPolicyService() *schema.Resource {
...
-	_, err := policyDataSourceResourceRead(d, getPolicyConnector(m), getSessionContext(d, m), "Service", nil)
+	connector := getPolicyConnector(m)
+	objID := d.Get("id").(string)
+	displayName := d.Get("display_name").(string)
+	lookupKey := objID
+	if lookupKey == "" {
+		lookupKey = displayName
+	}
+
+	if lookupKey != "" && IsCacheEnabled() {
```
salv-orlando: This inline cache-lookup block (repeated near-verbatim across ~10 data sources: Group, IpAddressPool, LBService, Service, Tier1Gateway, TransitGateway, Vpc, VpcConnectivityProfile, VpcServiceProfile, VpcSubnet) drops safety behavior that `policyDataSourceResourceFilterAndSet` (used by the non-cached fallback just below) provides when looking up by `display_name`:

- No prefix-match support — the uncached path matches on `strings.HasPrefix(name, objName)` as a fallback; the cache only does exact-key lookup via `converListToMapByType`'s indexing.
- No "multiple matches" error — the uncached path explicitly fails with `"Found multiple %s with name '%s'"` when more than one resource shares a name/prefix; the cache path just returns whichever object happens to be indexed under that display_name key (`indexCacheMapKeySecondary` keeps "first-seen" on collisions), silently picking one.

For configurations that rely on `display_name` lookups (no `id` set), this changes user-visible behavior (silent wrong-resource selection instead of a clear ambiguity error) purely based on whether `NSXT_CACHE_MODE` is enabled. Given the duplication across ~10 files, this might also be a good candidate for a shared helper rather than repeating the block per data source.

nsxt/cache.go
salv-orlando: More details on this:

d.GetOk("path") doesn't ask "is this resource keyed by path" — it asks "does the schema have a path attribute, and is it currently non-empty in state." Almost every policy resource in this provider sets path in its schema (via getPathSchema()) So:

For Group (id = short UUID, but schema also exposes path): after the first Create, path gets set on Read → every subsequent Update/Delete marks the bypass under Group::<path>, while resourceNsxtPolicyGroupGeneralRead looks up under Group::<id>. 

This will cause the mismatch, and it's the common case for any resource that has a path attribute in its schema.

The only resources where this branch would coincidentally work are ones where d.Id() itself is set to the path string .

MarkPostWriteForResourceTypeFromData has no way to know what key resourceID the Read function will use for this exact resource type, because that's a decision made independently at each call site (CacheAwareResourceRead[T](..., resourceID, ...)). There's no schema convention enforced that ties "what key does Read use" to "what does path/id look like" — it's just whatever the resource author wrote in that Read function. Guessing from schema contents can't reconstruct that decision.

To fix this:
Option 1

explicit key parameter.
 Change the signature so the caller states the key directly, collapsing the two marker functions into the one that already exists and is correct: MarkPostWriteForResourceTypeKey(resourceType, resourceID). Then MarkPostWriteAndInvalidateCacheForResourceType becomes:

func MarkPostWriteAndInvalidateCacheForResourceType(resourceType, resourceID string) {
    MarkPostWriteForResourceTypeKey(resourceType, resourceID)
    InvalidateCacheForResourceType(resourceType)
}
Every call site (MarkPostWriteAndInvalidateCacheForResourceType("Group", d)) becomes MarkPostWriteAndInvalidateCacheForResourceType("Group", id) where id is the exact same variable already in scope and already passed to CacheAwareResourceRead/TryCacheRead a few lines below. 

This is mechanical, low-risk to applly and it makes correctness self-evident at the call site instead of hidden inside a heuristic.

Option 2 (if we don't want to change the method signature and make changes in all the callers)
Keep the MarkPostWriteForResourceTypeFromData(resourceType, d) signature, but replace the GetOk("path")-then-Id() heuristic with a lookup against the same kind of registry cache.go already uses for shouldIndexByPath(resourceType) (line 156) — i.e., resourceType deterministically maps to "this type's Read keys by path" or "by id," matching whatever converListToMapByType indexes it by. This preserves the current call-site shape (MarkPostWriteAndInvalidateCacheForResourceType("Group", d)) so it's a smaller diff at each of the ~50 call sites, but it introduces a second source of truth that has to be kept in sync with shouldIndexByPath and with whatever literal resourceID each Read function passes — which is exactly the kind of implicit coupling that caused this bug in the first place.

nsxt/cache.go
salv-orlando: This comments is a valid and refers to a configuration we have not validated while testing this patch (and that to the best of my knowledge we never validated).

See the following example - basically a single TF config that manages multiple NSX managers via aliased providers.

```
provider "nsxt" {
  alias    = "mgr_a"
  host     = "nsx-mgr-a.corp.com"
  # context_id left unset -> ""
}

provider "nsxt" {
  alias    = "mgr_b"
  host     = "nsx-mgr-b.corp.com"
  # context_id left unset -> ""
}

resource "nsxt_policy_group" "g1" {
  provider = nsxt.mgr_a
  ...
}

resource "nsxt_policy_group" "g2" {
  provider = nsxt.mgr_b
  ...
}
```

Terraform instantiates one nsxt plugin binary and multiplexes both aliases through it via distinct nsxtClients/connector instances, so gcache — the package-level var gcache = &typeScopedCache{...} — is genuinely shared between g1 and g2's Reads. Nothing about ClientType/ProjectID/VPCID/tags distinguishes "the Local-manager context on host A" from "the Local-manager context on host B," so the second Read to populate/consult the Group bucket sees the first alias's cached data if the query string matches, which it will by default.

Potential fixes:

Cheapest fix, preserves your design as-is: fold something manager-identifying into the cache key/gcache sharding — e.g. hash of connector's target host, or require context_id to be validated as non-empty (and unique-ish) whenever the provider config detects it's one of multiple aliased instances. This doesn't change global-mode semantics at all (still deliberately unscoped within one manager) — it only prevents collisions across managers, which was never the intent to allow.
Cheaper still, if multi-manager + caching together is out of scope for v1: explicitly detect and warn/error when IsCacheEnabled() is true and more than one distinct NSX Manager host is in play within the same provider process — i.e., make the current unsafe-by-omission behavior fail loudly instead of silently. That's a much smaller change and matches "we didn't design for this yet" rather than pretending the collision can't happen.

nsxt/data_source_nsxt_policy_service.go
salv-orlando: Follow-up with some concrete options for closing the gap described above, roughly in order of preference:

**Option 1 — extract a shared helper that mirrors `policyDataSourceResourceFilterAndSet`'s semantics.** Add something like `cacheAwareDataSourceRead(d, m, connector, resourceType, bindingType) (*T, error)` in `cache.go` that, on a `display_name` lookup, doesn't just call `gcache.readCache(lookupKey, ...)` for a single key — it reads the whole populated bucket for the current query (the map is already fully materialized in `resourceTypeCache.data[query]` after the first miss) and replicates the exact-match/prefix-match/multi-match-error logic that `policyDataSourceResourceFilterAndSet` already implements in `policy_search.go`. This would need a small new accessor on `resourceTypeCache` (e.g. `allValues(query string) []*data.StructValue`) alongside the existing single-key `getQueryResult`. This removes the ~10x duplication *and* restores the dropped safety behavior in one pass, rather than fixing each data source's copy-pasted block individually.

**Option 2 — narrow the cache fast-path to ID-only lookups.** If a full reimplementation of prefix/multi-match against the cache is more than wants to land now, the cheaper fix is to gate the cache branch on `objID != ""` only (drop the `displayName` fallback into `lookupKey` for the cache path), and let any `display_name`-based lookup always fall through to `policyDataSourceResourceRead`. This forfeits some of the cache's benefit for name-based data source lookups, but it's a small, low-risk change that eliminates the silent-wrong-match risk entirely rather than trying to reproduce prefix/ambiguity semantics against the cache's exact-match index.

**Option 3 — status quo plus a warning.** If neither is feasible short-term, at minimum add a `log.Printf("[WARN] ...")` when the cache path is taken via `display_name` (not `id`), so users debugging an unexpected resource selection have a lead, and call out the behavior difference in the docs/changelog for `NSXT_CACHE_MODE`. This doesn't fix the correctness gap, just makes it debuggable.

Given the duplication is already 10 files deep, Option 1 seems like the best long-term investment, but Option 2 is the smallest safe patch if this needs to merge soon.

nsxt/policy_utils.go
```diff
@@ -204,6 +204,58 @@ func shouldIgnoreScope(scope string, scopesToIgnore []string) bool {
...
+	for _, t := range userTags {
+		if isManagedDefaultTag(t) {
+			continue
+		}
+		res = append(res, t)
+	}
+	return res
+}
+
+func getPolicyTagsWithProviderManagedDefaults(d *schema.ResourceData, m interface{}) []model.Tag {
```
salv-orlando, line 261: This function already calls `getPolicyTagsFromSchema(d)` internally, making it a strict superset of the plain call. Yet ~45 Create/Update call sites across the diff repeat this exact 6-line shape instead of calling this function directly:

```go
var tags []model.Tag
if isConfigScopedCacheMode() {
    tags = getPolicyTagsWithProviderManagedDefaults(d, m)
} else {
    tags = getPolicyTagsFromSchema(d)
}
```

Folding the mode check into this function would collapse every one of those call sites to a single line:

```go
func getPolicyTagsWithProviderManagedDefaults(d *schema.ResourceData, m interface{}) []model.Tag {
	userTags := getPolicyTagsFromSchema(d)
	if !isConfigScopedCacheMode() {
		return userTags
	}
	runID := m.(nsxtClients).CommonConfig.contextID
	if runID == "" {
		log.Printf("[DEBUG] context_id is empty; skipping provider-managed tag scope %q", managedDefaultTagScope)
	}
	managedDefaults := getProviderManagedDefaultTags(runID)
	return mergeManagedDefaultAndUserTags(managedDefaults, userTags)
}
```

removing ~150+ duplicated lines with no behavior change, and the duplication risk of one of the 45 sites drifting from the rest if this condition ever needs to change.

One thing to keep separate: there's a second, different shape using the same guard that should **not** be touched by this refactor — e.g. in `resource_nsxt_policy_security_policy.go` and `resource_nsxt_vpc_gateway_policy.go`:

```go
if isConfigScopedCacheMode() {
    _ = d.Set("tag", initPolicyTagsSet(getPolicyTagsWithProviderManagedDefaults(d, m)))
}
```

That variant writes back into schema state rather than consuming the return value through an if/else, so it serves a different purpose (seeding `d`'s tag set pre-Create) and should stay as-is.

nsxt/cache.go
```diff
@@ -0,0 +1,837 @@
...
+	}
+	rules := structValuesToRules(children)
+	merged := attachRulesByParentPath(sp, rules,
+		func(p model.SecurityPolicy) *string { return p.Path },
+		func(p *model.SecurityPolicy, r []model.Rule) { p.Rules = r },
+	)
+	return modelsToStructValues(merged, model.SecurityPolicyBindingType())
+}
+
+func CacheAwareResourceRead[T any](d *schema.ResourceData, m interface{}, connector client.Connector, resourceID string, resourceType string, bindingType bindings.BindingType, backendRead func() (*T, error), patchFunc func(obj *T) error) (*T, bool, bool, error) {
```
salv-orlando, line 731: **Readability/maintainability:** `TryCacheRead` (above, ~line 413) and `CacheAwareResourceRead` duplicate roughly 25 lines of logic almost verbatim — the post-write-bypass lookup, `gcache.readCache` call, converter/type-assert dance, the two `[WARNING]` log lines on conversion/assertion failure, and the `stripProviderManagedTagsFromAny` call on success. The only real difference between the two functions is what happens when the cache doesn't produce a usable value: `TryCacheRead` returns `nil`, `CacheAwareResourceRead` falls through to `backendRead()`.

Consider extracting the shared "try cache, convert, type-assert, strip tags" sequence into a helper both call, e.g.:

```go
func tryReadFromCache[T any](d *schema.ResourceData, m interface{}, connector client.Connector, resourceID, resourceType string, bindingType bindings.BindingType) (*T, bool, bool) {
    // returns (value, cacheUsed, cacheAttempted); nil value means "no usable cache hit"
}
```

`TryCacheRead` becomes a thin wrapper that also checks the post-write-bypass map and returns immediately; `CacheAwareResourceRead` calls the same helper and falls through to `backendRead()` on a miss. This would cut ~40 lines of duplicated logic and mean a future fix to the conversion/logging behavior (e.g. the type-assertion/conversion-failure logging) only needs to happen once.

nsxt/cache.go
```diff
@@ -0,0 +1,837 @@
...
+	cacheAttempted := false
+	postWriteBypass := false
+
+	if isRefreshPhase(d) && IsCacheEnabled() {
+		if _, ok := postWriteByKey.LoadAndDelete(postWriteKey(resourceType, resourceID)); ok {
+			// Ensure read-your-writes semantics for this resource type (one-shot).
+			// Search-backed cache may be briefly stale right after a write.
+			postWriteBypass = true
+			cacheAttempted = true
+			goto backend
```
salv-orlando: **Readability:** the `goto backend` / `backend:` label here is the only goto in the file and adds a control-flow jump that isn't needed — nothing in between requires jumping past loop/defer boundaries. This reads more directly as a guard computed up front:

```go
postWriteBypass := false
if isRefreshPhase(d) && IsCacheEnabled() {
    if _, ok := postWriteByKey.LoadAndDelete(postWriteKey(resourceType, resourceID)); ok {
        postWriteBypass = true
        cacheAttempted = true
    } else {
        cacheAttempted = true
        // ... existing cache-hit lookup block, unchanged ...
    }
}

if postWriteBypass {
    log.Printf("[DEBUG] Cache post-write bypass: direct GET for resourceType=%s id=%s", resourceType, resourceID)
} else {
    log.Printf("[DEBUG] Cache fallback: direct GET for resourceType=%s id=%s", resourceType, resourceID)
}
obj, err := backendRead()
...
```

Same behavior, but a reader doesn't need to track a label across ~30 lines to see where control resumes. Minor, but worth cleaning up alongside the duplication with `TryCacheRead` noted above, since both changes touch this function's structure.

nsxt/cache.go
```diff
@@ -0,0 +1,837 @@
...
+func indexCacheMapKey(ret map[string]*data.StructValue, key string, obj *data.StructValue) {
+	if existing, seen := ret[key]; seen {
+		// Keep the first-seen mapping to avoid overwriting with duplicates.
+		_ = existing
+		return
+	}
+	ret[key] = obj
+}
+
+func indexCacheMapKeySecondary(ret map[string]*data.StructValue, key string, obj *data.StructValue) {
```
salv-orlando: **Readability:** `indexCacheMapKey` (above) and `indexCacheMapKeySecondary` are byte-for-byte identical — same "first-seen wins, ignore collisions" body, differing only in the doc comment describing which caller uses them (primary ID-based keys vs. best-effort display_name/path keys). Since the behavior is identical, consider collapsing to one function and, if the primary/secondary distinction is worth preserving for readers, noting it at the call sites in `converListToMapByType` instead (e.g. `// secondary key: display_name` above the relevant call) rather than via two near-identical function names.

nsxt/cache.go
```diff
@@ -0,0 +1,837 @@
...
+	return IsCacheEnabled() && isRefreshPhase(d)
+}
+
+func shouldIndexByPath(resourceType string) bool {
+	switch resourceType {
+	case "PolicyNatRule", "SegmentPort", "IpAddressPoolStaticSubnet", "IpAddressPoolBlockSubnet", "Service":
+		return true
+	default:
+		return false
+	}
```
salv-orlando, line 212: **Maintainability — worth doing despite the surface area, given it's mechanical:** the `resourceType` string is passed as a raw literal at ~50 call sites (`"Group"`, `"gatewaypolicy"`, `"securitypolicy"`, `"rule"`, `"IpAddressPoolStaticSubnet"`, etc. — 33 distinct values, ~190 occurrences total) and is used both as the cache's internal dictionary key *and* interpolated directly into the NSX Search query (`resource_type:%s` in `getQueryString`/`getListOfPolicyResources`). A typo at any call site doesn't fail to compile — it silently produces a query that returns zero/wrong results, or silently misses the `shouldIndexByPath`/`compositeCacheRegistry` special-casing for that type, because nothing here ties the literal to those switch/map keys at compile time.

Suggest introducing typed constants (e.g. a `resourceTypeGroup = "Group"`-style block, or a small `type cacheResourceType string` with named values) and updating:
- every `MarkPostWriteAndInvalidateCacheForResourceType(...)`, `CacheAwareResourceRead[...](...)`, `TryCacheRead[...](...)`, `InvalidateCacheForResourceType(...)` call site to pass the constant instead of a literal,
- the `compositeCacheRegistry` map keys (`"gatewaypolicy"`, `"securitypolicy"`) and `shouldIndexByPath`'s `switch` cases to use the same constants.

**Important constraint for whoever/whatever does this mechanically:** the literal *values* must not change, only how they're referenced — e.g. `"gatewaypolicy"`, `"securitypolicy"`, and `"rule"` are lowercase while most others are PascalCase, and that's not an inconsistency to "clean up," it's presumably what NSX Search's `resource_type` field actually stores for those types. Renaming call sites to use constants while silently normalizing casing would break Search queries for gateway/security policies and rules. This should be a pure find-and-replace of literal → constant with identical string values, verified by a diff showing zero literal value changes.

Given the mechanical nature of this change, this seems like a good candidate to hand to a coding agent rather than doing by hand.

nsxt/cache.go
poojav25: Good catch, the bypass key needs to match the id used in Read, I'll switch it to use the same resourceID directly.

nsxt/resource_nsxt_policy_security_policy_rule.go
poojav25: You're right, I'll pass the rule's own id instead of the parent policyID so the bypass key actually matches.

nsxt/resource_nsxt_policy_gateway_policy_rule.go
poojav25: Same issue as the security policy rule case, I'll fix the key here too.

nsxt/cache.go
poojav25: Fair point, we hadn't tested multi-manager aliasing, I'll look at folding the manager identity into the cache key.

nsxt/cache.go
poojav25: Agreed, I'll switch to RLock for hits and only take the exclusive lock when populating the cache.

nsxt/data_source_nsxt_policy_service.go
poojav25: Thanks for laying out the options, I'll go with extracting a shared helper so we don't lose the prefix/multi-match handling.

nsxt/policy_utils.go
poojav25: Makes sense, I'll fold the mode check into the function and clean up the duplicated call sites.

nsxt/cache.go
poojav25: Agreed, I'll extract the shared cache-read logic into one helper.

nsxt/cache.go
poojav25: You're right, the goto isn't needed, I'll rewrite it as a plain guard.

nsxt/cache.go
poojav25: Good point, I'll collapse these into a single function.

nsxt/cache.go
poojav25: Done, added typed constants for all the resource type strings and swapped every call site over to them.

nsxt/data_source_nsxt_policy_service.go
poojav25: Correction: this is already on option 2, the cache fast path only takes ID lookups, display_name always falls through to the uncached path with its prefix/multi-match handling intact.

nsxt/cache.go
```diff
@@ -0,0 +1,892 @@
...
+// MarkPostWriteAndInvalidateCacheForResourceType marks resourceID as just written and invalidates
+// the type's cache bucket. resourceID must be the same key used in the corresponding
+// CacheAwareResourceRead/TryCacheRead call (typically d.Id(), or the parent's ID for child resources).
+func MarkPostWriteAndInvalidateCacheForResourceType(resourceType, resourceID string) {
+	MarkPostWriteForResourceTypeKey(resourceType, resourceID)
+	InvalidateCacheForResourceType(resourceType)
+}
+
+func currentCacheMode() cacheMode {
+	raw := strings.TrimSpace(os.Getenv(envNSXTCacheMode))
```
ksamoray: Cache should be controlled as a provider attribute. Users do not configure using environment variables. Use the same practice which is used for NSXT_CONTEXT_ID in this PR.

ksamoray:
`MarkPostWriteAndInvalidateCacheForResourceType` is not called in multiple resource delete funcs. This is missing at:
resource_nsxt_policy_ip_pool.go, resource_nsxt_policy_connectivity_policy.go, dhcp_v4/v6_static_binding.go, distributed_flood_protection_profile_binding.go, group.go/vpc_group.go, ip_address_allocation.go, project_ip_address_allocation.go, resource_nsxt_vpc_*.go and maybe others. Is this on purpose?

I assume that this should be as done at resource_nsxt_policy_gateway_policy.go:322

nsxt/policy_utils.go
```diff
@@ -229,6 +284,15 @@ func setCustomizedPolicyTagsInSchema(d *schema.ResourceData, tags []model.Tag, s
 }
 
 func getPolicyTagsFromSchema(d *schema.ResourceData) []model.Tag {
+	if d == nil {
```
ksamoray, line 319: This change will mishandle `ignore_tags` case which is handled by `getCustomizedPolicyTagsFromSchema`.
Failure scenario: a resource has ignore_tags configured to preserve an NSX-discovered tag scope, and the user configures no tag block. The next Update PATCHes with an empty tag list, silently wiping the tag ignore_tags was meant to preserve. This isn't gated by NSXT_CACHE_MODE at all — it fires unconditionally, so it's a regression for every user of ignore_tags, not just cache users.

ksamoray:
All 10 touched data sources call the package-private gcache.readCache(...) directly instead of going through TryCacheRead/CacheAwareResourceRead.
gcache.readCache never consults postWriteByKey, so the one-shot "bypass cache right after a write" guarantee that CacheAwareResourceRead/TryCacheRead provide doesn't apply here. Failure scenario: create/update a resource and reference its id via a matching data source in the same apply — the data source can repopulate the bucket from NSX Search before the write is index-consistent, returning stale fields instead of falling back to a direct GET.

This ~15-line block is also copy-pasted identically across all 10 data source files — worth factoring into a shared helper so future schema changes don't need to be kept in sync by hand across 10 places.

nsxt/cache.go
```diff
@@ -0,0 +1,892 @@
...
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+)
+
+type resourceTypeCache struct {
+	mu       sync.RWMutex
+	data     map[string]map[string]*data.StructValue // key: query, key: resource ID
+	cacheHit int
```
ksamoray: cacheHit and cacheMiss aren't concurrency-safe. Consider atomic.Int64 type instead.

ksamoray:
Can we utilize `go test -race` to validate this PR?

nsxt/cache.go
poojav25: Good point — added a `cache_mode` provider attribute (same pattern as `context_id`/`NSXT_CONTEXT_ID`): it's Optional in the schema and defaults from `NSXT_CACHE_MODE` via `EnvDefaultFunc`, so it can now be set directly in the provider block, not just via env var. The resolved value is picked up once at Configure time.

nsxt/policy_utils.go
poojav25: You're right, this was a real bug. When there's no `tag` block in config, the code was returning an empty tag list outright, which would wipe out any tags being preserved via `ignore_tags`. Fixed it to return the ignored tags in that case instead of an empty list, so they survive the next update. Not pushed yet, will follow shortly.

nsxt/cache.go
poojav25: Agreed, switched cacheHit and cacheMis to atomic.Int64 and updated the increment/read sites in readCache to use Add()/Load() instead of ++.

nsxt/data_source_nsxt_policy_service.go
poojav25: Done — extracted a shared generic helper (cacheAwareDataSourceReadByID[T]) that handles the cache lookup, conversion, and sets id/display_name/description/path (mirroring policyDataSourceResourceFilterAndSet's core semantics). All ~10 data sources (Group, IpAddressPool, LBService, Service, Tier1Gateway, TransitGateway, Vpc, VpcConnectivityProfile, VpcServiceProfile, VpcSubnet) now call it instead of duplicating the block; each just sets its own extra fields (realized_id, is_default, etc.) on the returned object. Confirmed the display_name/id fallthrough and prefix/multi-match handling in the uncached path is untouched.

nsxt/policy_utils.go
poojav25: Done — added getPolicyTagsWithProviderManagedDefaultsValidated(d, m), the error-returning counterpart, which uses getValidatedTagsFromSchema instead of the silent getPolicyTagsFromSchema so the mode-check and validation both live in one place. Updated the 6 call sites that needed the error path (resource_nsxt_policy_transit_gateway.go x2, resource_nsxt_vpc_connectivity_profile.go x2, resource_nsxt_vpc_service_profile.go x2) to call it directly instead of duplicating the branch. Left the ~45 existing call sites on the plain getPolicyTagsWithProviderManagedDefaults alone since they don't need the error return and changing that signature would've meant touching all of them for no behavior change.

nsxt/data_source_nsxt_policy_group.go
poojav25: After digging deeper: this doesn't need a code change. The cache fast-path only triggers on ID-based lookups (`objID != ""`), never on display_name — display_name lookups always fall through to the uncached, domain-scoped policyDataSourceResourceRead path below, so the cross-domain display_name collision this comment describes can't actually reach the cache. And for ID lookups, NSX IDs are globally unique, so a shared (non-domain-scoped) cache bucket is already safe — the exact ID always resolves to the right object regardless of domain.

I also checked the cost of doing the fold anyway: the current bucket is populated by one bulk NSX Search per resource type across all domains. Scoping the query by domain would split that into a separate bulk search per domain, meaning a config referencing groups in N domains would trigger N Search calls instead of 1 - directly against the point of this cache - while fixing no live bug, since ID lookups are already domain-safe.

Given that, I'm leaving this as-is rather than folding domain into the key.

nsxt/provider.go
```diff
@@ -84,6 +86,17 @@ func Provider() *schema.Provider {
 				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("NSXT_ALLOW_UNVERIFIED_SSL", false),
 			},
+			"context_id": {
```
ksamoray, line 89: Provider arguments should be documented here:
https://github.com/vmware/terraform-provider-nsxt/blob/master/docs/index.md?plain=1#L162

nsxt/provider.go
```diff
@@ -1159,11 +1174,14 @@ func initCommonConfig(d *schema.ResourceData) commonProviderConfig {
...
 		Password:               password,
 		LicenseKeys:            licenses,
+		contextID:              contextId,
+		CacheMode:              cacheMode,
 	}
 }
 
 func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	commonConfig := initCommonConfig(d)
+	setConfiguredCacheMode(commonConfig.CacheMode)
```
ksamoray: Let's store the cache mode in the same way that we store the rest of the config params.

Claude has a long explanation about why this is better:

setConfiguredCacheMode is only called once per providerConfigure() invocation, not on every read. The problem isn't when it's set, it's where it's stored: it lands in a single package-level global, rather than a field on nsxtClients/commonProviderConfig scoped to that specific provider instance.

That distinction matters because Terraform's plugin SDK calls providerConfigure() once per aliased provider block, not once per process.

ksamoray (commented):
Also see below:

This is the same issue from my earlier pass and it's still present in the current head. I re-verified programmatically against every file that reads through CacheAwareResourceRead/TryCacheRead: 33 of 40 cache-integrated resource files have no MarkPostWrite*/InvalidateCache* call anywhere in their Delete function — resource_nsxt_policy_ip_pool.go, lb_*.go (6 files), tier1_gateway.go/tier1_gateway_interface.go, transit_gateway.go/transit_gateway_attachment.go, service.go, static_route.go, group.go, ip_address_allocation.go/project_ip_address_allocation.go, connectivity_policy.go, distributed_flood_protection_profile_binding.go, segment_common.go (nsxtPolicySegmentDelete), and effectively every resource_nsxt_vpc_*.go file including resource_nsxt_vpc_gateway_policy.go (which shares the composite GatewayPolicy bucket with resource_nsxt_policy_gateway_policy.go, which does invalidate correctly — so the same bucket is inconsistently maintained depending on which resource deleted from it).

Only 7 files get this right: gateway_policy.go, gateway_policy_rule.go, security_policy_rule.go, nat_rule.go, ip_pool_block_subnet.go, ip_pool_static_subnet.go, segment_port.go.

Concrete failure scenario: terraform destroy (or a targeted terraform apply removing one resource) on e.g. an nsxt_policy_ip_pool succeeds against NSX, but the type's search-query cache bucket still has the deleted object cached. Any subsequent read of that type in the same provider lifetime — another IP pool's refresh, or a data "nsxt_policy_ip_pool" lookup — can serve the deleted object back as if it still exists, until an unrelated Create/Update on that type happens to invalidate the bucket. Given this is the exact "does caching actually behave correctly" concern this PR needs to nail, I'd treat this as a blocker.

nsxt/provider.go
poojav25: Added — `context_id` and `cache_mode` are now documented in the Argument Reference section of docs/index.md.

nsxt/provider.go
poojav25: Fixed. Removed the package-level `configuredCacheMode` global and `setConfiguredCacheMode` entirely. Cache-mode checks now take `m interface{}` and read `m.(nsxtClients).CommonConfig.CacheMode` directly, the same per-instance pattern already used for `context_id`. Threaded `m` through `IsCacheEnabled`, `isGlobalSearchCacheMode`, `isConfigScopedCacheMode`, `isCacheEnabledForRead`, `MarkPostWriteForResourceTypeKey`, `MarkPostWriteAndInvalidateCacheForResourceType`, `InvalidateCacheForResourceType`, and `buildTagQuery`, and updated all call sites accordingly. Verified with gofmt/build/vet/tests and golangci-lint (0 issues).

poojav25:
Fixed the missing cache invalidation on Delete for cache-integrated resources.

`MarkPostWriteAndInvalidateCacheForResourceType(...)` is now called after a successful delete in all 32 affected Delete paths: `resource_nsxt_policy_ip_pool.go`, all LB profile/pool/virtual-server resources, `tier1_gateway.go`/`_interface.go`, `transit_gateway.go`/`_attachment.go`, `service.go`, `static_route.go`, `group.go`, `ip_address_allocation.go`/`project_ip_address_allocation.go`, `connectivity_policy.go`, `distributed_flood_protection_profile_binding.go`, `segment_common.go`, and every `resource_nsxt_vpc_*.go` file (including `vpc_gateway_policy.go`, so the shared `GatewayPolicy` bucket is now invalidated consistently regardless of which resource deletes from it).

Two special cases handled to match existing conventions rather than a naive 1-call-per-file patch:
- The 6 LB monitor-profile resources share one delete helper (`resourceNsxtPolicyLBMonitorProfileDelete` in `lb_utils.go`); only TCP and UDP are cache-integrated, so both buckets are invalidated there unconditionally (harmless no-op for the other 4 types).
- `nsxt_policy_dhcp_v4/v6_static_binding` share one generic delete function with no way to tell which type is being deleted from `d`, so both buckets are invalidated there — same reasoning already used elsewhere in this PR.
- Group: mirrored the existing Create/Update pattern where `resourceTypeGroup` invalidates inside the shared general-delete function and `resourceTypeVPCGroup` invalidates in the VPC-specific wrapper.

Verified: `gofmt`, `go build`, `go vet`, `golangci-lint run ./nsxt/...` (0 issues), and existing unit tests all pass.

nsxt/cache.go
```diff
@@ -0,0 +1,975 @@
...
+// ID, mirroring the common id/display_name/description/path semantics of
+// policyDataSourceResourceFilterAndSet. On success it sets those schema attributes and
+// returns the typed object so the caller can set any resource-specific fields. Returns
+// ok=false when the cache is disabled, objID is empty, or the object can't be
+// found/converted — callers should fall through to the regular (uncached) read path.
+func cacheAwareDataSourceReadByID[T any](d *schema.ResourceData, m interface{}, connector client.Connector, objID string, resourceType string, bindingType bindings.BindingType) (*T, bool) {
+	if objID == "" || !IsCacheEnabled(m) {
+		return nil, false
+	}
+	val, err := gcache.readCache(objID, resourceType, d, m, connector)
```
ksamoray, line 547: Shouldn't TryCacheRead or CacheAwareResourceRead be more correct here?
Seems like they call gcache.readCache but perform additional validation.

ksamoray:
Doc never mentions that config_scope mode writes tags etc - it's visible to users and they should be aware of this.
Maybe it's worthwhile to create a caching guide as we do with other stuff, under `docs/guides`?
We can also describe in such doc the global mode and what are the differences between these, and issue some recommendations for troubleshooting.

nsxt/cache.go
poojav25: Good catch — checked the two things `TryCacheRead`/`CacheAwareResourceRead` do that this didn't:

1. **`isRefreshPhase(d)` gate**: doesn't apply here — data sources always call this with a fresh `d` (`d.Id() == ""`), so requiring `isRefreshPhase` would disable data-source caching entirely. That gap is intentional, not a bug.
2. **Post-write one-shot bypass (`postWriteByKey`)**: this one was a real gap — `cacheAwareDataSourceReadByID` didn't check it at all, so a data source reading the same ID a resource just wrote to in the same run could get a cache hit instead of honoring the one-shot bypass the write set up.

Added the same `postWriteByKey.LoadAndDelete(...)` check here, mirroring `TryCacheRead`. Worth noting honestly: for data sources specifically this is a defense-in-depth/consistency fix more than a functional one today, since the non-cached fallback for an ID-based data source lookup (`policyDataSourceResourceRead` → `listPolicyResourcesByID`) is itself NSX-Search-based, not a direct GET — so bypassing the cache doesn't dodge Search's own indexing lag the way it does for resource Reads (which fall back to a direct `client.Get`). Still worth having so the three cache-entry-point functions behave consistently and stay correct if that fallback path ever changes.

`TryCacheRead`/`CacheAwareResourceRead` aren't drop-in replacements for this function's actual call sites — `CacheAwareResourceRead` requires a `backendRead`/`patchFunc` pair for write-time tag stamping that data sources don't have, and returns a different shape. Kept `cacheAwareDataSourceReadByID` as its own function, just aligned its cache-bypass behavior with the other two.

Verified with gofmt/build/vet/golangci-lint (0 issues) and existing tests.

salv-orlando (approved):
Discussed with Kobi and we are ok with check this PR in.

PR #1950, opened 2026-01-14
Update action to be computed-only for IPsecvpnRule

poojav25:


nsxt/resource_nsxt_policy_ipsec_vpn_service_test.go
```diff
@@ -351,14 +351,13 @@ resource "nsxt_policy_ipsec_vpn_service" "test" {
...
     destinations = ["%s"]
-    action       = "%s"
   }
 
   tag {
     scope = "scope1"
     tag   = "tag1"
   }
-}`, attrMap["display_name"], attrMap["description"], gatewayOrLocaleServicePath, attrMap["enabled"], attrMap["ha_sync"], attrMap["ike_log_level"], attrMap["sources"], attrMap["destinations"], attrMap["action"])
+}`, attrMap["display_name"], attrMap["description"], gatewayOrLocaleServicePath, attrMap["enabled"], attrMap["ha_sync"], attrMap["ike_log_level"], attrMap["sources"], attrMap["destinations"])
```
salv-orlando, line 360: Can you remove the action also from `accTestPolicyIPSecVpnServiceCreateAttributes` and `accTestPolicyIPSecVpnServiceUpdateAttributes` ?

docs/resources/policy_ipsec_vpn_service.md
```diff
@@ -44,7 +43,7 @@ The following arguments are supported:
 * `bypass_rule` - (Optional) Set the bypass rules for this IPSec VPN Service.
     * `sources` - (Required) List of source subnets. Subnet format is ipv4 or IPv6 CIDR.
     * `destinations` - (Required) List of destination subnets. Subnet format is ipv4 or IPv6 CIDR.
-    * `action` - (Required) `PROTECT` or `BYPASS`.
+    * `action` - (Computed) The effective rule action, always returned as BYPASS for service-level rules.
```
ksamoray, line 46: We mark `Computed` as optional. Since TF has no "readonly" attributes a user could assign a value here but it would be overwritten by TF.
We usually mark those as "Optional"

ksamoray: OK I've grepped the code - we do have other attributes which are marked as "Computed", so let's leave this as-is.

PR #1951, opened 2026-01-14
Use product version instead of node version to determine manager version

ksamoray:
To avoid upgrade failure


salv-orlando (approved):
LGTM

PR #1955, opened 2026-01-20
Fix data source schema to make display_name optional

poojav25:
enables fetching data sources by id only

### DIFF
--- a/nsxt/resource_nsxt_policy_ipsec_vpn_dpd_profile_test.go
+++ b/nsxt/resource_nsxt_policy_ipsec_vpn_dpd_profile_test.go
@@ -198,6 +198,5 @@ resource "nsxt_policy_ipsec_vpn_dpd_profile" "test" {
 }
 data "nsxt_policy_ipsec_vpn_dpd_profile" "test" {
   id = nsxt_policy_ipsec_vpn_dpd_profile.test.id
-  depends_on   = [nsxt_policy_ipsec_vpn_dpd_profile.test]
 }`, accTestPolicyIPSecVpnDpdProfileUpdateAttributes["display_name"])
 }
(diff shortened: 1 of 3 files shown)


nsxt/resource_nsxt_policy_ipsec_vpn_dpd_profile_test.go
```diff
@@ -194,5 +195,9 @@ func testAccNsxtPolicyIPSecVpnDpdProfileMinimalistic() string {
 	return fmt.Sprintf(`
 resource "nsxt_policy_ipsec_vpn_dpd_profile" "test" {
   display_name = "%s"
+}
+data "nsxt_policy_ipsec_vpn_dpd_profile" "test" {
+  id = nsxt_policy_ipsec_vpn_dpd_profile.test.id
+  depends_on   = [nsxt_policy_ipsec_vpn_dpd_profile.test]
```
ksamoray: I think that as line 200 creates an implicit dependency, the explicit dependency in line 201 is not needed (it would be if we would add an id property in the resource).

nsxt/resource_nsxt_policy_ipsec_vpn_ike_profile_test.go
```diff
@@ -201,5 +202,9 @@ resource "nsxt_policy_ipsec_vpn_ike_profile" "test" {
   display_name          = "%s"
   dh_groups             = ["%s"]
   encryption_algorithms = ["%s"]
+}
+data "nsxt_policy_ipsec_vpn_ike_profile" "test" {
+  id = nsxt_policy_ipsec_vpn_ike_profile.test.id
+  depends_on   = [nsxt_policy_ipsec_vpn_ike_profile.test]
```
ksamoray: Same here

nsxt/resource_nsxt_policy_ipsec_vpn_tunnel_profile_test.go
```diff
@@ -203,5 +204,9 @@ resource "nsxt_policy_ipsec_vpn_tunnel_profile" "test" {
   display_name          = "%s"
   encryption_algorithms = ["%s"]
   dh_groups             = ["%s"]
+}
+data "nsxt_policy_ipsec_vpn_tunnel_profile" "test" {
+  id = nsxt_policy_ipsec_vpn_tunnel_profile.test.id
+  depends_on   = [nsxt_policy_ipsec_vpn_tunnel_profile.test]
```
ksamoray: Same

ksamoray (approved):

PR #2032, opened 2026-04-13
Add duplicate value validation to tier0_gateway_paths attribute of the project

ksamoray:
Address a bug when duplicate values are added to the list of tier0_gateway_paths.

### DIFF
--- a/nsxt/resource_nsxt_policy_project.go
+++ b/nsxt/resource_nsxt_policy_project.go
@@ -34,6 +34,13 @@ func resourceNsxtPolicyProject() *schema.Resource {
 		},
 
 		CustomizeDiff: func(ctx context.Context, d *schema.ResourceDiff, m interface{}) error {
+			if v, ok := d.GetOk("tier0_gateway_paths"); ok {
+				if list, ok := v.([]interface{}); ok && len(list) > 0 {
+					if err := ValidateStringListNoDuplicateValues(list, "tier0_gateway_paths"); err != nil {
+						return err
+					}
+				}
+			}
 			if util.NsxVersionHigherOrEqual("9.1.0") {
 				c := m.(nsxtClients)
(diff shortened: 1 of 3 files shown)


aruntony005 (approved):
LGTM

docs/resources/policy_tier1_gateway.md
```diff
@@ -97,7 +97,7 @@ resource "nsxt_policy_tier1_gateway" "tier1_gw" {
   default_rule_logging      = "false"
   enable_firewall           = "true"
   enable_standby_relocation = "false"
-  tier0_path                = data.nsxt_policy_project.demoproj.tier0_gateway_paths.0
+  tier0_path                = one(data.nsxt_policy_project.demoproj.tier0_gateway_paths)
```
salv-orlando: What does "one" do?

salv-orlando (approved):
LGTM, but I'm curious about the change in docs

docs/resources/policy_tier1_gateway.md
ksamoray: one() will:
* Return the only set element if the set has just one
* Will return null if the set has no elements
* Raise an error if the set has multiple elements

Alternately, I can use here:
tolist(data.nsxt_policy_project.demoproj.tier0_gateway_paths)[0]

But then it'll fail if the set is empty (as it would before). I think that it's equally good the way it is (the project def is a data source, we do not know how many T0 it has).

salv-orlando: sounds good.
But in a way, isn't this a backward incompatible change?

Existing configurations may break. It's a remote corner case so we can document that if tier0_path is picked from a project data source / resource users will need to leverage this syntax.

salv-orlando (approved):
LGTM

PR #2033, opened 2026-04-15
Allow updates for node_settings attribute of edge_transport_node

### LINKED ISSUE #2029: node_settings - are not enforced/applied - nsxt_edge_transport_node
### Terraform

1.14.8

### Terraform Provider

3.11.0

### VMware NSX

9.0.1.0.24952111

### Description

Configuring DNS and Search Domain through Terraform provider is not working. Given configuration is NOT reflected in NSX and therefore "not applied" even though Terraform shows change is applied.

This is part of "nsxt_edge_transport_node" where is section of "node_settings" are two parameters:
dns_servers - (Optional) List of DNS servers.
search_domains - (Optional) List of Search domain names.

Upon their definition the configuration is accepted, deployed/applied but NOT reflected in NSX GUI.

It seems that every Terraform run (terraform apply) this settings dns_servers & search_domain are always applied again and again.

### Affected Resources or Data Sources

Edge Nodes configuration of their DNS and Search domain.
Edge nodes were completely build by Terrafom as well.

### Terraform Configuration

**Setup/Usage:**
  node_settings {
    hostname             = "hostname.domain.xyz"
    allow_ssh_root_login = true
    enable_ssh           = true
    **dns_servers          = ["10.1.2.3"]
    search_domains       = ["domain.xyz"]**
  }

### Debug Output

No debug.

### Panic Output

N/A

### Expected Behavior

Defined values for DNS servers and search domain is pushed/applied properly to defined Edge Nodes.

### Actual Behavior

DNS configuration and Search domain settings is NOT applied on the Edge nodes.

### Steps to Reproduce

Simply adding given parameters into Edge nodes definition in Terraform.

### Environment Details

_No response_

### Screenshots

Image

### References

_No response_

ksamoray:
Attributes which are nested under node settings weren't updatable.

Fixes: #2029

### DIFF
--- a/nsxt/resource_nsxt_edge_transport_node.go
+++ b/nsxt/resource_nsxt_edge_transport_node.go
@@ -882,18 +882,16 @@ func getTransportNodeFromSchema(d *schema.ResourceData, m interface{}) (*mpmodel
 		return nil, fmt.Errorf("failed to create Transport Node: %v", err)
 	}
 
+	nodeSettings, err := getEdgeNodeSettingsFromSchema(d.Get("node_settings"))
+	if err != nil {
+		return nil, err
+	}
+
 	var nodeDeploymentInfo *data.StructValue
+	converter := bindings.NewTypeConverter()
 	if nodeID == "" {
-		/*
-			node_id attribute conflicts with node_deployment_info. As node_deployment_info is a complex object which has
-			attributes with default values, this schema property will always have values - therefore there is no simple way
-			to enforce this conflict within the provider (e.g check if node_id and node_deployment_info have values, then
-			fail.
-			So the provider will ignore node_deployment_info properties when node_id has a value - which would mean that
-			this edge appliance was created externally.
-		*/
+		// Full EdgeNode: VM deployment plus node_settings. node_id conflicts with deployment_config / fqdn / etc. in schema.
 		log.Printf("node_id not specified, will deploy edge using values in deploymentConfig")
-		converter := bindings.NewTypeConverter()
 		var dataValue data.DataValue
 		var errs []error
 
@@ -902,10 +900,6 @@ func getTransportNodeFromSchema(d *schema.ResourceData, m interface{}) (*mpmodel
 		ipAddresses := interfaceListToStringList(d.Get("ip_addresses").([]interface{}))
 
 		deploymentConfig, err := getEdgeNodeDeploymentConfigFromSchema(d.Get("deployment_config"))
-		if err != nil {
-			return nil, err
-		}
-		nodeSettings, err := getEdgeNodeSettingsFromSchema(d.Get("node_settings"))
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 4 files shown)


nsxt/resource_nsxt_edge_transport_node.go
```diff
@@ -938,6 +938,42 @@ func getTransportNodeFromSchema(d *schema.ResourceData, m interface{}) (*mpmodel
 	return &obj, nil
 }
 
+// applyNodeSettingsFromSchemaToTransportNodeUpdate merges node_settings from Terraform into the
+// existing EdgeNode when node_id is set. In that case getTransportNodeFromSchema omits
+// NodeDeploymentInfo (see deployment vs. register semantics); updates would otherwise never
+// send NodeSettings (e.g. dns_servers, search_domains). When getTransportNodeFromSchema
+// already supplied NodeDeploymentInfo (node_id empty), it is left unchanged. If
+// getEdgeNodeSettingsFromSchema returns nil (no node_settings in resource data), the
+// transport node update payload is unchanged for NodeDeploymentInfo.
```
salv-orlando, line 947: do we need this comment?

nsxt/resource_nsxt_edge_transport_node.go
```diff
@@ -938,6 +938,42 @@ func getTransportNodeFromSchema(d *schema.ResourceData, m interface{}) (*mpmodel
...
+func applyNodeSettingsFromSchemaToTransportNodeUpdate(d *schema.ResourceData, existing *mpmodel.TransportNode, obj *mpmodel.TransportNode) error {
+	if obj.NodeDeploymentInfo != nil {
+		return nil
+	}
```
salv-orlando, line 951: Is this check need? Not sure why the update should be skipped if the object we're building for sending the PUT request has NodeDeploymentInfo.

nsxt/resource_nsxt_policy_edge_transport_node.go
```diff
@@ -345,7 +345,7 @@ func resourceNsxtPolicyEdgeTransportNode() *schema.Resource {
 				Optional:      true,
 				Computed:      true,
 				Description:   "Unique Id of the fabric node",
-				ConflictsWith: []string{"advanced_configuration", "appliance_config", "credentials", "form_factor", "management_interface", "vm_deployment_config"},
```
salv-orlando, line 348: Are we sure about this change?
There is a comment below that "appliance_config does not conflict with node_id", but we also initially took this from NSX OpenAPI spec

nsxt/resource_nsxt_edge_transport_node.go
ksamoray: This if statement is meant to distinguish the case of an edge which has NodeDeploymentInfo in the backend, from a one that doesn't. I'm not sure that this is required TBH, I retained this to avoid a case where the predeployed edge is breaking due to having VM deployment config or such set by the user.
Maybe this can be done in a better way.

nsxt/resource_nsxt_policy_edge_transport_node.go
ksamoray: That can't be part of the API spec as node_id is not not a property that exists in the policy edge spec:
We chose to add it to retain the functionality that we offer for the MP edge, where an edge can be configured using "create" operation. The list here includes the properties which aren't valid for this use case and `appliance_config` isn't one of these as the issue raises (for MP, but the same applies here).

salv-orlando (approved):
LGTM

PR #2040, opened 2026-04-21
Address a bug with floodProtectionProfileBindingModelToSchema

ksamoray:
When null values are returned from NSX, function causes a null pointer exception.


salv-orlando (approved):
LGTM

PR #2047, opened 2026-04-27
Fix edge_transport_node issues

ksamoray:
Two different issues found in edge_transport_node:
* When a user specifies a policy path for the profiles or TZs, the value is accepted. Yet NSX returns the edge struct with IDs instead of the sent values. This PR replaces the submitted values with IDs.
* NSX returns audit_username when not specified. This should be computed.


salv-orlando (approved):
LGTM

PR #2048, opened 2026-04-27
Improve acceptance tests coverage for IDPS tests

ksamoray:
Create tier1 gateway instead of relying on a pre-created gateway.


aruntony005 (approved):
LGTM

PR #2051, opened 2026-04-28
Create self signed certificate to imporve acceptance test coverage

ksamoray:
Instead of dependency on a pre-existing certificate which is referenced by the acceptance test, create a certificate within the test scope.


salv-orlando (approved):
LGTM

PR #2052, opened 2026-04-28
Remove external dependency for LB service

ksamoray:
Create a loadbalancer service within the test context instead of having an external dependency.


salv-orlando (approved):
LGTM

PR #2058, opened 2026-05-04
Implement VpcAdnsServer resources

ksamoray:
Modified resources:
nsxt_vpc_service_profile

New resources:
nsxt_policy_dns_service
nsxt_policy_dns_zone
nsxt_policy_dns_record
nsxt_policy_dns_rule
nsxt_policy_dns_record_auto_config


nsxt/resource_nsxt_policy_dns_record.go
```diff
@@ -0,0 +1,258 @@
...
+				Description: "DNS record data values consistent with record_type.",
+			},
+			"zone_path": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validatePolicyPath(),
+				Description:  "Policy path to the ProjectDnsZone. Immutable after creation.",
+			},
+			"ttl": {
```
yuanyouyao, line 78: Do we need a validation function here,

ValidateFunc: validation.IntBetween(30, 86400),

The same with nsxt_policy_dns_zone, and nsxt_policy_dns_record_auto_config

ksamoray: We often delegate validations to NSX. Not sure whether this case worths making an exception.

yuanyouyao: Make sense.

nsxt/resource_nsxt_policy_dns_zone.go
```diff
@@ -0,0 +1,350 @@
...
+				},
+				Description: "Optional list of VPC policy paths for split-horizon DNS. When set, only workloads in the specified VPCs can resolve this zone. When empty, all VPCs in the project can resolve it.",
+			},
+			"ttl": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Default:     300,
+				Description: "Default Time-To-Live in seconds for DNS records in this zone (30-86400).",
+			},
+			"soa": {
```
salv-orlando, line 65: Potential permadiff:

spec explicitly states: "All fields are optional; omitted fields use system defaults" with documented defaults (expire_time: 604800, refresh_interval: 3600, retry_interval: 900, negative_cache_ttl: 300).

Only serial_number is marked Computed: true in the schema — the other four numeric fields and primary_nameserver/responsible_party are not.

If a user sets the soa block with just one field populated (e.g., primary_nameserver), the rest resolve to their "zero" value, but NSX returns its real defaults on Read — causing a diff on every subsequent plan 

Documentation (docs/resources/policy_dns_zone.md:53) and acceptance test (resource_nsxt_policy_dns_zone_test.go:230-237) hard-code every sub-field to the exact system-default values, meaning that the acceptance tests might not catch the permadiff because of this.

nsxt/resource_nsxt_policy_dns_service.go
```diff
@@ -0,0 +1,268 @@
...
+				ValidateFunc: validatePolicyPath(),
+				Description:  "Policy path to the transit gateway providing north-south connectivity.",
+			},
+			"forwarder_config": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"cache_size": {
```
salv-orlando, line 64: Permadiff warning

Spec confirms cache_size defaults to 5000 server-side.

It's Optional only, no Computed, no Default. If a user sets forwarder_config with just upstream_servers, cache_size will drift after the first apply. 

Notably, resource_nsxt_policy_dns_service_test.go never exercises forwarder_config at all — this whole block is untested, which is likely why the gap wasn't caught.

nsxt/resource_nsxt_policy_dns_rule.go
```diff
@@ -0,0 +1,273 @@
...
+				ValidateFunc: validation.StringInSlice(dnsRuleActionTypeValues, false),
+				Description:  "Action to apply to DNS queries matching domain_patterns. UPDATE_MEMBERSHIP or FORWARD.",
+			},
+			"domain_patterns": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				Description: "Domain name patterns matched via longest-prefix match. Supports wildcards (e.g. '*.example.com'). Required unless action_type is FORWARD and shared_zone_path is set.",
```
salv-orlando, line 59: It looks like we have not implemented this validation in the provider.
Is that intentional, ie: we will let NSX do the validation and return the error, or did you want to implement it?

nsxt/resource_nsxt_policy_dns_zone.go
```diff
@@ -0,0 +1,350 @@
...
+					Type:         schema.TypeString,
+					ValidateFunc: validatePolicyPath(),
+				},
+				Description: "Optional list of VPC policy paths for split-horizon DNS. When set, only workloads in the specified VPCs can resolve this zone. When empty, all VPCs in the project can resolve it.",
+			},
+			"ttl": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Default:     300,
+				Description: "Default Time-To-Live in seconds for DNS records in this zone (30-86400).",
```
salv-orlando: I believe we tend to implement in the provider this kind of static validations.
Said that, I don't have a strong opinion.

nsxt/resource_nsxt_policy_dns_record.go
salv-orlando: As commented elsewhere, when I seem to recall that when the constraint are exclusively on the specific attribute value, we do the validation in the provider.

But I don't have a strong opinion, and also it's maybe best to let NSX do the validation in case the allowed TTL range changes in future releases.

nsxt/resource_nsxt_policy_dns_zone.go
```diff
@@ -0,0 +1,350 @@
...
+			soaMap["expire_time"] = int(*obj.Soa.ExpireTime)
+		}
+		if obj.Soa.NegativeCacheTtl != nil {
+			soaMap["negative_cache_ttl"] = int(*obj.Soa.NegativeCacheTtl)
+		}
+		d.Set("soa", []interface{}{soaMap})
+	}
+}
+
+func dnsZoneParentsFromPath(parentPath string) (org, project, dnsServiceID string, err error) {
```
salv-orlando: code duplication:

dnsZoneParentsFromPath (resource_nsxt_policy_dns_zone.go:226-232) and dnsRuleParentsFromPath (resource_nsxt_policy_dns_rule.go:124-130) are byte-for-byte identical logic (parseStandardPolicyPathVerifySize(parentPath, 3, ...) returning org, project, dnsServiceID), differing only in the local path-example variable name — which itself is the same string in both files ("/orgs/[org]/projects/[project]/dns-services/[dns-service]"). 

These can be collapsed into a single shared helper (e.g. dnsServiceParentsFromPath).

nsxt/resource_nsxt_policy_dns_zone.go
```diff
@@ -0,0 +1,350 @@
...
+		}
+		if v, ok := soaMap["expire_time"].(int); ok && v > 0 {
+			et := int64(v)
+			soa.ExpireTime = &et
+		}
+		if v, ok := soaMap["negative_cache_ttl"].(int); ok && v >= 0 {
+			nc := int64(v)
+			soa.NegativeCacheTtl = &nc
+		}
+		if !reflect.DeepEqual(soa, &model.DnsZoneSoa{}) {
```
salv-orlando, line 190: This ensures we don't send an empty SOA struct, which is somewhat related to the comment above related to permadiffs. 

A one-line comment explaining why (avoid overriding server-computed defaults with an empty struct) would help the next reader avoid re-introducing the permadiff while "fixing" it)

nsxt/resource_nsxt_policy_dns_rule.go
ksamoray: We usually leave these validations to NSX, as they tend to change the validations occasionally. I don't mind adding it though.

nsxt/resource_nsxt_policy_dns_zone.go
ksamoray: Same here.

docs/resources/policy_dns_service.md
```diff
@@ -0,0 +1,80 @@
...
+resource "nsxt_policy_dns_service" "example" {
+  context {
+    project_id = data.nsxt_policy_project.demoproj.id
+  }
+
+  display_name = "my-dns-service"
+  description  = "Terraform provisioned DNS Service"
+
+  allocated_listener_ips = [
+    nsxt_project_ip_address_allocation.listener.path,
```
yuanyouyao: The schema requires its a ip address, is it OK to use path?

ksamoray: Good catch, seems like the API spec has been revised.

aruntony005:
LGTM

aruntony005 (approved):

PR #2060, opened 2026-05-05
Fix VM multitenancy and other VM testing issues

ksamoray:
Some test issues are due to usage of the wrong VM id in multitenancy environments.
Also, tests were looking up segment ids which weren't exposed.

For `nsxt_policy_vm_tags`, there are two issues which this handles which are beyond the testing scope:
* The code uses `path` to import a multitenancy VM tags resource. Yet, this resource has no path attribute. This has been replaced with instance_id::project_id import id.
* There's no coverage for the case of VPC VM. This has been added as well, with instance_id::project_id::vpc_id import id scheme.

### DIFF
--- a/nsxt/resource_nsxt_policy_vm_tags.go
+++ b/nsxt/resource_nsxt_policy_vm_tags.go
@@ -43,11 +43,19 @@ func resourceNsxtPolicyVMTagsImporter(d *schema.ResourceData, m interface{}) ([]
 		return nil, ErrEmptyImportID
 	}
 
-	if idx := strings.Index(importID, vmTagsImportIDSeparator); idx >= 0 {
-		instanceID := importID[:idx]
-		projectID := importID[idx+len(vmTagsImportIDSeparator):]
+	parts := strings.SplitN(importID, vmTagsImportIDSeparator, 3)
+	switch len(parts) {
+	case 1:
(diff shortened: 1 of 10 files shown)


yuanyouyao (commented):
LGTM

nsxt/resource_nsxt_policy_vm_tags.go
```diff
@@ -35,7 +37,50 @@ var cliVirtualMachinesClient = realized_enforcement_points.NewVirtualMachinesCli
...
+	importID := d.Id()
+	if isSpaceString(importID) {
+		return nil, ErrEmptyImportID
+	}
+
+	parts := strings.SplitN(importID, vmTagsImportIDSeparator, 3)
+	switch len(parts) {
+	case 1:
+		// Plain instance ID — local manager (no project, no VPC).
+		if !isValidResourceID(parts[0]) {
```
aruntony005, line 50: Should we add the same validation for the case 2 & 3 below?

aruntony005 (approved):
A small validation check.
Otherwise LGTM

nsxt/resource_nsxt_policy_vm_tags.go
ksamoray: Good idea, I wonder whether we should do this elsewhere to harden the validations in the provider.

PR #2061, opened 2026-05-05
Implement nsxt_policy_virtual_network_appliance_cluster

ksamoray:
resource and datasource

These are required for testing of other resources and might be needed sooner or later.
Add also nsxt_policy_virtual_network_appliance_cluster_realization datasource.


nsxt/resource_nsxt_policy_virtual_network_appliance_cluster.go
```diff
@@ -0,0 +1,424 @@
...
+				},
+			},
+		},
+	}
+}
+
+func resourceNsxtPolicyVirtualNetworkApplianceClusterRead(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+	sessionContext := getSessionContext(d, m)
+	client := cliVNAClustersClient(sessionContext, connector)
```
yuanyouyao, line 161: check the client?

This may apply to other occurrences in the file too.

nsxt/resource_nsxt_policy_virtual_network_appliance_cluster.go
```diff
@@ -0,0 +1,424 @@
...
+				Description: "ID of the enforcement point this Virtual Network Appliance Cluster belongs to",
+				Optional:    true,
+				ForceNew:    true,
+				Default:     "default",
+			},
+			"appliance_form_factor": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Description:  "Form factor for virtual network appliances in this cluster",
+				ValidateFunc: validation.StringInSlice(vnaClusterFormFactorValues, false),
```
yuanyouyao, line 76: Do we need set 
Computed:    true,  ?

I'm not sure if this will create drift.

nsxt/resource_nsxt_policy_virtual_network_appliance_cluster.go
ksamoray: Client validation pretty much across the entire repo. IDK if this would be relevant specifically here - if we consider this, let's do it everywhere.

nsxt/resource_nsxt_policy_virtual_network_appliance_cluster.go
ksamoray: I'm not sure either. We use `Computed` only when we find it to be required in testing, as it could cause issues (Terraform doesn't handle this very well).

aruntony005 (approved):

PR #2062, opened 2026-05-05
Implement VpcDnsForwarder resources

ksamoray:
Modified resources:
nsxt_vpc_service_profile

Also revised tests as required

### DIFF
--- a/nsxt/resource_nsxt_vpc_service_profile_test.go
+++ b/nsxt/resource_nsxt_vpc_service_profile_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 )
 
 var accTestPolicyVpcServiceProfileCreateAttributes = map[string]string{
@@ -29,6 +30,8 @@ var accTestPolicyVpcServiceProfileUpdateAttributes = map[string]string{
 	"dns_server_ips":   "7.7.7.2",
 	"server_addresses": "11.11.11.111",
 }
+
+var accTestPolicyVpcServiceProfileDnsForwarderHelper = getAccTestResourceName()
 
 func TestAccResourceNsxtVpcServiceProfile_basic(t *testing.T) {
 	testResourceName := "nsxt_vpc_service_profile.test"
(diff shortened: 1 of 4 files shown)


nsxt/resource_nsxt_vpc_service_profile_test.go
```diff
@@ -533,3 +583,72 @@ resource "nsxt_vpc_service_profile" "test" {
...
+resource "nsxt_vpc_service_profile" "test" {
+  %s
+  display_name = "%s"
+
+  dhcp_config {
+  }
+
+  dns_forwarder_config {
+    default_forwarder_zone_path = nsxt_policy_dns_forwarder_zone.test.path%s
+    log_level                   = "INFO"
```
poojav25: Magic strings: "INFO" and cache_size = 1024 here are raw literals with no compile-time link to `model.PolicyVpcDnsForwarder_LOG_LEVEL_INFO`. If that enum value ever changes upstream, this HCL string won't get a compiler error — only a failing acceptance run. Worth a short comment noting it must match `model.PolicyVpcDnsForwarder_LOG_LEVEL_*`.

nsxt/resource_nsxt_vpc_subnet_test.go
```diff
@@ -603,6 +603,60 @@ func TestAccResourceNsxtVpcSubnet920_subnetDhcpv6Config(t *testing.T) {
...
+			testAccOnlyVPC(t)
+			testAccNSXVersion(t, "9.2.0")
+		},
+		Providers: testAccProviders,
+		CheckDestroy: func(state *terraform.State) error {
+			return testAccNsxtVpcSubnetCheckDestroy(state, accTestVpcSubnet920DnsPrefUpdateAttrs["display_name"])
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: testAccNsxtVpcSubnetDnsServerPreferenceTemplate("DNS_FORWARDER_PREFERRED_OVER_PROFILE_DNS_SERVERS"),
```
poojav25: Magic strings: "DNS_FORWARDER_PREFERRED_OVER_PROFILE_DNS_SERVERS" / "PROFILE_DNS_SERVERS_PREFERRED_OVER_DNS_FORWARDER" are raw literals here with no compile-time link to `model.SubnetDhcpConfig_DNS_SERVER_PREFERENCE_*` (the mock tests use the real constants for this). If the enum value ever changes upstream, this string won't get a compiler error — only a failing acceptance run. Worth a short comment noting it must match `model.SubnetDhcpConfig_DNS_SERVER_PREFERENCE_*`.

nsxt/resource_nsxt_vpc_service_profile_test.go
```diff
@@ -533,3 +583,72 @@ resource "nsxt_vpc_service_profile" "test" {
 		attrMap["v6_lease"], attrMap["v6_pref"], attrMap["v6_ntp"], attrMap["v6_sntp"], attrMap["v6_dns"],
 		attrMap["fwd_log"], attrMap["svc_cidr"])
 }
+
+func testAccNsxtVpcServiceProfileWithDnsForwarder(withConditional bool) string {
```
poojav25: Duplicated HCL template: `testAccNsxtVpcServiceProfileWithDnsForwarder` and `testAccNsxtVpcServiceProfileWithDnsForwarderNoConfig` (below) repeat the exact same two `nsxt_policy_dns_forwarder_zone` resource blocks verbatim. The only real difference is whether `dns_forwarder_config` is present. If someone tweaks the zone resource later (e.g. adds a field), they'll have to remember to update it in both places.

Suggestion: factor the shared zone blocks into a helper (e.g. `dnsForwarderZoneResources() string`) and interpolate it into both templates, or better, build the whole config with a small `dnsForwarderConfigBlock string` parameter that's empty for the "no config" case, collapsing these two functions into one.

nsxt/resource_nsxt_vpc_service_profile_test.go
ksamoray: Done. But if NSX ever changes these values, the user impact will be much more severe than the impact on the provider itself...

nsxt/resource_nsxt_vpc_subnet_test.go
ksamoray: Same here

poojav25:
LGTM

salv-orlando (approved):
LGTM

PR #2071, opened 2026-05-08
@ANS-TF-IDPS: support gateway tier0/tier1 security config enablement

bgujar:
+ Datasource for gateway tier0/tier1 security config enablement
+ resource for the same
+ documentation
+ tests

Jira: #NSEC-13547


ksamoray:
If gfw_multicast_enabled, geoip_monitoring_enabled are always false, never passed to NSX, why do we include them in the schema?
Anyway if a user sets a true value to these, he'll have a permadiff which is bad. Maybe remove these?

nsxt/resource_nsxt_policy_gateway_security_config.go
```diff
@@ -0,0 +1,390 @@
...
+}
+
+// gatewaySecurityConfigImporter parses import IDs in the format "tier0/<id>" or "tier1/<id>".
+func gatewaySecurityConfigImporter(_ context.Context, d *schema.ResourceData, _ interface{}) ([]*schema.ResourceData, error) {
+	importID := d.Id()
+	parts := strings.SplitN(importID, "/", 2)
+	if len(parts) != 2 || parts[1] == "" {
+		return nil, fmt.Errorf("invalid import ID %q: expected format 'tier0/<gateway-id>' or 'tier1/<gateway-id>'", importID)
+	}
+	gwType := strings.ToLower(parts[0])
```
ksamoray: Do we ever expect this to be non-lowercase?

bgujar: ideally not. but do not want to break if it does.

bgujar:
> If gfw_multicast_enabled, geoip_monitoring_enabled are always false, never passed to NSX, why do we include them in the schema? Anyway if a user sets a true value to these, he'll have a permadiff which is bad. Maybe remove these?

Sure. will remove them. They will be required to be supported in future, hence were added. But Makes more sense to remove.

nsxt/resource_nsxt_policy_gateway_security_config.go
ksamoray: Let's remain consistent with the rest of the provider. If a user chooses to modify the policy path to uppercase, or invalidate it in any other way, we should return an error. Unless invalid inputs cause a crash or such.

bgujar: sure will remove the conversion

ksamoray (approved):

PR #2074, opened 2026-05-10
Implement nsxt_policy_route_controller resource and datasource

ksamoray:
Add resource and data source for the Route Controller API
(/infra/route-controllers/{id}) with an inline bgp_config block backed
by a single atomic H-API call.

Key implementation details:
- H-API (Infra.Patch) creates/updates RouteController and its
  RouteControllerBgpRoutingConfig child atomically in one request,
  mirroring the Tier-0 / transit-gateway BGP pattern.
- RouteControllerBgpRoutingConfig.Id="bgp" and ResourceType are set
  explicitly so NSX can route the child correctly; omitting them caused
  the "Resource type not provided for id:null" error seen previously.
- MarkedForDelete child is sent when the user removes bgp_config, so
  cleanup is also atomic.
- ha_mode and virtual_network_appliance_cluster_path use OmitIfEmpty
  metadata, letting SchemaToStruct skip nil-pointer serialisation that
  NSX rejected with a field-validation error.
- Acceptance tests depend on nsxt_policy_virtual_network_appliance_cluster
  (from the VNA branch base) rather than a hardcoded env-var path.


nsxt/resource_nsxt_policy_virtual_network_appliance_cluster.go
```diff
@@ -0,0 +1,424 @@
...
+				},
+			},
+		},
+	}
+}
+
+func resourceNsxtPolicyVirtualNetworkApplianceClusterRead(d *schema.ResourceData, m interface{}) error {
+	connector := getPolicyConnector(m)
+	sessionContext := getSessionContext(d, m)
+	client := cliVNAClustersClient(sessionContext, connector)
```
aruntony005, line 143: Can you add nil check here

nsxt/resource_nsxt_policy_virtual_network_appliance_cluster.go
```diff
@@ -0,0 +1,424 @@
...
+
+func resourceNsxtPolicyVirtualNetworkApplianceClusterExists(siteID, epID, id string, connector client.Connector) (bool, error) {
+	sessionContext := utl.SessionContext{ClientType: utl.Local}
+	siteClient := cliSitesClient(sessionContext, connector)
+	_, err := siteClient.Get(siteID)
+	if err != nil {
+		msg := fmt.Sprintf("failed to read site %s", siteID)
+		return false, logAPIError(msg, err)
+	}
+	vnaClient := cliVNAClustersClient(sessionContext, connector)
```
aruntony005, line 226: can you add nil check to avoid provider crash

nsxt/resource_nsxt_policy_virtual_network_appliance_cluster.go
```diff
@@ -0,0 +1,424 @@
...
+	}
+	if v, ok := data["overlay_transport_zone_path"].(string); ok && v != "" {
+		cfg.OverlayTransportZonePath = &v
+	}
+	return cfg
+}
+
+func resourceNsxtPolicyVirtualNetworkApplianceClusterExists(siteID, epID, id string, connector client.Connector) (bool, error) {
+	sessionContext := utl.SessionContext{ClientType: utl.Local}
+	siteClient := cliSitesClient(sessionContext, connector)
```
aruntony005, line 217: same here

nsxt/resource_nsxt_policy_route_controller.go
```diff
@@ -0,0 +1,461 @@
...
+	bgpObj := model.RouteControllerBgpRoutingConfig{
+		Ecmp:                  &ecmp,
+		MultipathRelax:        &multipathRelax,
+		GracefulRestartConfig: &restartConfigStruct,
+	}
+
+	if localAsNum, ok := cfgMap["local_as_num"].(string); ok && localAsNum != "" {
+		bgpObj.LocalAsNum = &localAsNum
+	}
+	if timer, ok := cfgMap["peer_route_convergence_timer"].(int); ok && timer > 0 {
```
poojav25, line 213: The schema allows 0 via IntAtLeast(0) but this guard silently drops it. Either change the guard to ok (without > 0), or tighten the validator to IntAtLeast(1) to match

nsxt/resource_nsxt_policy_route_controller_test.go
```diff
@@ -0,0 +1,293 @@
...
+					resource.TestCheckResourceAttrSet(testResourceName, "revision"),
+					resource.TestCheckResourceAttr(testResourceName, "tag.#", "0"),
+					resource.TestCheckResourceAttrSet(testDataSourceName, "path"),
+				),
+			},
+		},
+	})
+}
+
+func TestAccResourceNsxtPolicyRouteController_withBgp(t *testing.T) {
```
poojav25, line 110: No test step covers removing bgp_config — the MarkedForDelete path in resourceNsxtPolicyRouteControllerUpdate is untested

nsxt/resource_nsxt_policy_virtual_network_appliance_cluster.go
poojav25, line 140: resourceNsxtPolicyVirtualNetworkApplianceClusterDelete function has hardcoded session context , would it possible to change it to getSessionContext(d, m)

poojav25:
LGTM

aruntony005:
Please take a look at the pipeline failure.
Otherwise LGTM

aruntony005 (approved):

poojav25 (approved):

PR #2083, opened 2026-05-14
Fix three bugs in nsxt_edge_transport_node_rtep resource (bug 3698901)

ksamoray:
Issue 1 - Import by transport node UUID was broken: Read, Update and Delete all used d.Get("edge_id") to look up the edge, which is empty during terraform import (only d.Id() is populated at that point). Switch all three to d.Id(); add d.Set("edge_id", id) in Read so the attribute is written back to state after an import.

Issue 2 - Update sent an empty string pointer for NamedTeamingPolicy whenever the attribute was not set, causing NSX to reject the request with "uplink teaming policy name/s [] not specified". Apply the same non-empty guard that Create already had.

Issue 3 - DHCP and no_ipv4 IP assignment types are not supported by the NSX RTEP API (a DHCP request triggers a server-side NPE, code 99). Add getIPAssignmentSchemaForRTEP() which strips those unsupported fields from the shared schema, so an unsupported assignment type is rejected at terraform plan time instead of reaching the API. Update docs and unit tests accordingly.

### DIFF
--- a/docs/resources/edge_transport_node_rtep.md
+++ b/docs/resources/edge_transport_node_rtep.md
@@ -21,7 +21,7 @@ resource "nsxt_edge_transport_node_rtep" "test_rtep" {
 
   host_switch_name = "someSwitch"
   ip_assignment {
-    assigned_by_dhcp = true
+    static_ip_pool = "ip-pool-uuid"
   }
   named_teaming_policy = "tp123"
   rtep_vlan            = 500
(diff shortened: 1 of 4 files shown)


docs/resources/edge_transport_node_rtep.md
```diff
@@ -35,7 +35,6 @@ The following arguments are supported:
 * `edge_id` - (Required) Edge ID to associate with remote tunnel endpoint.
 * `host_switch_name` - (Required) The host switch name to be used for the remote tunnel endpoint.
 * `ip_assignment` - (Required) - Specification for IPs to be used with host switch virtual tunnel endpoints. Should contain exactly one of the below:
-    * `assigned_by_dhcp` - (Optional) Enables DHCP assignment.
```
aruntony005, line 38: This field is still used [here](https://github.com/ksamoray/terraform-provider-nsxt/blob/fix-edge-rtep/nsxt/resource_nsxt_edge_transport_node.go#L2000) and same with [no_ipv4](https://github.com/ksamoray/terraform-provider-nsxt/blob/fix-edge-rtep/nsxt/resource_nsxt_edge_transport_node.go#L2003) as well

ksamoray: @aruntony005 we reuse this func in various places and each of these allow different types of allocations (e.g for edge management interface, DHCP is usable). So the parse func is always the same, we should make sure that each schema allows only the valid kinds for the specific case.

aruntony005 (approved):

PR #2087, opened 2026-05-17
Use latest dev SDK

ksamoray:
Use the SDK build 25431216


ksamoray:
Added to earlier PR, no need for this one.

PR #2094, opened 2026-05-20
Fix data source policy_project populating ipv6_blocks when API omits …

ksamoray:
…field

When NSX version >= 9.2.0 but a project has no IPv6 blocks configured, the API omits the ipv6_blocks field entirely. The previous code called d.Set("ipv6_blocks", nil) unconditionally, causing Terraform to write ipv6_blocks=[] into the tfstate. This broke the FVT test test_data_policy_project, which compared the tfstate against the direct API response and found a mismatch.

Fix: only set ipv6_blocks in state when obj.Ipv6Blocks is non-nil, i.e. when the API actually returned the field.


yuanyouyao (approved):
LGTM

salv-orlando (approved):
LGTM

PR #2095, opened 2026-05-20
Fix TransitGateway routing/BGP config apply and add BGP acceptance test

ksamoray:
The NSX API rejects TransitGatewayRoutingConfig as an H-API child of TransitGateway (error 500165 "Invalid Dto type hierarchy"), so routing config is now applied via a direct API PATCH after TGW creation and update, while BgpRoutingConfig (which IS supported as an H-API child) continues to be included in the OrgRoot H-API call.

The /routing endpoint also returns an empty body until the TGW routing engine initializes on the edge (eventual consistency), so redistribution config changes are applied optimistically and the resource uses ExpectNonEmptyPlan in acceptance tests to acknowledge this behavior.

Adds TestAccResourceNsxtPolicyTransitGateway_withBgpConfig covering centralized-config creation, bgp_config + redistribution_config update, and bgp param updates across 3 Terraform apply steps.


salv-orlando (approved):
LGTM

PR #2097, opened 2026-05-24
Fix TestAccResourceNsxtVpcSubnetDhcpV4StaticBindingConfig920_basic te…

ksamoray:
…st failure

NSX 9.2.0 enforces (error code 503293) that the gateway_address in a DHCP v4 static binding must match the subnet's actual gateway (first host of the subnet CIDR). The shared update-attribute map used gateway_address 192.168.240.2 which is rejected on 9.2.0+ for subnet 192.168.240.0/26 whose gateway is 192.168.240.1.

Add accTestVpcSubnetDhcpV4StaticBindingConfig920UpdateAttributes with gateway_address 192.168.240.1 and a matching
testAccNsxtVpcSubnetDhcpV4StaticBindingConfig920Template helper. Update TestAccResourceNsxtVpcSubnetDhcpV4StaticBindingConfig920_basic to use these so the update step stays valid on NSX 9.2.0 while the pre-9.2.0 test remains untouched.

### DIFF
--- a/nsxt/resource_nsxt_vpc_dhcp_v4_static_binding_config_test.go
+++ b/nsxt/resource_nsxt_vpc_dhcp_v4_static_binding_config_test.go
@@ -44,6 +44,24 @@ var accTestVpcSubnetDhcpV4StaticBindingConfigUpdateAttributes = map[string]strin
 	"opt-other-value": "45",
 }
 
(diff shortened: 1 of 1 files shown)


salv-orlando (approved):
LGTM

PR #2101, opened 2026-05-26
Prevent index panic in policy path importer

yuanyouyao:
The policy resource importer helper in policy_utils.go relied on strict hardcoded slice indices (e.g. index 6) when splitting user-provided import ID paths.

When a user attempts to import an 'nsxt_policy_static_route' with a truncated or invalid ID path format, the provider crashes with an out of bounds index panic: "runtime error: index out of range [6] with length 6".


salv-orlando (approved):
LGTM

PR #2105, opened 2026-05-27
Fix drift in dhcpv6_relay_config.server_addresses (bug 3709234)

ksamoray:
The NSX API may return DHCPv6 relay server addresses in a different order than they were written. Because server_addresses was a TypeList, Terraform treated the reordering as a change, causing perpetual drift on every subsequent plan.

Change server_addresses inside dhcpv6_relay_config to TypeSet so that order differences from the API are ignored. Also add an acceptance test that configures two relay addresses and asserts the second plan is a no-op.


yuanyouyao (approved):
LGTM

PR #2106, opened 2026-05-27
Fix nsxt_policy_project data source ipv6_blocks version guard (bug 3705663)

ksamoray:
NSX 9.1.x returns "ipv6_blocks": [] (explicit empty array) in the project API response, but the previous version guard NsxVersionHigherOrEqual("9.2.0") prevented the provider from reading/setting that value for NSX < 9.2.0. This left the Terraform state with ipv6_blocks=null while the API returned [], causing the FVT test verify_data to fail with:
  "Value None for attribute ipv6_blocks not equal to API"

Fix: remove the version guard from the data source read entirely. The API response is the source of truth — always reflect whatever the API returns for ipv6_blocks, regardless of NSX version. The write-path guard in resourceNsxtPolicyProjectPatch (9.2.0+) is unchanged.


salv-orlando (approved):
LGTM

PR #2108, opened 2026-05-28
Fix nsxt_edge_transport_node node_settings perpetual drift (bug 3689817)

ksamoray:
Fix nsxt_edge_transport_node node_settings perpetual drift (bug 3689817)

Root cause (empirically verified): when a freshly-deployed edge VM boots,
its management plane agent (MPA) performs an initial sync that reports the
VM's current running state to NSX.  NSX writes those MPA-reported values
back into node_deployment_info.node_settings, temporarily overwriting the
desired state that Terraform had just PUT.  Reading those stale values in
the subsequent resourceNsxtEdgeTransportNodeRead causes the same changes to
re-appear on every plan (perpetual drift).

Empirical evidence: for an already-initialized edge (nsxedge.1) a PUT is
reflected in GET immediately; the overwrite is specific to the new-edge MPA
initial-boot sync window.  Changes are not lost — once the window closes,
NSX re-delivers the desired state to the edge MPA which applies it.

Fix: resourceNsxtEdgeTransportNodeRead intentionally skips refreshing
node_settings from the API.  The Terraform config is treated as the
authoritative source for these fields, eliminating the drift.

Dead code removed:
  - setEdgeNodeSettingsInSchema (only caller was the now-removed block)
  - setKeyValueListForSchema in utils.go (only caller was the above)
  - TestUnitNsxt_setKeyValueListForSchema unit test

Mock test updates:
  - edgeNodeStructValue() no longer includes NodeSettings (irrelevant
    to Read now); comment explains the MPA sync root cause.
  - edgeNodeStructValueWithNodeSettings() helper added.
  - New test case "Read does not overwrite node_settings from API"
    verifies that an MPA-overwritten stale hostname does not replace
    the config-driven value in state after Read.

Acceptance test added (TestAccResourceNsxtEdgeTransportNodeNodeSettings):
  - Step 1: deploy edge TN with minimal node_settings (hostname only).
  - Step 2: update node_settings to all supported attributes after a
    60-second PreConfig delay (simulates the MPA boot-sync window).
  - Step 3: plan-only after another 60-second delay; ExpectNonEmptyPlan=false
    confirms zero drift.

ImportStateVerifyIgnore updated for both import tests to exclude
"node_settings", since Read no longer populates these fields from the API.

### DIFF
--- a/nsxt/policy_utils_and_utils_unit_test.go
+++ b/nsxt/policy_utils_and_utils_unit_test.go
@@ -247,11 +247,3 @@ func TestUnitNsxt_getKeyValuePairListFromSchema(t *testing.T) {
 	require.Len(t, out, 1)
 	assert.Equal(t, "k", *out[0].Key)
 }
-
-func TestUnitNsxt_setKeyValueListForSchema(t *testing.T) {
-	ka2, vb2 := "a", "b"
-	out := setKeyValueListForSchema([]mp_model.KeyValuePair{
-		{Key: &ka2, Value: &vb2},
-	}).([]interface{})
-	require.Len(t, out, 1)
-}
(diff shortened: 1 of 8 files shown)


salv-orlando:
Expected NSX API behaviour, it does not require fix in Terraform provider

PR #2112, opened 2026-06-04
Fix nsx_id client type and TGW routing child parent_path bugs

ksamoray:
Root cause (client type)
-------------------------
getOrGenerateIDWithParent passed getSessionContext(d, m) to the
presence-checker, which derives the client type from the schema context
block. Resources that carry context in parent_path (such as
nsxt_policy_transit_gateway_prefix_list) have no context block, so
clientType resolved to Local and the Multitenancy-only client factory
returned nil, producing "unsupported client type". The bug was only
triggered when nsx_id was explicitly set (the existence check is
skipped when nsx_id is absent).

Fix (nsxt/policy_utils.go)
--------------------------
Pass getParentContext(d, m, parentPath) instead of getSessionContext,
so the correct Multitenancy context is derived from parent_path.
Affects all 16 resources that call getOrGenerateIDWithParent.

Root cause (parent_path on import)
-----------------------------------
NSX nests some Transit Gateway children (prefix lists, route maps,
community lists, BFD peers) under an intermediate "routing" singleton
that has no corresponding schema attribute, so the real API path is
.../transit-gateways/[tgw]/routing/prefix-lists/[id]. The generic
nsxtParentPathResourceImporter truncated only the last two path
segments, leaving a spurious "/routing" suffix in parent_path that
didn't match the Transit Gateway path users configure, failing
ImportStateVerify.

Fix (nsxt/policy_utils.go)
--------------------------
Added nsxtTGWRoutingChildResourceImporter, which wraps the generic
importer and strips the "/routing" suffix, and switched
nsxt_policy_transit_gateway_prefix_list/route_map/community_list/
bfd_peer to use it. This keeps the special case visible at each
resource's Importer declaration rather than hidden inside the shared
helper.

Tests
-----
Unit test (utgomock_resource_nsxt_policy_transit_gateway_prefix_list
_test.go): TestMockResourceNsxtPolicyTransitGatewayPrefixListCreate
NsxIDProjectContext installs a context-sensitive client factory that
returns nil for non-Multitenancy contexts, mirroring NewPrefixListsClient.
Without the fix the test fails with "unsupported client type".

Unit test (policy_utils_test.go):
TestUnitNsxt_nsxtTGWRoutingChildResourceImporter verifies the
"/routing" suffix is stripped from parent_path on import.

Acceptance tests (resource_nsxt_policy_transit_gateway_prefix_list
_test.go): three tests exercising basic CRUD, nsx_id creation (direct
regression for BZ#3713681), and import. Prerequisites use an explicit
nsxt_policy_transit_gateway resource with centralized_config (matching
the bug-report setup) rather than a default TGW data source, which
ensures the prefix-list API is available. CheckDestroy correctly uses
err != nil to catch surviving resources.

Added the same basic/import acceptance test coverage - previously
missing entirely - for the other three TGW routing children:
resource_nsxt_policy_transit_gateway_community_list_test.go,
resource_nsxt_policy_transit_gateway_route_map_test.go, and
resource_nsxt_policy_transit_gateway_bfd_peer_test.go. Verified against
a live NSX manager: prefix_list, community_list, and route_map all
pass end-to-end (basic + import). bfd_peer's config applies correctly
and exercises the same fix, but GET on
.../routing/bfd-peers/[id] returns an empty body on that manager - a
server-side gap unrelated to this fix - so its test is included but
unverified in that environment.

### DIFF
--- a/nsxt/resource_nsxt_policy_transit_gateway_prefix_list_test.go
+++ b/nsxt/resource_nsxt_policy_transit_gateway_prefix_list_test.go
@@ -1,0 +1,284 @@
+// © Broadcom. All Rights Reserved.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MPL-2.0
+
+package nsxt
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+)
+
+var accTestPolicyTransitGatewayPrefixListCreateAttributes = map[string]string{
+	"display_name": getAccTestResourceName(),
+	"description":  "terraform created",
+	"network":      "10.0.0.0/8",
+	"action":       "PERMIT",
+}
+
+var accTestPolicyTransitGatewayPrefixListUpdateAttributes = map[string]string{
+	"display_name": getAccTestResourceName(),
+	"description":  "terraform updated",
+	"network":      "172.16.0.0/12",
+	"action":       "DENY",
+}
+
+func TestAccResourceNsxtPolicyTransitGatewayPrefixList_basic(t *testing.T) {
+	testResourceName := "nsxt_policy_transit_gateway_prefix_list.test"
+	prereqName := getAccTestResourceName()
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccOnlyLocalManager(t)
+			testAccNSXVersion(t, "9.2.0")
+		},
+		Providers: testAccProviders,
+		CheckDestroy: func(state *terraform.State) error {
+			return testAccNsxtPolicyTransitGatewayPrefixListCheckDestroy(state, accTestPolicyTransitGatewayPrefixListUpdateAttributes["display_name"])
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: testAccNsxtPolicyTransitGatewayPrefixListTemplate(true, prereqName),
+				Check: resource.ComposeTestCheckFunc(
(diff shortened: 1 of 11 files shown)


nsxt/resource_nsxt_policy_transit_gateway_prefix_list_test.go
```diff
@@ -0,0 +1,293 @@
...
+	connector := getPolicyConnector(testAccProvider.Meta().(nsxtClients))
+	for _, rs := range state.RootModule().Resources {
+		if rs.Type != "nsxt_policy_transit_gateway_prefix_list" {
+			continue
+		}
+		resourceID := rs.Primary.Attributes["id"]
+		parentPath := rs.Primary.Attributes["parent_path"]
+		sessionContext := getSessionContextFromParentPath(testAccProvider.Meta(), parentPath)
+		exists, err := resourceNsxtPolicyTransitGatewayPrefixListExists(sessionContext, parentPath, resourceID, connector)
+		if err == nil {
```
yuanyouyao: If the existence check completes successfully (which means err is nil), the function immediately returns nil (success). It completely skips the if exists check on line 194. As a result, if a test resource fails to delete and still exists on the NSX manager, the acceptance test will incorrectly report that the resource has been successfully destroyed, allowing a major failure to pass silently!

Hence line#191 should be if err != nil {

yuanyouyao (changes requested):

nsxt/resource_nsxt_policy_transit_gateway_prefix_list_test.go
ksamoray: Seems like this bug also exists elsewhere. I'll fix them all in the scope of this change

yuanyouyao (approved):
Please fix the checks.

Others LGTM.

yuanyouyao (approved):
LGTM

PR #2113, opened 2026-06-07
Fix credential drift in nsxt_policy_virtual_network_appliance (bug 37…

ksamoray:
…15433)

When an Update adds the credentials block for the first time, Terraform places the planned passwords in the diff area of ResourceData (accessible via d.Get) but not yet in the writer. The previous setVNACredentialsInSchema returned nil early when obj.Credentials was nil, never calling d.Set, so the passwords were never transferred to the writer and were lost from state, causing perpetual + credentials drift on every subsequent plan.

Fix: always call d.Set("credentials", …) when credentials are present in d.Get("credentials"), regardless of whether obj.Credentials is nil. This ensures passwords are written to the ResourceData writer and persisted to state.

Add three new mock sub-tests that cover:
- Read preserves passwords when API returns nil credentials
- Read preserves passwords and populates computed usernames when API returns a Credentials object
- Read does not inject a credentials block when none is configured

Add TestAccResourceNsxtPolicyVirtualNetworkAppliance_credentialsDrift, an acceptance test that deploys a VNA without credentials then updates it to add credentials, verifying the plan is a no-op after the update (the scenario described in bug update #1).

### DIFF
--- a/nsxt/resource_nsxt_policy_virtual_network_appliance.go
+++ b/nsxt/resource_nsxt_policy_virtual_network_appliance.go
@@ -208,22 +208,28 @@ func getVNACredentialsFromSchema(creds interface{}) *model.VirtualNetworkApplian
 }
 
 func setVNACredentialsInSchema(d *schema.ResourceData, obj *model.VirtualNetworkApplianceCredential) error {
-	if obj == nil {
-		return nil
-	}
-	// Only update computed usernames if the credentials block was configured.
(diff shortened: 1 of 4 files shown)


salv-orlando (commented):
I now understand how this fixes the issue.
The check at line 211 was simply completely preventing updating the schema with the credentials block if obj was nil.
LGTM

salv-orlando (approved):

PR #2117, opened 2026-06-08
Fix nsxt_policy_project ipv6_blocks version guard in read (bug 3705663)

ksamoray:
Both NSX 9.1.x and 9.2.x return "ipv6_blocks": [] (an explicit empty array) in the project API response, even when no IPv6 blocks are configured. The version guard NsxVersionHigherOrEqual("9.2.0") therefore prevented the data source and resource reads from setting ipv6_blocks in state on NSX 9.1.x, leaving it as null while the API returned []. The FVT verify_data test then failed with:
  "Value None for attribute ipv6_blocks not equal to API"

Fix: remove the version guard from ipv6_blocks in both dataSourceNsxtPolicyProjectRead and resourceNsxtPolicyProjectRead so the state always reflects the API response, regardless of NSX version. The write-path guard in resourceNsxtPolicyProjectPatch (9.2.0+) is intentionally kept.

### DIFF
--- a/nsxt/data_source_nsxt_policy_project.go
+++ b/nsxt/data_source_nsxt_policy_project.go
@@ -10,8 +10,6 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
-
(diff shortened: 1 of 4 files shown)


salv-orlando (approved):
LGTM

PR #2121, opened 2026-06-09
Documentation issues with policy_transit_gateway zone_based_spans

ksamoray:
use_all_zones has no implementation or API behind it. zone_external_ids is required as it's the only attribute in the zone_based_span block.


salv-orlando (approved):
LGTM

PR #2122, opened 2026-06-09
Fix state drift for dhcp_server_addresses when DHCP is deactivated

salv-orlando:
Suppress plan diffs for VPC subnet 'dhcp_server_addresses' when IPv4 DHCP or IPv6 DHCP is deactivated, as the NSX API does not persist these addresses in such states.

Filter out empty or deactivated DHCP server addresses from the subnet update payload to prevent API format validation errors.

Add unit tests for the suppress function and expand acceptance tests to verify automatic state population when DHCP is transitioned from deactivated to active.


ksamoray (approved):
LGTM. isDhcpv4Deactivated, isDhcpv6Deactivated is almost same and can be consolidated but that's really a nit.

PR #2125, opened 2026-06-11
Add withIdempotencyChecks test helper; apply to edge transport node

ksamoray:
Add a reusable withIdempotencyChecks() helper in utils_test.go that wraps a []resource.TestStep slice and inserts a PlanOnly step after each non-import, non-plan-only step, verifying that every apply leaves no residual diff.

Apply the helper to both acceptance tests in
resource_nsxt_edge_transport_node_test.go.


salv-orlando (approved):
LGTM

PR #2129, opened 2026-06-15
Gate slow/redundant acc tests behind NSXT_TEST_EXTRA_COVERAGE

ksamoray:
Add testAccNsxtExtraCoverage helper to utils_test.go with documentation describing when to apply it, then gate 8 tests that are either expensive to run (ETN basic+import at ~800s each), stable IPv6 variants (T0/T1 interface _withV6/_withIPv6), or redundant IDPS data source variants (~163-423s). Tests remain available in periodic CI runs where NSXT_TEST_EXTRA_COVERAGE=true is set.


aruntony005 (approved):
LGTM

PR #2130, opened 2026-06-15
Fix VNA credential drift after terraform import (bug 3715433)

ksamoray:
After 'terraform import' the NSX API does not return passwords, leaving the credentials block absent from state and causing every subsequent 'terraform plan' to report +credentials drift.

Two provider changes fix this:

1. Importer seeds an empty credentials block before Read is called. setVNACredentialsInSchema already skips its early return when a block is present, so it then writes NSX-returned usernames into state.

2. Add suppressIfEmptyPriorState DiffSuppressFunc on cli_password and root_password. When old=="" (state has no password after import) and d.Id()!="" (resource already exists), the diff is suppressed so the subsequent plan shows "No changes." The d.Id()!="" guard ensures the suppression does not fire during Create or in unit tests.

Testing:

- Add unit tests covering the importer-seeded path and the DiffSuppressFunc behaviour (TestSuppressIfEmptyPriorState, Read_with_importer_seeded_block_writes_usernames_to_state).
- Add acceptance test TestAccResourceNsxtPolicyVirtualNetworkAppliance_ importWithCredentials that reproduces the exact FVT failure: create with credentials → import → assert plan is empty.
- Add withImportIdempotencyChecks helper to utils_test.go (companion to withIdempotencyChecks) that auto-inserts a PlanOnly step after each ImportState step; reusable for any resource with write-only fields.

### DIFF
--- a/nsxt/resource_nsxt_policy_virtual_network_appliance.go
+++ b/nsxt/resource_nsxt_policy_virtual_network_appliance.go
@@ -75,10 +75,11 @@ func resourceNsxtPolicyVirtualNetworkAppliance() *schema.Resource {
 							Description: "Audit username (computed)",
 						},
 						"cli_password": {
-							Type:        schema.TypeString,
-							Required:    true,
-							Sensitive:   true,
-							Description: "Node CLI admin password",
+							Type:             schema.TypeString,
+							Required:         true,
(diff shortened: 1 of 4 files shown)


salv-orlando (approved):
LGTM

PR #2131, opened 2026-06-15
nsxt_policy_project: guard ipv6_blocks d.Set with nil check (bug 3705663)

ksamoray:
Some NSX 9.2.0 builds omit ipv6_blocks from the project API response entirely when no IPv6 blocks are configured. Unconditionally calling d.Set("ipv6_blocks", nil) serialises the TypeList as [] in state, which causes the FVT verify_data check to fail with "Attribute not found in API output".

Guard the d.Set call in both dataSourceNsxtPolicyProjectRead and resourceNsxtPolicyProjectRead so ipv6_blocks is only written to state when the API actually returned the field (obj.Ipv6Blocks != nil). Add unit sub-tests covering the omitted-field case for both read paths.


salv-orlando (approved):
LGTM

PR #2133, opened 2026-06-19
Fix parallel test race in IDPS signature version acceptance tests

ksamoray:
Both TestAccResourceNsxtPolicyIdpsSignatureVersion_basic and _makeActive wrote to the same singleton NSX object (nsx_id="DEFAULT") while running as ParallelTest, causing one test to clobber the other. Switch both to resource.Test (sequential) following the same pattern used by the IDPS settings acceptance tests.


salv-orlando (approved):
LGTM

PR #2134, opened 2026-06-21
Fix VNA import panic on nil credentials list element

ksamoray:
The Terraform Plugin SDK v2 normalises an empty TypeList element set by the importer to nil on the subsequent d.Get call.  setVNACredentialsInSchema then panicked with an unsafe type-assertion on c[0].  Switch to a nil-safe assertion and add a unit test that exercises this path directly.


salv-orlando (approved):
LGTM

PR #2136, opened 2026-06-22
Fix parallel test race in DistributedVlanConnection acceptance tests

ksamoray:
_basic and _importBasic both resolved display_name from the static package-level accTestPolicyDistributedVlanConnectionCreateAttributes/ UpdateAttributes maps. Running in parallel, both created an NSX object with the same display_name; the data source lookup then found two matches and failed with "Found multiple DistributedVlanConnection".

Give _basic its own createName/updateName local variables and thread them through testAccNsxtPolicyDistributedVlanConnectionTemplate. Give _importBasic its own updateName (the existing `name` local) and thread it through testAccNsxtPolicyDistributedVlanConnectionMinimalistic. Each test now operates on uniquely-named resources with no shared names.

### DIFF
--- a/nsxt/resource_nsxt_policy_distributed_vlan_connection_test.go
+++ b/nsxt/resource_nsxt_policy_distributed_vlan_connection_test.go
@@ -29,6 +29,8 @@ var accTestPolicyDistributedVlanConnectionUpdateAttributes = map[string]string{
 func TestAccResourceNsxtPolicyDistributedVlanConnection_basic(t *testing.T) {
 	testResourceName := "nsxt_policy_distributed_vlan_connection.test"
 	testDataSourceName := "data.nsxt_policy_distributed_vlan_connection.test"
+	createName := getAccTestResourceName()
+	updateName := getAccTestResourceName()
 
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck: func() {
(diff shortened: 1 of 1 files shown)


nsxt/resource_nsxt_policy_distributed_vlan_connection_test.go
yuanyouyao, line 189: It checks if err == nil { return err }. If the existence check succeeds without an API error (err == nil), the function returns nil (success) immediately. It completely bypasses the if exists check on line 191.
If a resource fails to destroy and still exists on the NSX manager, the test suite will still report that the destroy check passed!

yuanyouyao (changes requested):

nsxt/resource_nsxt_policy_distributed_vlan_connection_test.go
ksamoray: This bug exists in multiple tests - but is outside of the scope of this PR. I'll address all of these in a separate PR.

ksamoray: See #2141

yuanyouyao: sounds good.

yuanyouyao (approved):
LGTM

PR #2137, opened 2026-06-22
Fix parallel test races in acceptance tests

ksamoray:
Generate unique names per test invocation instead of using package-level variables for resource display_names (LBService, ProjectIpAddressAllocation, StaticRouteBfdPeer, VpcIpAddressAllocation) and shared prerequisite infra names (TransitGateway attachment/service/endpoint/static-route/nat-rule).

### DIFF
--- a/nsxt/resource_nsxt_policy_lb_service_test.go
+++ b/nsxt/resource_nsxt_policy_lb_service_test.go
@@ -32,24 +32,25 @@ var accTestPolicyLBServiceUpdateAttributes = map[string]string{
 	"size":              "SMALL",
 }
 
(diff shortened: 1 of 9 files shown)


nsxt/resource_nsxt_policy_lb_service_test.go
salv-orlando: Maybe not important, but should these names be generated on a test case basis now that we're running them in parallel?

ksamoray: They could, for sure this will be cleaner if the test is referencing these later by name

salv-orlando (approved):
LGTM

PR #2139, opened 2026-06-23
Fix panic in getSpanFromSchema on nil span list element

ksamoray:
The Terraform SDK populates optional blocks that contain only empty/ default attributes (e.g. zone_based_span { zone_external_ids = [] }) with a single nil element rather than an empty slice. The previous code performed unchecked type assertions on that element, triggering a panic at apply time.

Replace the bare assertions with nil/length guards at every level of the nested structure. Use the existing getElemOrEmptyMapFromMap helper for the cluster_based_span and zone_based_span inner lookups so that nil elements are silently skipped via the helper's comma-ok path.


salv-orlando (approved):
LGTM

PR #2140, opened 2026-06-23
Guard against nil list elements in schema-to-struct helpers

ksamoray:
Five locations shared the same latent panic: a len > 0 (or != nil) guard on a []interface{} attribute was sufficient to detect an absent block, but not a block present with all-default values that the Terraform SDK represents as a list containing a single nil element. A direct [0].(map[string]interface{}) assertion on that nil element triggers a panic at apply/read time.

Fixes:
- resource_nsxt_compute_manager.go: getCredentialData – replace the two-part guard with getElemOrEmptyMapFromMap, which already uses comma-ok on the element assertion.
- resource_nsxt_policy_gateway_route_map.go: policyGatewayRouteMapBuildEntry set block – same replacement.
- resource_nsxt_edge_transport_node.go: setEdgeDeploymentConfigInSchema node_user_settings – collapse the nil-guarded if/else into a single getElemOrEmptyMapFromMap call.
- segment_common.go: getOldProfileDataForRemoval – add comma-ok and nil-element guards (raw interface{} param, no map+key helper applies).
- metadata/metadata.go: polyNestedSchemaToStruct – replace the bare v.([]interface{}) cast and [0].(map[string]interface{}) chain with a vList intermediate, guarded by isList, len, and nil checks.

### DIFF
--- a/nsxt/metadata/metadata.go
+++ b/nsxt/metadata/metadata.go
@@ -661,7 +661,12 @@ func polyNestedSchemaToStruct(ctx string, elem reflect.Value, dataList []interfa
 	for i, dataElem := range dataList {
 		dataMap := dataElem.(map[string]interface{})
 		for k, v := range dataMap {
-			if len(v.([]interface{})) == 0 {
+			vList, isList := v.([]interface{})
+			if !isList || len(vList) == 0 || vList[0] == nil {
+				continue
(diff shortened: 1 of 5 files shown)


salv-orlando (approved):
LGTM

PR #2141, opened 2026-06-23
Fix inverted error check in CheckDestroy test functions

ksamoray:
All affected CheckDestroy functions used `if err == nil { return err }` after calling a resourceExists helper. Since err is nil in that branch, this always returned success and never reached the `if exists` guard, meaning a resource surviving destroy would go undetected.

Change the condition to `if err != nil { return err }` in all 67 affected acceptance test files.


yuanyouyao (approved):
LGTM

salv-orlando:
A pedant nit for a follow up PR.
In files like nsxt/resource_nsxt_policy_lb_service_test.go, checkDestroy works correctly:

```
func testAccNsxtPolicyLBServiceCheckDestroy(state *terraform.State, displayName string) error {
	connector := getPolicyConnector(testAccProvider.Meta().(nsxtClients))
	sessionContext := testAccGetSessionContext()
	nsxClient := infra.NewLbServicesClient(sessionContext, connector)

		if rs.Type != "nsxt_policy_lb_service" {
			continue
		}

		_, err := nsxClient.Get(resourceID)
		if err == nil {
			return fmt.Errorf("Policy LBService %s still exists", displayName)
		}
	}
	return nil
}
```

But a more appropriate verification would require checking err before returning nil, e.g.:

```
    _, err := nsxClient.Get(resourceID)
    if err == nil {
        return fmt.Errorf("Policy LBService %s still exists", displayName)
    }
    if !isNotFoundError(err) {
        return err  // surface unexpected API errors
    }
```

PR #2144, opened 2026-06-28
nsxt_policy_route_controller: mark virtual_network_appliance_cluster_path as Required

ksamoray:
The NSX API enforces virtual_network_appliance_cluster_path as a required field on RouteController objects (error code 255: required property missing), but the Terraform schema declared it Optional. This caused apply to fail at the API layer with no prior Terraform-level validation.

- Change schema attribute from Optional to Required
- Remove OmitIfEmpty from metadata (not applicable to required fields)
- Update docs and example usage to reflect the requirement


salv-orlando (approved):
LGTM

PR #2146, opened 2026-06-28
nsxt_policy_route_controller_interface: mark interface_address as Required

ksamoray:
The NSX API enforces interface_address as a required field on RouteControllerInterface objects (error code 255: required property missing), but the Terraform schema declared it Optional. This caused apply to fail at the API layer with no prior Terraform-level validation.

- Change interface_address from Optional to Required with MinItems: 1
- Update docs to reflect the requirement


salv-orlando (approved):
LGTM

PR #2147, opened 2026-06-28
fix: address route controller interface acc test failures on LM 9.1

ksamoray:
Two distinct backend changes broke 7 acceptance tests:

1. NSX now enforces (code 640171) that interface_address subnets must belong to the same network as floating_ip_subnets. Fix the BGP neighbor test helper: align both to 192.168.200.0/24, consistent with the source_addresses used in the BGP neighbor configs.

2. NSX always returns mtu=1500 (default) in read responses. The mtu schema field was Optional-only, causing a perpetual diff (1500->null) after every apply. Add Computed:true so Terraform accepts the API-returned default when the user has not explicitly set the field.

### DIFF
--- a/nsxt/resource_nsxt_node_user_test.go
+++ b/nsxt/resource_nsxt_node_user_test.go
@@ -17,19 +17,17 @@ import (
 )
 
 var accTestNodeUserCreateAttributes = map[string]string{ //nolint:gosec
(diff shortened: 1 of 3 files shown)


salv-orlando (approved):
LGTM

PR #2153, opened 2026-06-30
fix(nsxt_policy_route_controller_bgp_neighbor): mark source_addresses as Required

ksamoray:
The NSX API enforces source_addresses as a required field on RouteControllerBgpNeighbor objects (error code 255: required property missing), but the Terraform schema declared it Optional. This caused updates to fail at the API layer with no prior Terraform-level validation.

- Change source_addresses from Optional to Required with MinItems: 1
- Update docs to reflect the requirement


salv-orlando (approved):
LGTM

PR #2154, opened 2026-06-30
fix(nsxt_policy_virtual_network_appliance_cluster): mark appliance_form_factor as Computed

ksamoray:
The NSX API always returns appliance_form_factor in read responses (defaulting to "MEDIUM" when not explicitly set). The schema field was Optional-only, so Terraform perpetually detected drift (MEDIUM -> null) when the field was omitted from a minimal config.

Add Computed: true, consistent with the other server-defaulted fields in this resource (appliance_type, service_type, password_managed_by_vcf). Update docs to reflect (Optional, Computed).


salv-orlando (approved):
LGTM

PR #2156, opened 2026-07-02
fix(nsxt_policy_transit_gateway): mark span.zone_based_span as Computed

ksamoray:
span.cluster_based_span and span.zone_based_span are mutually exclusive nested blocks (governed by ExactlyOneOf), but only cluster_based_span was declared Computed. This asymmetry caused Terraform to fail to reconcile the server-returned zone_based_span value with state, showing a perpetual "add zone_based_span" diff on every re-apply.

Add Computed: true to zone_based_span to match its sibling. Update docs to reflect (Optional, Computed).


salv-orlando (approved):
LGTM

PR #2159, opened 2026-07-06
Force delete policy sites on destruction via safe_to_force_delete

yuanyouyao:
When destroying an NSX-T policy site, the standard delete operation can be blocked or fail if there are lingering dependencies or active connections. This leaves the system in an inconsistent or partially deleted state.

Address this by adding a boolean attribute 'safe_to_force_delete' (default: false) to the resource. When set to true, a forceDelete flag is passed to the API client's Delete call. This ensures the policy site is forcefully and successfully removed during Terraform's resource destruction phase.


salv-orlando (approved):
LGTM

ksamoray (approved):
LGTM

PR #2160, opened 2026-07-06
fix(nsxt_policy_route_controller_bgp_neighbor): suppress password drift after import

ksamoray:
NSX never returns password on GET, so terraform import leaves it empty in state, causing a spurious diff/re-apply on every plan. Reuse the suppressIfEmptyPriorState DiffSuppressFunc already used by nsxt_policy_virtual_network_appliance for the same write-only-field import issue.


salv-orlando (approved):
LGTM

PR #2161, opened 2026-07-06
fix(nsxt_policy_transit_gateway): fix zone_based_span empty-list drop

ksamoray:
getSpanFromSchema treated an explicitly empty zone_external_ids list as an absent zone_based_span block, because the SDKv2 diff engine emits no attribute entries for a nested list's contents when it stays at zero elements, leaving the reconstructed element as a bare nil. Since neither span variant registered as present, no span was sent on create, so NSX fell back to its own default ClusterBasedSpan and every subsequent plan showed a perpetual span drift.

Determine block presence from the list length (which the diff engine does populate correctly) instead of the reconstructed element map, and tolerate a nil element by defaulting to an empty zone list.


salv-orlando (approved):
LGTM

PR #2162, opened 2026-07-06
fix(nsxt_policy_transit_gateway): wire up advanced/redistribution config

ksamoray:
advanced_config.forwarding_up_timer and redistribution_config.rule were declared in the schema but never read into the API request, and Read unconditionally reset both to nil on every refresh regardless of what NSX returned. Both map onto the ForwardingUpTimer and RouteRedistributionConfig fields of the same TransitGatewayBgpRoutingConfig struct already used for bgp_config (same /routing/bgp API), not a separate object as a stale comment and the docs assumed.

Merge them into getTGWBgpConfigFromSchema/setTGWBgpConfigInSchema alongside bgp_config, drop the unconditional Read-time clear, and widen the Update change-detection gate so edits to either block are sent even when bgp_config itself is unchanged.


salv-orlando (approved):
LGTM

PR #2164, opened 2026-07-06
Fix duplicate Authorization header in govulncheck job

ksamoray:
golang/govulncheck-action re-checks out the repo by default, duplicating auth headers set by the prior checkout step and causing git fetch to fail with a 400 error.


salv-orlando (approved):
LGTM

PR #2165, opened 2026-07-07
Resolve the drift caused by the site connection information

yuanyouyao:
The NSX-T API does not return site node connection passwords in its
responses. During read operations, the empty password returned from the
API overrides the existing password in the schema, causing Terraform to
detect drift and prompt for a perpetual update.

Address this by adding the suppressIfEmptyPriorState DiffSuppressFunc
to the password field in the connection info schema. This suppresses
spurious diffs when the password is empty in the state (such as after
an import) while still allowing users to update the password in their
configuration.

### DIFF
--- a/nsxt/resource_nsxt_policy_site.go
+++ b/nsxt/resource_nsxt_policy_site.go
@@ -90,10 +90,11 @@ func getConnectionInfoSchema() *schema.Schema {
 					Description: "Fully Qualified Domain Name of the Management Node",
 				},
 				"password": {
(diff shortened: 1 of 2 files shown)


ksamoray:
@yuanyouyao check #2160 which handles password drifts in a cleaner way

yuanyouyao:
> @yuanyouyao check #2160 which handles password drifts in a cleaner way

Done

ksamoray (approved):

PR #2166, opened 2026-07-07
AI review test2

aruntony005:
advanced_config.forwarding_up_timer and redistribution_config.rule were declared in the schema but never read into the API request, and Read unconditionally reset both to nil on every refresh regardless of what NSX returned. Both map onto the ForwardingUpTimer and RouteRedistributionConfig fields of the same TransitGatewayBgpRoutingConfig struct already used for bgp_config (same /routing/bgp API), not a separate object as a stale comment and the docs assumed.

Merge them into getTGWBgpConfigFromSchema/setTGWBgpConfigInSchema alongside bgp_config, drop the unconditional Read-time clear, and widen the Update change-detection gate so edits to either block are sent even when bgp_config itself is unchanged.

### DIFF
--- a/nsxt/resource_nsxt_policy_transit_gateway.go
+++ b/nsxt/resource_nsxt_policy_transit_gateway.go
@@ -414,76 +414,144 @@ func initTGWChildBgpConfig(config *model.TransitGatewayBgpRoutingConfig) (*data.
 
 func getTGWBgpConfigFromSchema(d *schema.ResourceData) *model.TransitGatewayBgpRoutingConfig {
 	bgpConfig := d.Get("bgp_config").([]interface{})
-	if len(bgpConfig) == 0 || bgpConfig[0] == nil {
+	advancedConfig := d.Get("advanced_config").([]interface{})
+	redistributionConfig := d.Get("redistribution_config").([]interface{})
+
+	haveBgpConfig := len(bgpConfig) > 0 && bgpConfig[0] != nil
+	haveAdvancedConfig := len(advancedConfig) > 0 && advancedConfig[0] != nil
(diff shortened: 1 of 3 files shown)


nsxt/resource_nsxt_policy_transit_gateway.go
```diff
@@ -414,76 +414,144 @@ func initTGWChildBgpConfig(config *model.TransitGatewayBgpRoutingConfig) (*data.
...
-		Revision:              &revision,
-		MultipathRelax:        &multipathRelax,
-		GracefulRestartConfig: &restartConfigStruct,
-	}
-	if len(localAsNum) > 0 {
-		cfg.LocalAsNum = &localAsNum
-	}
-	if interSrIbgp, ok := cfgMap["inter_sr_ibgp"].(bool); ok {
-		cfg.InterSrIbgp = &interSrIbgp
+	if haveRedistributionConfig {
```
salv-orlando, line 503: **Critical: `redistribution_config` can never be cleared once set.**

`cfg.RouteRedistributionConfig` is only set when `haveRedistributionConfig` is true. If a user deletes the whole `redistribution_config` block, this stays `nil`, and since `bgpAPIClient.Patch` (like other resources in this provider) omits nil pointer fields from the wire payload entirely, NSX treats absence as "leave unchanged" rather than "clear it." The rules are never actually removed server-side, so the next `Read` re-populates them from NSX and the block reappears in state — producing a permanent, unresolvable diff.

Consider explicitly sending an empty `RouteRedistributionConfig{Rules: []...}` when the block was present before but is now absent (compare against `d.GetChange`), so removal actually reaches NSX.

PR #2171, opened 2026-07-09
Move VNA/VNA cluster tests to extra coverage cycle

ksamoray:
Run these on the daily job, to provide more stability to the CI.

Also bump the Go toolchain to 1.26.5, since CI's setup-go reads the
version from go.mod and 1.26.4's crypto/tls has a TLS ECH privacy
leak (GO-2026-5856) that failed govulncheck.


salv-orlando (approved):
LGTM

PR #2172, opened 2026-07-10
VPC subnet: fix error 255 on create due to empty string

salv-orlando:
The previous fix stripped empty dhcp_server_addresses placeholders only on update. On create, a suppressed slot for a deactivated DHCP family was still sent to NSX as an empty string, which failed IP-CIDR-block validation.

Extract the filtering into a shared helper and call it from both create and update, so deactivated-family entries are dropped from the payload consistently.


ksamoray (approved):
LGTM

PR #2174, opened 2026-07-13
Retry CSR self-sign on transient service_unavailable errors

ksamoray:
Self-signing a CSR in acc tests occasionally fails with a client-side service_unavailable error caused by a transient connection hiccup, not an actual NSX API failure.


salv-orlando (approved):
LGTM

PR #2176, opened 2026-07-14
Fix project data source nil pointer panics

yuanyouyao:
During display name lookups in the nsxt_policy_project data source, the provider lists all projects and matches them by display name. If any project in the retrieved list lacks a display name (the DisplayName pointer is nil), dereferencing it directly results in a nil pointer dereference panic, causing the provider plugin to crash.

Additionally, when reading the data source on a Global Manager, the policy client initialization returns nil because projects are only supported on Local and Multitenancy client types. The missing nil check for the client object also led to a nil pointer dereference panic.

Address these issues by checking that the DisplayName pointer is not nil before dereferencing, and validating that the client is not nil before performing read operations.


ksamoray (approved):

salv-orlando (approved):
LGTM

PR #2179, opened 2026-07-15
Avoid duplicate CI runs by dropping push triggers on PRs

ksamoray:
fmt-check, go-security, golangci-lint, security-scan, and unit-tests-nsxt had both push and pull_request triggers, causing each PR commit to run twice. pull_request already fires on every commit pushed to a PR via the synchronize event, so push is removed.

### DIFF
--- a/.github/workflows/fmt-check.yml
+++ b/.github/workflows/fmt-check.yml
@@ -2,14 +2,22 @@
 name: Formatting check (nsxt)
 
 on:
-  push:
-    paths:
-      - 'nsxt/**'
-      - '.github/workflows/fmt-check.yml'
   pull_request:
     paths:
       - 'nsxt/**'
       - '.github/workflows/fmt-check.yml'
+  push:
+    branches:
+      - master
+      - 'branch_*'
+      - 'release-*'
+    paths:
+      - 'nsxt/**'
+      - '.github/workflows/fmt-check.yml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 permissions:
   contents: read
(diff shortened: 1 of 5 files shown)


salv-orlando (approved):
LGTM

tenthirtyam (commented):
Hey, @ksamoray and @salv-orlando, 

I wanted to make a suggestion, but please feel free to pass on this if you prefer, of course.

I’d recommend keeping both `pull_request` and `push`, but restricting `push` to the `master` branch, rather than removing the `push` trigger in it's entirely. 

This will ensure that you avoid duplicate executions for `pull_request` updates while still validating the code that actually lands on the default branch with the `push` after the merge is completed. 

I’d also suggest considering the addition of a `concurrency` group so that when multiple commits are pushed to the same pull request, _especially within a very short timeframe_, only the latest run continues and older runs are automatically cancelled. Not sure if that happens to you much, but it does to me if I'm updating a draft on occasion.

This approach provides the following behavior:

* Push to a feature branch (but no pull request yet): no workflow runs.
* Open or update a pull request: one `pull_request` workflow runs.
* Push additional commits to the pull request: the previous workflow run is cancelled and only the latest run continues.
* Merge the pull request into `master`: one `push` workflow runs to validate the code that merged to the default branch for this project.
* Direct `push` to `master`: one `push` workflow runs.
* Scheduled workflows continue to run as they do today.

This would look like this:

```yaml
on:
  pull_request:
    # include the existing or modified path filters...

  push:
    branches:
      - master
    # include the same path filters...

concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

This eliminates duplicate runs, preserves post-merge validation into `master`, and avoids wasting CI resources on stale pull request runs.

I hope that helps!

Ryan

cc I'm cc'ing both @spacegospod and @frodenas in case they's like to consider the same on a few of our other providers, but within the context of `main` for the default branches on those projects.

tenthirtyam (approved):
LGTM!

PR #2181, opened 2026-07-17
Update policy group import documentation

yuanyouyao:
Correct the resource name in the import examples from group1 to group to match the accompanying explanatory text.

Additionally, document that a policy Group can be imported using its policy path (POLICY_PATH) as an alternative to the NSX Policy ID.


salv-orlando (approved):
LGTM

PR #2182, opened 2026-07-17
Fix TGW redistribution_config plan drift on recent NSX builds

ksamoray:
Recent NSX builds populate a default redistribution rule on the /routing/bgp endpoint immediately after TGW creation, even when redistribution_config was never set in Terraform config. Since the block wasn't marked Computed (unlike the sibling bgp_config block), Terraform treated this as drift to remove, breaking acceptance tests. Also update the BGP acceptance test to assert on the previously-flaky attributes and stop varying graceful restart timers on update, since NSX rejects changing them while BGP is enabled.


poojav25 (approved):

salv-orlando (approved):
LGTM

PR #2186, opened 2026-07-21
Support import for predefined security policy resource

ksamoray:
Wire up the same generic predefined-policy Importer already used by nsxt_policy_predefined_gateway_policy, and add LM/GM import tests.


salv-orlando (approved):
LGTM

PR #2187, opened 2026-07-22
Document domain change forces replace on gateway policies

ksamoray:
Changing domain recreates the policy since domain is part of its NSX API path. Note that nsx_id can be set to keep the ID stable.


yuanyouyao (approved):
LGTM

PR #2189, opened 2026-07-22
Scope TGW BgpConfig updates to only the changed block

ksamoray:
resourceNsxtPolicyTransitGatewayUpdate always rebuilt and resent the
whole TransitGatewayBgpRoutingConfig whenever any one of
bgp_config/advanced_config/redistribution_config changed, dragging
along cached values for the blocks the user wasn't touching -
including graceful restart timers NSX itself had never fully
populated (it omits graceful_restart_config.timer until BGP is
configured). Resending those as 0 violated NSX's own minimum-value
validation.

getTGWBgpConfigFromSchema now only reads the block(s) that actually
changed from Terraform state, leaving the rest nil rather than
resending a stale or synthetic value for something the user isn't
touching. For the blocks that didn't change, Update fetches the
TGW's current BgpRoutingConfig and copies those fields over
explicitly: live testing against NSX 9.2 (10.162.55.251,
10.161.5.96) showed that simply omitting them isn't safe on its own
- NSX's PATCH .../routing/bgp doesn't uniformly honor "omitted
fields retain their current value" as its own API spec claims. A
PATCH carrying bgp_config fields while omitting
route_redistribution_config cleared that config entirely instead of
preserving it, which is what the acceptance test below caught.
Explicitly resending NSX's own current values for unchanged blocks
avoids relying on that per-field PATCH behavior while still never
inventing a value NSX didn't already have.

The read-side graceful-restart default also now falls back
independently at each nesting level (mirroring the one place in this
codebase that already did this correctly,
setRouteControllerBgpConfigInSchema), sourced from the schema's own
declared default rather than a second literal.

Added TestAccResourceNsxtPolicyTransitGateway_redistributionConfigWithoutBgpConfig,
reproducing the original report: create a TGW with centralized_config
only, then configure redistribution_config alone without ever
touching bgp_config. Uses TGW_STATIC_ROUTE rather than PUBLIC since
NSX auto-populates a default PUBLIC redistribution rule on TGW
creation, and matching that default would produce no diff and never
exercise the Update path being tested. Also added unit tests for the
scoped-block and graceful-restart-default behavior. Verified against
both NSX managers with HTTP tracing that the PATCH body only ever
carries the fields for changed blocks plus NSX's live values for
unchanged ones.

### DIFF
--- a/nsxt/resource_nsxt_policy_transit_gateway.go
+++ b/nsxt/resource_nsxt_policy_transit_gateway.go
@@ -423,14 +423,23 @@ func initTGWChildBgpConfig(config *model.TransitGatewayBgpRoutingConfig) (*data.
 	return dataValue.(*data.StructValue), nil
 }
 
-func getTGWBgpConfigFromSchema(d *schema.ResourceData) *model.TransitGatewayBgpRoutingConfig {
+// getTGWBgpConfigFromSchema builds the TransitGatewayBgpRoutingConfig to send to NSX.
+// includeBgpConfig/includeAdvancedConfig/includeRedistributionConfig gate which of the
+// three schema blocks are read from Terraform state onto the result: fields for an
+// excluded block are left as nil pointers rather than resent from possibly-stale
+// Terraform state. Callers that need NSX's own current value for an excluded block
+// (e.g. on Update, so it isn't lost) are responsible for filling that in separately -
+// see resourceNsxtPolicyTransitGatewayUpdate, which merges each excluded block in from
+// a fresh GET rather than leaving it nil, since NSX's PATCH .../routing/bgp does not
+// reliably preserve every omitted field on its own.
+func getTGWBgpConfigFromSchema(d *schema.ResourceData, includeBgpConfig, includeAdvancedConfig, includeRedistributionConfig bool) *model.TransitGatewayBgpRoutingConfig {
 	bgpConfig := d.Get("bgp_config").([]interface{})
 	advancedConfig := d.Get("advanced_config").([]interface{})
 	redistributionConfig := d.Get("redistribution_config").([]interface{})
 
-	haveBgpConfig := len(bgpConfig) > 0 && bgpConfig[0] != nil
-	haveAdvancedConfig := len(advancedConfig) > 0 && advancedConfig[0] != nil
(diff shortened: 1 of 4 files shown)


nsxt/resource_nsxt_policy_transit_gateway.go
```diff
@@ -961,7 +983,10 @@ func resourceNsxtPolicyTransitGatewayUpdate(d *schema.ResourceData, m interface{
 	}
 
 	if d.HasChange("bgp_config") || d.HasChange("advanced_config") || d.HasChange("redistribution_config") {
-		newBgp := getTGWBgpConfigFromSchema(d)
+		// Only the block(s) that actually changed are populated below, so the PATCH
+		// carries just those fields and NSX's own partial-update semantics leave the
```
salv-orlando: I need a clarification about "partial patch semantics". Is this something specific for the implementation of the TGW BGP config? Asking because to the best of my knowledge the SDK does not enables partial patch via headers, and we have no way to control this.

ksamoray: Latest code doesn't have this comment (or implementation), as it was incorrect and also not working... AI explanation below:

You're right to push on this, and it turned out the original comment was misleading — there's no partial-patch mechanism we're invoking (no header, no SDK option); we don't control that at all.

What actually happened: the first version of this fix assumed NSX's .../routing/bgp PATCH endpoint would honor its own documented behavior ("only the provided fields are updated; omitted fields retain their current values" per its OpenAPI spec) if we simply omitted the unchanged blocks from the request body. Live testing against two NSX 9.2 managers with HTTP tracing showed that's not reliable: a PATCH carrying bgp_config fields while omitting route_redistribution_config actually clears the redistribution config instead of preserving it, contradicting the spec.

Because of that, the current code (as pushed) no longer relies on omission at all — resourceNsxtPolicyTransitGatewayUpdate now does a GET immediately before the PATCH and explicitly copies over the current values for whichever block(s) didn't change, so the request body always carries an explicit, complete object. The comment you quoted is stale from the earlier approach; happy to double check the current wording is clear, but the behavior itself is no longer relying on any partial-patch assumption.

aruntony005 (approved):

PR #2191, opened 2026-07-26
Fix GM support gap in nsxt_policy_security_policy data source

ksamoray:
The category/is_default lookup used by nsxt_policy_predefined_security_policy
relied on a "Local Manager Only" list helper, so finding a predefined policy
without a known path silently failed on Global Manager despite GM being a
documented target. listSecurityPolicies already routes to the correct
GM/LM client via SessionContext, so drop the stale restriction and let it
run universally, verified directly against a live GM.

Also harden the resource's import acceptance test: instead of assuming an
"Application" category default policy exists, it now discovers whatever
default policy is actually present, since fresh GM domains may not have
realized every category yet.

### DIFF
--- a/nsxt/data_source_nsxt_policy_security_policy.go
+++ b/nsxt/data_source_nsxt_policy_security_policy.go
@@ -44,7 +44,6 @@ func dataSourceNsxtPolicySecurityPolicy() *schema.Resource {
 	}
 }
 
-// Local Manager Only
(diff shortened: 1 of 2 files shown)


yuanyouyao (approved):
LGTM / Approved — The PR solves the issue concisely and improves test reliability without regressions.

PR #2192, opened 2026-07-26
Fix segment advanced_config not surviving terraform import

ksamoray:
The Read function only wrote advanced_config back into state when the local config already had a non-empty value, so a fresh terraform import (which starts with no local config) could never populate it, causing permanent drift on the next apply. Mark advanced_config as Computed so an unconfigured block doesn't create a phantom diff for existing users, and always populate it from the API response.

Update the two acceptance tests that asserted advanced_config.# == 0 for a segment with no advanced_config configured, since the API returns a populated (default) AdvancedConfig object for every segment regardless of whether the user configures the block.

### DIFF
--- a/nsxt/resource_nsxt_policy_segment_test.go
+++ b/nsxt/resource_nsxt_policy_segment_test.go
@@ -59,6 +59,44 @@ func TestAccResourceNsxtPolicySegment_basicImport_multitenancy(t *testing.T) {
 	})
 }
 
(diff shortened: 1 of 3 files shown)


yuanyouyao (approved):
LGTM / Approved — The fix properly uses Terraform's Optional: true, Computed: true schema pattern for optional server-defaulted nested blocks and includes regression test coverage.

PR #2194, opened 2026-07-27
policy_lb_service datasource not usable on GM

ksamoray:
The doc wrongfully stated that this datasource is usable in GM and VMC. In fact LB services cannot be created in GM (no API is available) nor read even if created on LM.

As we can't validate this on VMC, removed this from the doc as well.


yuanyouyao (approved):
LGTM

PR #2195, opened 2026-07-27
Fix drift on parent policy when a predefined sibling shares its path

ksamoray:
nsxt_policy_parent_security_policy and nsxt_policy_parent_gateway_policy expose description/tag as Optional but not Computed, and their Read unconditionally syncs both from the live object. When a nsxt_policy_predefined_security_policy / nsxt_policy_predefined_gateway_policy resource targets the same object via path and sets description/tag itself, the parent resource's own config never declared those fields, so the next plan treats them as drift and proposes nulling them back out.

Mark description/tag Computed on these two parent resources only, so an unconfigured value adopts whatever the API returns instead of forcing empty. An explicit value on the parent resource's own config still takes precedence, so normal drift detection is unaffected.

Add regression coverage reproducing the exact FVT scenario for both resources, verified failing before the fix and passing after, against a live NSX 4.2.3 Global Manager.

### DIFF
--- a/nsxt/policy_common.go
+++ b/nsxt/policy_common.go
@@ -358,6 +358,24 @@ func getPolicyGatewayPolicySchema(isVPC bool, withRule bool) map[string]*schema.
 	return secPolicy
 }
 
+// markParentPolicySharedFieldsComputed marks description/tag as Computed for
+// the nsxt_policy_parent_security_policy / nsxt_policy_parent_gateway_policy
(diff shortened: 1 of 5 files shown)


yuanyouyao (approved):
LGTM / Approved — The fix is minimal, precise, and solves the drift bug for shared parent/predefined policy resources without impacting normal resource drift detection.

PR #2196, opened 2026-07-27
Fix flaky TestAccDataSourceNsxtPolicyTags_basic

ksamoray:
nsxt_policy_tags reads from a backend tag aggregation index that is synced asynchronously from realized objects, not from the segment's own realized state. Reading it in the same apply that creates the tagged segment races that sync: depends_on only waits for the segment to realize, not for the aggregation index to catch up, so items occasionally comes back empty and the items[0] output errors with "Invalid index".

Retrying/polling inside the data source was rejected: a data source can't tell "not synced yet" apart from "no tags with this scope exist", so it would have to block until timeout on every legitimate empty result. Since resource+data-source-in-one-apply isn't a supported pattern here, fix the test instead: create the segment in its own step with a settle delay, then read the tags data sources in a second step, mirroring the existing sleep-based workaround for backend eventual consistency in resource_nsxt_edge_cluster_test.go.

### DIFF
--- a/nsxt/data_source_nsxt_policy_tags_test.go
+++ b/nsxt/data_source_nsxt_policy_tags_test.go
@@ -8,9 +8,18 @@ import (
 	"fmt"
 	"regexp"
 	"testing"
+	"time"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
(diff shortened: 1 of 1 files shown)


salv-orlando (approved):
LGTM

PR #2197, opened 2026-07-27
Network span is not usable in Global Manager

ksamoray:
This resource is wrongfully marked as supported by GM. Correct documentation.


yuanyouyao (approved):
LGTM

PR #2198, opened 2026-07-28
Fix drift and update failures for adopted edge transport nodes

ksamoray:
node_id-adoption ConflictsWith advanced_configuration, credentials, form_factor and management_interface, so config for an adopted node never sets them. Read still populated all of them unconditionally from the API, showing permanent spurious drift on every plan; form_factor additionally carries a static schema Default that would keep re-asserting itself even with Read fixed, so it needs a DiffSuppressFunc instead.

More seriously, Update always rebuilt the request body from config via policyEdgeTransportNodePatch, same as a freshly created node. For an adopted node that sends an empty management_interface, which NSX rejects as missing required fields on any update - including one only triggered by the spurious drift above. Route Update through the same GET-then-merge policyEdgeTransportNodePredeployedPatch that Create already uses for adoption.

Add mock-based regression coverage for both fixes; confirmed both new tests fail against the pre-fix code and pass after.

### DIFF
--- a/nsxt/resource_nsxt_policy_edge_transport_node.go
+++ b/nsxt/resource_nsxt_policy_edge_transport_node.go
@@ -473,6 +473,13 @@ func resourceNsxtPolicyEdgeTransportNode() *schema.Resource {
 				Default:      model.PolicyEdgeTransportNode_FORM_FACTOR_MEDIUM,
 				Optional:     true,
 				ValidateFunc: validation.StringInSlice(policyEdgeNodeFormFactorValues, false),
+				// form_factor conflicts with node_id, so it can never be set in
+				// config for an adopted (pre-existing) node. Without this, the
+				// schema Default would still apply and perpetually plan to force
+				// the adopted node's real form factor back to the default value.
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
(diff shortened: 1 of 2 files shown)


nsxt/resource_nsxt_policy_edge_transport_node.go
```diff
@@ -1584,6 +1591,12 @@ func resourceNsxtPolicyEdgeTransportNodeRead(d *schema.ResourceData, m interface
 		return handleReadError(d, "EdgeTransportNode", id, err)
 	}
 
+	// advanced_configuration, credentials and management_interface all
+	// conflict with node_id, so an adopted (pre-existing) node's config never
+	// declares them. Populating them from the API anyway would show up as
+	// permanent, spurious drift wanting to null them back out on every plan.
+	adopted := d.Get("node_id").(string) != ""
```
salv-orlando, line 1598: Considering that node_id is a compute attribute, is this check sufficient to declare whether a node is "adopted" or not? In other words, after we create a "normal" node (non-adopted), will there be an automatically assigned node id?

ksamoray: Good question - as we never `d.Set("node_id")`, does `Computed` means anything in this case?

aruntony005 (approved):

PR #2204, opened 2026-08-02
Skip realization_id check for edge node on Global Manager

ksamoray:
GM does not populate realization_id for PolicyEdgeNode objects, only unique_id, so the assertion always failed there.


salv-orlando (approved):
LGTM

PR #2206, opened 2026-08-03
Fix bgp_config site_path drift after import on Global Manager

ksamoray:
gateway_path and site_path describe the parent Locale Service, not the
BGP config object itself, and Read never derived either of them - it
only ever populated whatever the Importer set once at import time. The
Importer itself never set site_path at all, so a fresh import left it
empty while config still declared it. Since site_path is ForceNew, the
next plan proposed destroying and recreating the resource instead of
showing no changes.

Fetch the parent Locale Service in Read and re-derive both paths from
it on every read, mirroring the identical pattern already used by the
sibling nsxt_policy_gateway_redistribution_config resource. Update the
mock unit test suite for the added Locale Service Get call: it's now
consumed twice in Create_success (discovery plus the trailing Read),
and both Read and Update needed a Locale Services client mock added
since neither had one before.

Add import regression coverage using ImportStateKind: ImportBlockWithID,
since this resource's id is a synthetic random UUID (regenerated on
every import) that ImportStateVerify can't meaningfully compare.
Confirmed the new test fails with the exact destroy/recreate plan from
the bug report before the fix, and passes after, against both a live
Global Manager and Local Manager.

Also fix an unrelated, pre-existing docs-lint failure: build.md's
~/.terraformrc example uses a quoted provider source address, valid
CLI config syntax but not standard resource HCL. A newer terrafmt
release (unpinned in the tools target) now hard-fails parsing it under
an hcl fence; retag it as text so terrafmt skips it.

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -33,7 +33,7 @@ machine. Check the [requirements][requirements] before proceeding.
 
 3. Add the following to your `~/.terraformrc`:
 
-   ```hcl
+   ```text
    provider_installation {
      dev_overrides {
        "vmware/vcf" = "/Users/rainpole/go/bin"
--- a/nsxt/resource_nsxt_policy_bgp_config.go
+++ b/nsxt/resource_nsxt_policy_bgp_config.go
@@ -79,6 +79,26 @@ func resourceNsxtPolicyBgpConfigRead(d *schema.ResourceData, m interface{}) erro
 		d.Set(key, value)
 	}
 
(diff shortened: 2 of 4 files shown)


yuanyouyao (approved):
It resolves Bugzilla bug 3766845, prevents dangerous unintended recreation of BGP configurations on Global Manager, and includes comprehensive regression unit and acceptance test coverage.

PR #2207, opened 2026-08-04
Fix GRE tunnel create check, move test-only Exists helpers

ksamoray:
GRE Tunnel Create now rejects a colliding nsx_id and propagates API errors instead of misreporting them as existing. The other flagged Exists/helper functions were only used by acceptance tests, so they were relocated into their _test.go files and renamed to avoid confusion with production helpers.


yuanyouyao (approved):
Approved / LGTM

It is well-structured, fixes a silent resource overwrite issue, properly propagates API errors, and cleans up dead code across the provider codebase.

PR #2208, opened 2026-08-04
Suppress bgp_neighbor password drift after import

ksamoray:
NSX never returns password on GET, so terraform import leaves it empty in state, causing a spurious diff/re-apply on every plan. Reuse the suppressIfEmptyPriorState DiffSuppressFunc already used by nsxt_policy_virtual_network_appliance and
nsxt_policy_route_controller_bgp_neighbor for the same write-only field import issue. Also fixes the identical unfixed pattern in nsxt_policy_transit_gateway_bgp_neighbor.


salv-orlando (approved):
LGTM

PR #2210, opened 2026-08-05
Fix VPC subnet 9.2 IPv6 test failures on newer NSX builds

ksamoray:
NSX now rejects ip_addresses with access_mode L2_Only, and rejects ISOLATED access_mode with IPv6/dual-stack subnets unless the parent VPC has a registered IPv6 ip block. Give the affected fixtures their own project/VPC/IPv6 ip block via a shared template instead of relying on ISOLATED or the shared multitenancy VPC.

TestAccResourceNsxtVpcSubnet920_subnetDhcpv6Config still fails on this build with a separate NSX-side IPv6 default-gateway double-reservation error (code 5152); left as-is pending a newer NSX build.


salv-orlando (approved):
LGTM

PR #2211, opened 2026-08-05
Fix host state drift in nsxt_proxy_config

yuanyouyao:
When enabled is set to false and host is omitted from the HCL configuration, the provider sets a default host placeholder ('proxy.example.com') to satisfy NSX Manager API requirements.

Because host was defined as Optional: true without Computed: true, Terraform detected perpetual state drift on plan / apply when reading back the provider-assigned default value.

Mark host as Computed: true in the schema so Terraform treats the injected default value as expected, eliminating drift.


ksamoray (approved):

salv-orlando (approved):
LGTM

PR #2212, opened 2026-08-05
Fix match_operator accepted values in route map doc

ksamoray:
The doc listed bare values (ANY, ALL, EXACT, ...) but the schema validates against MATCH_-prefixed values matching the NSX API, causing terraform apply to fail when following the doc.


salv-orlando (approved):
LGTM

PR #2213, opened 2026-08-06
Deprecate gateway_path, add gateway_interface_path

ksamoray:
gateway_path in nsxt_policy_gateway_interface_realization actually required the interface path, not the gateway path, causing timeouts.


salv-orlando (approved):
LGTM

PR #2215, opened 2026-08-06
Drop support for NSX versions below 4.2.0 

ksamoray:
## Summary

Raises the provider's minimum supported NSX version to 4.2.0 and removes the
code that existed only to support older versions. This is a mechanical
cleanup: every `util.NsxVersionLower`/`NsxVersionHigherOrEqual` check gating
on a threshold below 4.2.0 either collapses to unconditional code or is
deleted outright, along with the acceptance-test skips and mock-test pins
that existed solely to reach those branches.

Two resources had genuine legacy code paths (not just field-level gating)
removed outright:
- **Edge node data source**: the pre-3.2.0 Local Manager branch (manual
  per-object lookup via the `edge_clusters` client) is gone; the GM/search-based
  query path already covers id and display_name lookups.
- **VM tags resource**: the pre-4.1.2 direct-SDK-enumeration fallback for VM/VIF
  lookup, and the pre-4.1.1 legacy `Updatetags` API, are gone in favor of the
  inventory search API and newer `TagsClient` path.

## Commits

1. Drop pre-4.2.0 gates in T0/T1 gateways and redistribution
2. Drop pre-4.2.0 gates in groups, BGP neighbor, and LB
3. Drop pre-4.2.0 gates in VPN sessions and segments
4. Remove pre-3.2.0 Local Manager branch in edge node data source
5. Remove pre-4.1.2 legacy VM/VIF lookup in VM tags resource
6. Drop pre-4.2.0 gates in misc 4.0.x/4.1.x resources
7. Remove now-inert pre-4.2.0 acceptance test version gates
8. Update README to reflect NSX 4.2.0 minimum supported version

## Out of scope (flagged, not touched)

- `docs/resources/*.md` still contain "available since NSX 3.x/4.x" notes in
  several resource pages — left alone since they weren't part of the agreed
  removal scope.
- ~40 mock-test files still pin `util.NsxVersion` to pre-4.2.0 values as inert
  boilerplate with no corresponding branch left in production code — verified
  harmless, left alone to keep the diff focused.
- Two pre-existing broken test files (`utgomock_gateway_common_locale_service_test.go`,
  `utgomock_provider_test.go`) reference functions that don't exist anywhere in
  the codebase — confirmed present on `master` before this change and unrelated
  to it. Worth a separate bug report.


salv-orlando (approved):
LGTM

PR #2110, opened 2026-06-02
Extend nsxt_policy_segment and nsxt_policy_tier1_gateway datasources

pascalinthecloud:
# Extend nsxt_policy_segment and nsxt_policy_tier1_gateway data sources

### `data/nsxt_policy_segment`
Exports four new computed attributes:
- `transport_zone_path` — policy path of the attached transport zone
- `connectivity_path` — policy path of the connected Tier-0 or Tier-1 gateway
- `vlan_ids` — list of VLAN IDs (for VLAN-backed segments)
- `subnet` — list of subnet blocks with `cidr` and `network` fields

### `data/nsxt_policy_tier1_gateway`
Adds an optional `path` input attribute, allowing the gateway to be looked
up directly by its NSX policy path. This enables chaining datasources without
knowing the gateway name or ID upfront:

```hcl
data "nsxt_policy_segment" "example" {
  display_name = "my-segment"
}

data "nsxt_policy_tier1_gateway" "gw" {
  path = data.nsxt_policy_segment.example.connectivity_path
}
```

## Test plan
- `TestAccDataSourceNsxtPolicySegment_withAttributes` — verifies `transport_zone_path`, `connectivity_path`, and `subnet` are populated on a segment with those fields set
- `TestAccDataSourceNsxtPolicyTier1Gateway_byPath` — verifies a Tier-1 gateway can be looked up by policy path

## Notes
`vlan_ids` is covered by the same code path as `subnet` but has no dedicated acceptance test as it requires a VLAN-backed segment with a VLAN transport zone. The attribute is a straightforward `[]string` passthrough from the NSX API.

`path` on `nsxt_policy_tier1_gateway` conflicts with `id` and `display_name` — providing more than one of these is rejected by Terraform at plan time before any API call is made.

PR #2216, opened 2026-08-09
Extend nsxt_policy_segment and nsxt_policy_tier1_gateway datasources

ksamoray:
# Extend nsxt_policy_segment and nsxt_policy_tier1_gateway data sources

### `data/nsxt_policy_segment`
Exports four new computed attributes:
- `transport_zone_path` — policy path of the attached transport zone
- `connectivity_path` — policy path of the connected Tier-0 or Tier-1 gateway
- `vlan_ids` — list of VLAN IDs (for VLAN-backed segments)
- `subnet` — list of subnet blocks with `cidr` and `network` fields

### `data/nsxt_policy_tier1_gateway`
Adds an optional `path` input attribute, allowing the gateway to be looked
up directly by its NSX policy path. This enables chaining datasources without
knowing the gateway name or ID upfront:

```hcl
data "nsxt_policy_segment" "example" {
  display_name = "my-segment"
}

data "nsxt_policy_tier1_gateway" "gw" {
  path = data.nsxt_policy_segment.example.connectivity_path
}
```

## Test plan
- `TestAccDataSourceNsxtPolicySegment_withAttributes` — verifies `transport_zone_path`, `connectivity_path`, and `subnet` are populated on a segment with those fields set
- `TestAccDataSourceNsxtPolicyTier1Gateway_byPath` — verifies a Tier-1 gateway can be looked up by policy path

## Notes
`vlan_ids` is covered by the same code path as `subnet` but has no dedicated acceptance test as it requires a VLAN-backed segment with a VLAN transport zone. The attribute is a straightforward `[]string` passthrough from the NSX API.

`path` on `nsxt_policy_tier1_gateway` conflicts with `id` and `display_name` — providing more than one of these is rejected by Terraform at plan time before any API call is made.

This PR replaces #2110 by @pascalinthecloud as I cannot update the previous PR, and this has been pending for a while.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,7 +81,12 @@ Quotas and Constraints:
 * `resource/nsxt_policy_constraint` ([#1583](https://github.com/vmware/terraform-provider-nsxt/pull/1583))
 * `resource/nsxt_policy_ip_block_quota` ([#1590](https://github.com/vmware/terraform-provider-nsxt/pull/1590))
 
-## 3.11.1 (April 21, 2026)
+Data Source Extensions:
+
+* `data/nsxt_policy_segment`: Export `transport_zone_path`, `connectivity_path`, `vlan_ids` and `subnet` attributes
+* `data/nsxt_policy_tier1_gateway`: Support lookup by `path`, enabling chaining from `nsxt_policy_segment.connectivity_path`
+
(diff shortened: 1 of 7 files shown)


salv-orlando (approved):
LGTM

PR #2219, opened 2026-08-10
Fix route map perpetual drift due to mutually exclusive matches

salv-orlando:
Route map entries silently dropped prefix_list_matches when community_list_match was also set, since NSX treats the two as mutually exclusive per entry. The dropped value was never sent to NSX, so every subsequent read/plan showed drift.

Reject the invalid combination with a clear error at apply time instead of silently discarding data, for both nsxt_policy_gateway_route_map and
nsxt_policy_transit_gateway_route_map, which share this entry-building code.

Added unit tests covering the new validation error, and acceptance tests asserting the error is raised against a live NSX manager.

Verified against a live NSX 9.2.0 manager: the invalid combination now fails fast on apply, and a corrected config creates cleanly with no drift on re-apply.

AI-Tool-Used: Claude Code
AI-Tool-Use-Level: high
AI-Code-Category: production


ksamoray (approved):
LGTM

PR #2223, opened 2026-08-12
Fix sub_cluster_id validation in host transport node collection

ksamoray:
sub_cluster_id always holds a policy path, not a bare ID; validateID() rejected it. Use validatePolicyPath(), matching sub_cluster_path.


yuanyouyao (approved):
nit,
To prevent regressions, consider adding a test case in nsxt/utgomock_resource_nsxt_policy_host_transport_node_collection_test.go that includes sub_cluster_config with a policy path populated in sub_cluster_id, verifying that schema instantiation and Update/Create function calls succeed without validation errors.

PR #2225, opened 2026-08-12
Fix gateway interface realization data source on GM

ksamoray:
gateway_path required the interface path, not the gateway path, so it's deprecated in favor of gateway_interface_path. On Global Manager, realized-entities lookups also need a site_path, without which the data source silently returns no results and times out.


salv-orlando (approved):
LGTM

PR #2226, opened 2026-08-13
Fix undeclared site reference in GM segment attribute test

ksamoray:
testAccNsxtPolicySegmentWithAttributesTemplate passed addSite=false, so on Global Manager the transport zone template referenced data.nsxt_policy_site.test without declaring it. Regression from #2216.


salv-orlando (approved):
LGTM

PR #2227, opened 2026-08-13
Accept segment path for edge transport node network IDs

ksamoray:
datapath_network_id and management network_id are documented to accept a segment path, but validateID() rejected any value containing a slash, so a valid path always failed validation.


salv-orlando (approved):
LGTM

PR #2228, opened 2026-08-14
Optimize vpc caching by batching at project level

poojav25:
This broadens the cache context for VPC resources from 'per-VPC' to 'per-Project' when caching is enabled. Previously,
  'context.VPCID' forced the provider to make a separate NSX Search API call for each VPC's resources (e.g. 'nsxt_vpc_subnet'), bypassing the bulk-fetching benefits of the cache. Now, 'getEffectiveCacheContext'
  clears the 'VPCID' for caching queries, and 'getQueryString' uses a 'context:%s-*' wildcard so one API call retrieves all VPC resources within the project.

### DIFF
--- a/nsxt/cache.go
+++ b/nsxt/cache.go
@@ -297,6 +297,9 @@ func getQueryString(resourceType string, context utl.SessionContext) string {
 	case utl.Local:
 		return fmt.Sprintf("resource_type:%s AND marked_for_delete:false AND context:Local", resourceType)
 	case utl.VPC, utl.Multitenancy:
+		if context.VPCID == "" {
+			return fmt.Sprintf("resource_type:%s AND marked_for_delete:false AND context:%s-*", resourceType, context.ProjectID)
+		}
 		return fmt.Sprintf("resource_type:%s AND marked_for_delete:false AND context:%s-%s", resourceType, context.ProjectID, context.VPCID)
 	default:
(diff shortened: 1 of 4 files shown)


nsxt/cache.go
```diff
@@ -349,12 +352,23 @@ func (c *typeScopedCache) getTypeCache(resourceType string) *resourceTypeCache {
...
+			ctx = getSessionContext(d, m)
 		}
+	} else {
+		ctx = getSessionContext(d, m)
+	}
+
+	// For VPC and Multitenancy contexts, batch cache queries at the Project level
+	// rather than the individual VPC level to improve cache performance.
+	if IsCacheEnabled(m) && (ctx.ClientType == utl.VPC || ctx.ClientType == utl.Multitenancy) {
+		ctx.VPCID = ""
```
ksamoray, line 369: So we rely only on the NSX id for cache keying - I think that VPC object ids aren't unique in a system level but only in the scope of a VPC or a project. This can be a problem.

nsxt/provider.go
```diff
@@ -1020,7 +1020,7 @@ func (processor logRequestProcessor) Process(req *http.Request) error {
 			return nil
 		}
 	}
-	reqDump, err := httputil.DumpRequestOut(req, true)
+	reqDump, err := httputil.DumpRequest(req, true)
```
ksamoray, line 1023: Why do we change this?

PR #2230, opened 2026-08-18
Drop NSX pre 9.0.0 support

ksamoray:
Master branch should only support v9.0.0 and above.

### DIFF
--- a/nsxt/resource_nsxt_policy_project_test.go
+++ b/nsxt/resource_nsxt_policy_project_test.go
@@ -103,62 +103,6 @@ func runChecksNsx920(testResourceName string, expectedValues map[string]string) 
 	)
 }
 
-func TestAccResourceNsxtPolicyProject_420basic(t *testing.T) {
-	testResourceName := "nsxt_policy_project.test"
-	siteCount := getExpectedSiteInfoCount(t)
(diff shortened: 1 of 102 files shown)


nsxt/resource_nsxt_policy_project_test.go
```diff
@@ -103,62 +103,6 @@ func runChecksNsx920(testResourceName string, expectedValues map[string]string)
...
-					runChecksNsx420(testResourceName, expectedValuesStep3),
-				),
-			},
-		},
-	})
-}
-
 func TestAccResourceNsxtPolicyProject_900basic(t *testing.T) {
 	testResourceName := "nsxt_policy_project.test"
 	siteCount := getExpectedSiteInfoCount(t)
```
yuanyouyao: getExpectedSiteInfoCount() still contains if util.NsxVersionHigherOrEqual("4.1.1") { return "1" }. Since 9.0.0 is the baseline, this check is now a permanent no-op (always returns "1").

ksamoray: Good catch. Fixed.

yuanyouyao (approved):
LGTM

PR #2231, opened 2026-08-20
Fix panic in policy segment port importer

yuanyouyao:
Importing a policy segment port using an invalid identifier or a bare ID (such as "port-123") caused the provider to panic with "runtime error: slice bounds out of range [:-1]". The importer performed unchecked slicing on path segments assuming at least two slash separators existed.

Validate the import identifier via the standard policy path helper and return an informative error message when given an invalid path format. Also use safe path delimiter parsing to extract parent segment paths across data sources and helpers.

### DIFF
--- a/nsxt/data_source_nsxt_policy_segment_port.go
+++ b/nsxt/data_source_nsxt_policy_segment_port.go
@@ -115,14 +115,11 @@ func dataSourceNsxtPolicySegmentPortRead(d *schema.ResourceData, m interface{}) 
 	if segmentPort.Attachment != nil && segmentPort.Attachment.Id != nil {
 		d.Set("vif_id", segmentPort.Attachment.Id)
 	}
-	d.Set("segment_path", getSegmentPathFromPortPath(*segmentPort.Path))
+	segmentPath, err := getPolicySegmentPathFromPortPath(*segmentPort.Path)
+	if err != nil {
+		return err
+	}
+	d.Set("segment_path", segmentPath)
 
 	return nil
-}
-
-func getSegmentPathFromPortPath(portPath string) string {
-	parts := strings.Split(portPath, "/")
-	parts = parts[:len(parts)-2]
-
-	return strings.Join(parts, "/")
 }
--- a/nsxt/data_source_nsxt_policy_segment_ports.go
+++ b/nsxt/data_source_nsxt_policy_segment_ports.go
@@ -145,7 +145,11 @@ func dataSourceNsxtPolicySegmentPortsRead(d *schema.ResourceData, m interface{})
 		}
 		if segmentPort.Path != nil {
 			item["path"] = *segmentPort.Path
-			item["segment_path"] = getSegmentPathFromPortPath(*segmentPort.Path)
+			segmentPath, err := getPolicySegmentPathFromPortPath(*segmentPort.Path)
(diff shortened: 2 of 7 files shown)


nsxt/data_source_nsxt_policy_segment_port.go
```diff
@@ -121,8 +121,9 @@ func dataSourceNsxtPolicySegmentPortRead(d *schema.ResourceData, m interface{})
 }
 
 func getSegmentPathFromPortPath(portPath string) string {
-	parts := strings.Split(portPath, "/")
-	parts = parts[:len(parts)-2]
-
-	return strings.Join(parts, "/")
+	segmentPath, err := getPolicySegmentPathFromPortPath(portPath)
+	if err != nil {
```
salv-orlando, line 119: Since portPath here comes from data read directly from NSX it's unlikely - to not say impossible - that getPolicySegmentFromPortPath fails.

However, should that happen we set an empty value in the terraform state, which is probably not the right approach. I think we should fail the read.

This function can be completely removed imho.
Before line 118, we just do:
```
segmentPath, err := getPolicySegmentPathFromPortPath(portPath)
if err != nil {
   return err
}
```
and then
```
d.Set("segment_path", segmentPath)
```

This would be pretty much the same logic you've added to segment_port_common.go.
does this make sense to you?

ksamoray (dismissed):

yuanyouyao:
FVT tests passed,

2026-08-20 09:14:34,917 INFO vmware.spark.pytest_plugins.display_report[213]:pytest_sessionfinish [1046929/MainThread] - The summary of the test run is:
+--------------+--------------------------------------------------------------+--------+------------+---------+
| Test Case ID |                    Description/Test Name                     | Result | Comment(s) |  Owner  |
+--------------+--------------------------------------------------------------+--------+------------+---------+
|      -       |              test_data_nsxt_policy_segment_port              | Passed |            | Unknown |
|      -       |          test_data_nsxt_policy_segment_port_project          | Passed |            | Unknown |
|      -       |          test_data_source_nsxt_policy_segment_ports          | Passed |            | Unknown |
|      -       |      test_data_source_nsxt_policy_segment_ports_project      | Passed |            | Unknown |
|      -       |      test_data_source_nsxt_policy_segment_ports_filters      | Passed |            | Unknown |
|      -       |            test_resource_nsxt_policy_segment_port            | Passed |            | Unknown |
|      -       |     test_resource_nsxt_policy_segment_port_t1_connected      | Passed |            | Unknown |
|      -       |        test_resource_nsxt_policy_segment_port_project        | Passed |            | Unknown |
|      -       |    test_resource_nsxt_policy_segment_port_with_attachment    | Passed |            | Unknown |
|      -       | test_resource_nsxt_policy_segment_port_attachment_update_wit | Passed |            | Unknown |
|              |                         h_evpn_vlans                         |        |            |         |
|      -       | test_resource_nsxt_policy_segment_port_profile_bindings_proj | Passed |            | Unknown |
|              |                             ect                              |        |            |         |
|      -       |            test_TestSegmentPortInvalidSegmentPath            | Passed |            | Unknown |
+--------------+--------------------------------------------------------------+--------+------------+---------+

ksamoray (approved):

PR #2235, opened 2026-08-20
Fix nsxt_edge_transport_node perpetual drift

ksamoray:
## Summary
- The NSX MP API returns bare realization UUIDs for `uplink_profile`, `vtep_ha_profile`, `host_switch_profile` and `transport_zone`, while configuration uses policy paths. This caused a perpetual diff after every apply.
- NSX also silently injects a default `audit_username` ("audit") that was not present in configuration, causing a perpetual diff there too.
- Enforce policy paths via `ValidateFunc` on `uplink_profile`, `vtep_ha_profile`, `host_switch_profile` elements and `transport_zone`, making the prior state value for these fields a reliable policy path.
- In `resourceNsxtEdgeTransportNodeRead`, capture the prior state before `setHostSwitchSpecInSchema` overwrites it with API UUIDs, then restore any policy-path values via `etNodeNormalizeHostSwitchesInState`.
- Mark `audit_username` as `Optional`+`Computed` so Terraform accepts the NSX default value without showing a diff when the user omits the field.

This is a hand-port of the equivalent fix on master, adapted for this release branch (master's version also reverts an unported earlier fix attempt; this PR implements the final, corrected behavior directly against branch_3122's current schema).

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)

No acceptance/unit test coverage is added, matching the corresponding fix on master, which also shipped without one.

### DIFF
--- a/nsxt/resource_nsxt_edge_transport_node.go
+++ b/nsxt/resource_nsxt_edge_transport_node.go
@@ -183,7 +183,8 @@ func getEdgeNodeDeploymentConfigSchema() *schema.Schema {
 							"audit_username": {
 								Type:        schema.TypeString,
 								Optional:    true,
-								Description: "CLI \"audit\" username",
+								Computed:    true,
+								Description: "CLI \"audit\" username; if omitted, set from NSX after creation",
 							},
 							"cli_password": {
 								Type:        schema.TypeString,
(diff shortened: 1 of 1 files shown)


yuanyouyao (approved):
LGTM 
The changes are identical to what has been running stably on master since May 2026, cleanly address the perpetual drift on nsxt_edge_transport_node, and compile cleanly with zero conflicts against branch_3122.

PR #2234, opened 2026-08-20
Bump dependencies to fix CVEs, pin NSX API SDK version

ksamoray:
## Summary
- Takes master's `go.mod`/`go.sum` wholesale to pick up the CVE fixes already applied there (Go toolchain, grpc, golang.org/x/crypto, golang.org/x/net, cloudflare/circl, terraform-plugin-sdk/testing, go-vmware-nsxt, etc.)
- Keeps branch_3122's current pin for `github.com/vmware/vsphere-automation-sdk-go/services/nsxt`: master's version is newer because it carries the NSX 9.2-only Transit Gateway BGP-routing SDK reorg, which this NSX 9.1.x-and-below release branch must not pull in
- Ran `go mod tidy` afterward to reconcile against this branch's actual code: `golang.org/x/exp` stays as a direct dependency (still imported directly by 4 files here), `go.uber.org/mock` is not added (this branch has no mock-based unit tests to require it)
- Also retags `build.md`'s `~/.terraformrc` example from an `hcl` to a `text` fence (pre-existing docs-lint failure, same one-line fix already on master in #2206)

Supersedes #2233, which was accidentally closed by a branch-rename operation that doesn't retarget cross-fork PRs.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)
- [x] `golangci-lint run ./nsxt/` (0 issues)
- [x] `terrafmt diff ./docs --pattern '*.md'` (clean)
- [x] `markdownlint-cli2 docs/**/*.md` (0 issues)

PR #2236, opened 2026-08-20
Fix route map perpetual drift from mutually exclusive matches

ksamoray:
## Summary
- Route map entries silently dropped `prefix_list_matches` when `community_list_match` was also set, since NSX treats the two as mutually exclusive per entry. The dropped value was never sent to NSX, so every subsequent read/plan showed drift.
- Reject the invalid combination with a clear error at apply time instead of silently discarding data.

Scoped port of the equivalent master fix: this branch has no `nsxt_policy_transit_gateway_route_map` resource, so only the `nsxt_policy_gateway_route_map` (Tier0/Tier1) hunk is included.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)
- [x] New acceptance test `TestAccResourceNsxtPolicyGatewayRouteMap_mutuallyExclusiveMatches` compiles (`go test -c`)


salv-orlando (approved):
LGTM

PR #2237, opened 2026-08-20
Suppress bgp_neighbor password drift after import

ksamoray:
## Summary
- NSX never returns `password` on GET, so `terraform import` leaves it empty in state, causing a spurious diff/re-apply on every plan.
- Reuse the `suppressIfEmptyPriorState` DiffSuppressFunc already used by `nsxt_policy_virtual_network_appliance` for the same write-only field import issue.

Scoped port of the equivalent master fix: this branch has no `nsxt_policy_transit_gateway_bgp_neighbor` resource, so only the base `nsxt_policy_bgp_neighbor` hunk is included.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)


salv-orlando (approved):
LGTM

PR #2239, opened 2026-08-20
Fix GRE tunnel create check and Exists error handling

ksamoray:
## Summary
- `resourceNsxtPolicyTier0GatewayGRETunnelExists` returned `true, nil` for any non-not-found error, including real API failures.
- `Create` never checked for a colliding explicit `nsx_id` before issuing the Patch, silently overwriting an existing GRE tunnel instead of rejecting the create.
- Fix `Exists` to only report `true` when the `Get` succeeds, propagate real errors instead of misreporting them as existing, and have `Create` reject a colliding explicit `nsx_id` up front.

Hand-adapted port of the equivalent master fix: master's version of this file uses a different (SDK-wrapper-refactor) client-construction pattern not present on this branch, and the full source PR (#2207) bundled this fix with an unrelated 36-file "move test-only Exists helpers into `_test.go` files" refactor touching many resources absent from this branch (including several Transit-Gateway-routing-child resources). Only the isolated GRE tunnel logic fix is included here, rewritten against this branch's actual client-construction code.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)

Note: CI's `lint` job's `make tools` step was previously expected to fail on PRs against this branch pending #2234's Go-version bump — that has since merged, so this should be unaffected.

### DIFF
--- a/nsxt/resource_nsxt_policy_tier0_gateway_gre_tunnel.go
+++ b/nsxt/resource_nsxt_policy_tier0_gateway_gre_tunnel.go
@@ -233,19 +233,19 @@ func resourceNsxtPolicyTier0GatewayGRETunnelExists(id, localeServicePath string,
 	client := locale_services.NewTunnelsClient(connector)
 	_, err = client.Get(tier0id, localeSvcID, id)
 
+	if err == nil {
+		return true, nil
+	}
 	if isNotFoundError(err) {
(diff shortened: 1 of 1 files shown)


salv-orlando (approved):
LGTM

PR #2240, opened 2026-08-20
Only set policy edge VM host_id when non-empty

ksamoray:
## Summary
- `nsxt_policy_edge_transport_node` always sent `"host_id": ""` in the `vm_deployment_config` payload when `host_id` was left unconfigured, because `getVMDeploymentConfigFromSchema()` unconditionally set `HostId` to a pointer to the (possibly empty) string. NSX then runs its VLAN-segment accessibility validation against an empty host context and creation/update fails with error 16031, even though the same segments work fine when `host_id` is set explicitly or when deployed via the UI.
- Guard the assignment the same way `compute_folder_id` is already guarded, and the same way this was previously fixed for the non-policy `nsxt_edge_transport_node` resource.
- Adds a unit test pinning `getVMDeploymentConfigFromSchema()`'s behavior: `host_id` must stay unset in the vAPI payload when not configured, and must be set when provided. The existing acceptance test `TestAccResourceNsxtPolicyEdgeTransportNode_basic` already configures `vm_deployment_config` without `host_id`, so it already exercises this path end-to-end and needs no changes.

Note: master's version of this new unit test is gated behind a `//go:build unittest` tag, but this branch's `make test` target doesn't pass `-tags unittest` anywhere, which would make the test silently never run. Dropped the tag here so it runs under the normal `go test ./nsxt/` invocation.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)
- [x] `go test ./nsxt/ -run TestUnitNsxt_getVMDeploymentConfigFromSchema -v` — both new tests pass

### DIFF
--- a/nsxt/resource_nsxt_policy_edge_transport_node.go
+++ b/nsxt/resource_nsxt_policy_edge_transport_node.go
@@ -1169,7 +1169,6 @@ func getVMDeploymentConfigFromSchema(iVmDeploymentCfg interface{}) (*data.Struct
 		vmDeploymentCfg := model.PolicyVsphereDeploymentConfig{
 			ComputeId:              &computeId,
 			EdgeHostAffinityConfig: edgeHostAffinityConfig,
-			HostId:                 &hostID,
 			ReservationInfo:        reservationInfo,
 			StorageId:              &storageID,
 			VcId:                   &vcID,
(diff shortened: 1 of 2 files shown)


salv-orlando (approved):
LGTM

PR #2241, opened 2026-08-20
Fix drift and update failures for adopted edge transport nodes

ksamoray:
## Summary
- `node_id`-adoption `ConflictsWith` `advanced_configuration`, `credentials`, `form_factor` and `management_interface`, so config for an adopted node never sets them. `Read` still populated all of them unconditionally from the API, showing permanent spurious drift on every plan; `form_factor` additionally carries a static schema `Default` that would keep re-asserting itself even with `Read` fixed, so it needs a `DiffSuppressFunc` instead.
- More seriously, `Update` always rebuilt the request body from config via `policyEdgeTransportNodePatch`, same as a freshly created node. For an adopted node that sends an empty `management_interface`, which NSX rejects as missing required fields on any update — including one only triggered by the spurious drift above. Route `Update` through the same GET-then-merge `policyEdgeTransportNodePredeployedPatch` that `Create` already uses for adoption.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)

Note: master's source PR added mock-based regression tests, which this branch has no framework for (no `utgomock_*` infrastructure here); the production fix is otherwise a direct, unmodified port.

### DIFF
--- a/nsxt/resource_nsxt_policy_edge_transport_node.go
+++ b/nsxt/resource_nsxt_policy_edge_transport_node.go
@@ -469,6 +469,13 @@ func resourceNsxtPolicyEdgeTransportNode() *schema.Resource {
 				Default:      model.PolicyEdgeTransportNode_FORM_FACTOR_MEDIUM,
 				Optional:     true,
 				ValidateFunc: validation.StringInSlice(policyEdgeNodeFormFactorValues, false),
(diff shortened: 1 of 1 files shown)


salv-orlando (approved):
LGTM

PR #2242, opened 2026-08-22
Fix nil pointer dereference in policy edge RTEP

ksamoray:
## Summary
- Add nil pointer checks for `SwitchSpec` and `SwitchName` when reading or setting RTEP configuration on `PolicyEdgeTransportNode`.
- Without these checks, if an Edge transport node returns a nil `SwitchSpec` or a switch with a nil `SwitchName` from the NSX API, the provider crashes with a nil pointer dereference panic.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)

Note: master's source PR added mock-based regression tests, which this branch has no framework for; the production fix is otherwise a direct, unmodified port.


salv-orlando (approved):
LGTM

PR #2244, opened 2026-08-23
Fix three bugs in nsxt_edge_transport_node_rtep resource

ksamoray:
## Summary
- **Issue 1** - Import by transport node UUID was broken: Read, Update and Delete all used `d.Get("edge_id")` to look up the edge, which is empty during terraform import (only `d.Id()` is populated at that point). Switch all three to `d.Id()`; add `d.Set("edge_id", id)` in Read so the attribute is written back to state after an import.
- **Issue 2** - Update sent an empty string pointer for `NamedTeamingPolicy` whenever the attribute was not set, causing NSX to reject the request with "uplink teaming policy name/s [] not specified". Apply the same non-empty guard that Create already had.
- **Issue 3** - DHCP and `no_ipv4` IP assignment types are not supported by the NSX RTEP API (a DHCP request triggers a server-side NPE, code 99). Add `getIPAssignmentSchemaForRTEP()` which strips those unsupported fields from the shared schema, so an unsupported assignment type is rejected at terraform plan time instead of reaching the API. Update docs accordingly.

**Dependency note:** `getIPAssignmentSchemaForRTEP()` calls `getIPAssignmentSchema(false)` (Optional), flipping this branch's `ip_assignment` field from its current `Required` to `Optional`. This isn't a regression introduced here — a separate, already-queued fix (upstream PR #1966, "Fixing the resource diff issue") adds `Computed: true` to `getIPAssignmentSchema`, and `Required` + `Computed` is an invalid schema combination, so upstream had already flipped this same field to `Optional` before this fix landed. When #1966 is ported to this branch, its hunk touching this specific line will be a no-op (already handled here); #1966's `Computed: true` addition to `getIPAssignmentSchema` itself is unaffected and still needs to land separately.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)

Note: master's source PR added mock-based regression tests, which this branch has no framework for; the production fix is otherwise a direct, unmodified port aside from the noted schema-signature simplification.

### DIFF
--- a/docs/resources/edge_transport_node_rtep.md
+++ b/docs/resources/edge_transport_node_rtep.md
@@ -21,7 +21,7 @@ resource "nsxt_edge_transport_node_rtep" "test_rtep" {
 
   host_switch_name = "someSwitch"
   ip_assignment {
-    assigned_by_dhcp = true
+    static_ip_pool = "ip-pool-uuid"
   }
   named_teaming_policy = "tp123"
   rtep_vlan            = 500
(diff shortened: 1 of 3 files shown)


salv-orlando (approved):
LGTM

PR #2246, opened 2026-08-23
Fix stale revision on concurrent BGP config update

ksamoray:
## Summary
- `nsxt_policy_bgp_config`'s Update used the revision cached in Terraform state, which goes stale when `nsxt_policy_tier0_gateway`'s own inline `bgp_config` block updates the same underlying object in the same apply.
- Fetch the current revision immediately before each write and retry on conflict, matching the pattern already used by `gateway_redistribution_config` and `tier0_gateway_ha_vip_config`.

Hand-adapted port of the equivalent master fix: master's version of this file uses a unified `cliBgpClient(sessionContext, connector)` wrapper (from an unported SDK-wrapper refactor) that doesn't exist on this branch. This branch still branches explicitly on `isPolicyGlobalManager(m)` between `locale_services.NewBgpClient` and `gm_locale_services.NewBgpClient`, so the fetch-fresh-revision-and-retry fix is applied to both paths directly.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)

Note: master's source PR added mock-based regression tests, which this branch has no framework for; the production fix logic is otherwise equivalent.

### DIFF
--- a/nsxt/resource_nsxt_policy_bgp_config.go
+++ b/nsxt/resource_nsxt_policy_bgp_config.go
@@ -235,7 +235,6 @@ func resourceNsxtPolicyBgpConfigCreate(d *schema.ResourceData, m interface{}) er
 func resourceNsxtPolicyBgpConfigUpdate(d *schema.ResourceData, m interface{}) error {
 	connector := getPolicyConnector(m)
 
-	revision := int64(d.Get("revision").(int))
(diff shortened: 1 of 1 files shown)


salv-orlando (approved):
LGTM

PR #2247, opened 2026-08-23
Fix bgp_config site_path drift after import on Global Manager

ksamoray:
## Summary
- `gateway_path` and `site_path` describe the parent Locale Service, not the BGP config object itself, and Read never derived either of them — it only ever populated whatever the Importer set once at import time. The Importer itself never set `site_path` at all, so a fresh import left it empty while config still declared it. Since `site_path` is `ForceNew`, the next plan proposed destroying and recreating the resource instead of showing no changes.
- Fetch the parent Locale Service in Read and re-derive both paths from it on every read, mirroring the identical pattern already used by the sibling `nsxt_policy_gateway_redistribution_config` resource.
- Add import regression coverage using `ImportStateKind: ImportBlockWithID`, since this resource's `id` is a synthetic random UUID (regenerated on every import) that `ImportStateVerify` can't meaningfully compare.

Hand-adapted port of the equivalent master fix: master's version uses a unified `cliTier0LocaleServicesClient(sessionContext, connector)` wrapper (from an unported SDK-wrapper refactor); this branch instead reuses the existing `tier0s.NewLocaleServicesClient(getSessionContext(d, m), connector)` wrapper already used by the sibling `gateway_redistribution_config` resource, with the same LM/GM `site_path` derivation logic.

Touches the same file as #2246 (different functions — Read vs Update), should merge cleanly regardless of order.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)
- [x] `go test ./nsxt/ -run TestAccResourceNsxtPolicyBgpConfig_importBasic -c` — compiles

Note: master's source PR added mock-based regression tests, which this branch has no framework for; the acceptance test coverage is otherwise a direct, unmodified port.

### DIFF
--- a/nsxt/resource_nsxt_policy_bgp_config.go
+++ b/nsxt/resource_nsxt_policy_bgp_config.go
@@ -14,6 +14,7 @@ import (
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/tier_0s/locale_services"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 
+	tier0s "github.com/vmware/terraform-provider-nsxt/api/infra/tier_0s"
 	"github.com/vmware/terraform-provider-nsxt/nsxt/util"
 )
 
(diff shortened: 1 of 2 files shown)


salv-orlando (approved):
LGTM

PR #2248, opened 2026-08-23
Skip empty bgp_route_leaking and static_route_advertisement blocks

ksamoray:
## Summary
- When `bgp_route_leaking {}` or `static_route_advertisement {}` is specified without any nested fields, the schema parser produced a model object with zero-value pointers. The NSX API silently discards such objects, so the subsequent Read returns an empty list / nil struct. On re-apply Terraform sees config has one empty block while state has none, producing a perpetual `+ bgp_route_leaking {}` or `+ static_route_advertisement {}` update.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)


salv-orlando (approved):
LGTM

PR #2249, opened 2026-08-23
Preserve redistribution config on T0 gw update if unset

ksamoray:
## Summary
- Updating the T0 gateway with a `locale_service` block (e.g. to set `preferred_edge_paths`) would overwrite and remove any existing route redistribution config when the `locale_service` block itself doesn't specify one — such as when redistribution is managed separately via `nsxt_policy_gateway_redistribution_config`. Preserve the existing config in that case.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)


salv-orlando (approved):
LGTM

PR #2250, opened 2026-08-23
Fix zone_based_span empty-list drop/panic in getSpanFromSchema

ksamoray:
## Summary
- The SDKv2 diff engine emits no attribute entries for a nested list's contents when it stays at zero elements, so a present-but-empty `cluster_based_span`/`zone_based_span` block can reconstruct its single list element as a bare nil.
- Branch_3122's `getSpanFromSchema` already determined block presence via list length (unlike master's pre-fix version, which checked the reconstructed element map and could treat a present-but-empty block as absent) — but it still used an unchecked type assertion on that single element, which would panic instead of producing the (possibly empty) `ZoneBasedSpan`/`ClusterBasedSpan`.
- Tolerate a nil element by defaulting to an empty zone list / empty span path instead of panicking.

Adds a plain unit test (no mock framework needed) reproducing the exact nil-element shape and asserting `getSpanFromSchema` doesn't panic and still round-trips through `setSpanFromSchema` correctly.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)
- [x] `go test ./nsxt/ -run TestGetSpanFromSchemaZoneBasedEmptyZones -v` — passes

### DIFF
--- a/nsxt/resource_nsxt_policy_transit_gateway.go
+++ b/nsxt/resource_nsxt_policy_transit_gateway.go
@@ -200,9 +200,13 @@ func getSpanFromSchema(iSpan interface{}) (*data.StructValue, error) {
 
 	// We're limiting to one span of any kind in the schema
 	span := iSpan.([]interface{})[0].(map[string]interface{})
-	if len(span["cluster_based_span"].([]interface{})) > 0 {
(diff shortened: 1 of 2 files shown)


salv-orlando (approved):
LGTM

PR #2251, opened 2026-08-23
Mark span.zone_based_span as Computed

ksamoray:
## Summary
- `span.cluster_based_span` and `span.zone_based_span` are mutually exclusive nested blocks (governed by `ExactlyOneOf`), but only `cluster_based_span` was declared `Computed`. This asymmetry caused Terraform to fail to reconcile the server-returned `zone_based_span` value with state, showing a perpetual "add zone_based_span" diff on every re-apply.
- Add `Computed: true` to `zone_based_span` to match its sibling. Update docs to reflect `(Optional, Computed)`.

Scoped to just this fix: master's source PR bundled unrelated route_controller_bgp_neighbor test IP-address tweaks and a `testAccNsxtExtraCoverage` helper change, neither of which relate to this bug.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)


salv-orlando (approved):
LGTM

PR #2252, opened 2026-08-23
Fix panic in getSpanFromSchema on nil span list element

ksamoray:
## Summary
- The Terraform SDK populates optional blocks that contain only empty/default attributes with a single nil element rather than an empty slice, or can hand this function a bare nil. The previous code performed unchecked type assertions on that element, triggering a panic at apply time.
- Add nil/type guards on `iSpan` itself and its single list element, in addition to the existing length-based presence checks for `cluster_based_span` and `zone_based_span`.

This hardens the *outer* unwrapping level of `getSpanFromSchema`; #2250 (already open) separately hardens the *inner* per-block element extraction (`cluster_based_span`/`zone_based_span`'s own single element). Different lines in the same function — should merge cleanly regardless of order.

Adds a unit test covering the nil-span and nil-element shapes that previously panicked (master's source PR had no test).

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)
- [x] `go test ./nsxt/ -run TestGetSpanFromSchemaNilSpan -v` — passes

### DIFF
--- a/nsxt/resource_nsxt_policy_transit_gateway.go
+++ b/nsxt/resource_nsxt_policy_transit_gateway.go
@@ -194,13 +194,20 @@ func resourceNsxtPolicyTransitGatewayExists(sessionContext utl.SessionContext, i
 }
 
 func getSpanFromSchema(iSpan interface{}) (*data.StructValue, error) {
-	if len(iSpan.([]interface{})) == 0 {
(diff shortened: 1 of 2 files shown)


salv-orlando (approved):
LGTM

PR #2253, opened 2026-08-23
Suppress ipsec_vpn_session psk drift after import

ksamoray:
## Summary
- NSX never returns the pre-shared key on GET, so terraform import leaves `psk` empty in state, causing a spurious diff/re-apply on every plan.
- Reuse the `suppressIfEmptyPriorState` DiffSuppressFunc already used for other write-only fields. `getIPSecVpnSessionCommon` is shared by `nsxt_policy_ipsec_vpn_session` and `nsxt_policy_transit_gateway_ipsec_vpn_session`, so this fixes both.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)


salv-orlando (approved):
LGTM

PR #2254, opened 2026-08-23
Use DiffSuppressFunc for transport_tunnels locale-service path drift

ksamoray:
## Summary
- NSX expands the flat IPsec VPN session path supplied in `transport_tunnels` (`/infra/tier-0s/<gw>/ipsec-vpn-services/<svc>/sessions/<id>`) to the locale-service-scoped form on storage (`/infra/tier-0s/<gw>/locale-services/default/ipsec-vpn-services/<svc>/sessions/<id>`).
- The user's original config value is preserved verbatim in state; a `DiffSuppressFunc` normalises both sides at plan time and returns true when they refer to the same session.
- `Read` is simplified to a direct `d.Set` that always fires, also fixing a stale-state bug when the API returns an empty list.

Ports the two pure-logic unit tests from the source PR (`normalizeL2VpnTransportTunnelPath`, `suppressL2VpnTransportTunnelsDiff`) into a plain test file; the mock-based `Read` regression tests are dropped since this branch has no mock-client test framework.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)
- [x] `go test ./nsxt/ -run 'TestUnitNsxt_normalizeL2VpnTransportTunnelPath|TestUnitNsxt_suppressL2VpnTransportTunnelsDiff' -v` — all 15 sub-tests pass

### DIFF
--- a/nsxt/resource_nsxt_policy_l2_vpn_session.go
+++ b/nsxt/resource_nsxt_policy_l2_vpn_session.go
@@ -54,9 +54,10 @@ func resourceNsxtPolicyL2VPNSession() *schema.Resource {
 				Default:     true,
 			},
 			"transport_tunnels": {
-				Type:        schema.TypeList,
-				Description: "List of transport tunnels(vpn sessions path) for redundancy",
-				Required:    true,
+				Type:             schema.TypeList,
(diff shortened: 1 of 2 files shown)


salv-orlando (approved):
LGTM

PR #2255, opened 2026-08-23
Accept policy path on import for IPSec VPN DPD/IKE/tunnel profiles

ksamoray:
## Summary
- `nsxt_policy_ipsec_vpn_dpd_profile`, `_ike_profile`, and `_tunnel_profile` used `schema.ImportStatePassthrough`, so importing with a full policy path (rather than a bare NSX ID) set the resource ID to the literal path string, which does not match what Read expects and breaks the subsequent lookup.
- Switch to `getPolicyPathOrIDResourceImporter`, which accepts either a bare ID or a full policy path and extracts the ID correctly from either form.

This is a scoped-down port of master's `1f34297b` ("Fix project context loss on import for IPSec VPN service, session, local endpoint, ike profile resources and add dedicated L2 VPN session importer"). That commit's core intent — fixing project/multitenancy context loss on import via `extractProjectIDFromPolicyPath`/`getContextSchema` for `ipsec_vpn_service`/`_session`/`_local_endpoint`/`l2_vpn_session` — depends on multitenancy support for IPSec VPN resources (added upstream by a separate, unported PR, #1978) that doesn't exist on this branch at all; those resources have no `context` schema field and no project-scoping logic whatsoever here, so that part of the fix is not applicable.

The DPD/IKE/tunnel profile importer switch is independently valid and doesn't depend on multitenancy, so it's included here on its own. Deliberately uses a plain (non-multitenancy) path example in the importer's error messages and docs, since advertising multitenancy-path support would be misleading — this branch's Read logic has no project-context awareness for these resources.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)

### DIFF
--- a/docs/resources/policy_ipsec_vpn_dpd_profile.md
+++ b/docs/resources/policy_ipsec_vpn_dpd_profile.md
@@ -55,3 +55,9 @@ terraform import nsxt_policy_ipsec_vpn_dpd_profile.test UUID
 ```
 
 The above command imports IPSec VPN DPD Profile named `test` with the NSX ID `UUID`.
+
+For example:
+
+```shell
+terraform import nsxt_policy_ipsec_vpn_dpd_profile.test /infra/ipsec-vpn-dpd-profiles/my-dpd-profile
+```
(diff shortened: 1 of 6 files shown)


salv-orlando (approved):
LGTM

PR #2257, opened 2026-08-24
Add documentation for missing data sources

yuanyouyao:
Add missing documentation files for four registered data sources:
- manager_cluster_node
- policy_gateway_prefix_list
- policy_gateway_route_map
- provider_info

These data sources are registered in nsxt/provider.go but lacked corresponding documentation under docs/data-sources/. Adding these files ensures complete provider documentation coverage and satisfies documentation linting requirements.

### DIFF
--- a/docs/data-sources/manager_cluster_node.md
+++ b/docs/data-sources/manager_cluster_node.md
@@ -1,0 +1,29 @@
+---
+subcategory: "Fabric"
+page_title: "NSXT: manager_cluster_node"
+description: A NSX manager cluster node data source.
+---
+
+# nsxt_manager_cluster_node
+
+This data source provides information about a specific NSX manager cluster node.
+
+## Example Usage
+
+```hcl
+data "nsxt_manager_cluster_node" "node1" {
+  display_name = "nsx-manager-node1"
+}
+```
+
+## Argument Reference
+
+* `id` - (Optional) Unique ID of the manager cluster node to retrieve.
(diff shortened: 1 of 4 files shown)


docs/data-sources/manager_cluster_node.md
```diff
@@ -0,0 +1,29 @@
+---
+subcategory: "Manager"
```
ksamoray: This could also  be in Fabric, where we have manager_cluster.

ksamoray (approved):
@yuanyouyao IDK how these are covered by testing - if we don't have coverage maybe place these as Beta. Otherwise LGTM.

docs/data-sources/manager_cluster_node.md
yuanyouyao: We already have coverage for all 4 data sources.

Fixed the subcategory

docs/data-sources/provider_info.md
```diff
@@ -0,0 +1,20 @@
+---
```
salv-orlando, line 1: This was deliberately not documented:

commit f4b3d131838aa64b6f4b317d5e770fcdae32df07
Author: Anna Khmelnitsky <akhmelnitsky@vmware.com>
Date:   Thu Nov 19 20:48:23 2020 -0800

    Introduce provider_info data source
    
    This data source exposes commit and build date for the provider, in
    order to ease testing effort. This data source is for internal use
    and thus not documented.

I am fine with documenting it, but we don't really have a category for it. Manager isn't the right one imho.

PR #2259, opened 2026-08-24
Fail when a user specifies a transport_zone_path for a project-based …

ksamoray:
…segment (#1962)

For project-based segments, NSX will always use the default transport zone. So when a transport_zone is explicitly specified for a segment which belongs to a project, NSX ignores it.
The provider should fail here, so the user will know that the config hasn't been applied instead of having a permadiff here.


salv-orlando (approved):
LGTM

PR #2082, opened 2026-05-14
Add Computed to advanced_config.connectivity to prevent drift on re-a…

yuanyouyao:
Add Computed to advanced_config.connectivity to prevent drift on re-apply

nsxt_policy_segment advanced_config.connectivity was declared Optional without
Computed. NSX defaults the field to ON when omitted, so the Read function
wrote ON back to state, causing a perpetual plan diff on every subsequent apply.

Two changes in segment_common.go:
- Add Computed: true to the connectivity schema attribute so Terraform treats
  a server-assigned value as authoritative when the user omits the field.
- Guard the Read assignment with a nil check and dereference the pointer
  consistently with all other pointer fields in the same block.

PR #2260, opened 2026-08-24
Add Computed to advanced_config.connectivity to prevent drift on re-a…

ksamoray:
…pply (#2082)

nsxt_policy_segment advanced_config.connectivity was declared Optional without Computed. NSX defaults the field to ON when omitted, so the Read function wrote ON back to state, causing a perpetual plan diff on every subsequent apply.

Two changes in segment_common.go:
- Add Computed: true to the connectivity schema attribute so Terraform treats a server-assigned value as authoritative when the user omits the field.
- Guard the Read assignment with a nil check and dereference the pointer consistently with all other pointer fields in the same block.


salv-orlando (approved):
LGTM

PR #2264, opened 2026-08-24
Fix GM support gap in security policy data source and its test

ksamoray:
nsxt_policy_predefined_security_policy relies on the nsxt_policy_security_policy data source to look up a policy by category/is_default when no path is known. That lookup was gated by a stale "Local Manager Only" restriction, so it silently failed on Global Manager despite GM being a documented target. The underlying SecurityPoliciesClient already routes to the correct GM/LM client via SessionContext, so drop the restriction and let it run universally; verified directly against a live GM.

Also harden the resource's import acceptance test: rather than assuming a default category policy already exists (NSX does not consistently auto-realize these under GM domains across versions -- a fresh NSX 4.2.x GM had none at all), the test now provisions its own target policy via nsxt_policy_parent_security_policy, matching what the original bug report's own FVT test did. Verified against live GM environments on NSX 9.0.2 and 4.2.3, and against LM.

### DIFF
--- a/nsxt/data_source_nsxt_policy_security_policy.go
+++ b/nsxt/data_source_nsxt_policy_security_policy.go
@@ -45,7 +45,6 @@ func dataSourceNsxtPolicySecurityPolicy() *schema.Resource {
 	}
 }
 
-// Local Manager Only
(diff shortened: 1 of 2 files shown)


salv-orlando (approved):
LGTM

PR #2266, opened 2026-08-24
Fix dhcp_server_addresses drift and create/update validation errors

ksamoray:
## Summary
- NSX does not persist `advanced_config.dhcp_server_addresses` while DHCP is deactivated, so the server-returned (empty) value would otherwise permanently conflict with whatever is left in config. Suppress the diff on this attribute whenever DHCP is deactivated.
- Also filter `dhcp_server_addresses` out of the payload sent to NSX (on both create and update) when DHCP is deactivated, and drop any empty-string placeholder `metadata.SchemaToStruct` can populate for elements suppressed by the diff. Without this, NSX rejects the payload with a `dhcp_server_addresses` ip-cidr-block format validation error.

Combines master's #2122 and #2172 into a single fix: this branch's `nsxt_vpc_subnet` only has a single (IPv4) DHCP family, unlike master's dual-stack IPv4/IPv6 schema, so the two upstream PRs (state drift for the deactivated family, then extending the create-path fix that only landed in Update) collapse into one change here.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)

### DIFF
--- a/nsxt/resource_nsxt_vpc_subnet.go
+++ b/nsxt/resource_nsxt_vpc_subnet.go
@@ -164,7 +164,8 @@ var vpcSubnetSchema = map[string]*metadata.ExtendedSchema{
 									SchemaType: "string",
 								},
 							},
-							Optional: true,
+							Optional:         true,
+							DiffSuppressFunc: suppressDhcpServerAddresses,
 						},
 						Metadata: metadata.Metadata{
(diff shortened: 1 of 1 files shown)


salv-orlando (approved):
LGTM

PR #2050, opened 2026-04-28
Ignore list order for nsxt_policy_project tier0_gateway_paths attribute

ksamoray:
NSX may return the list elements in a different order than posted. While this order is redundant, and we could use a TypeSet here, changing this attribute now will break backward compatibility. Therefore we use CustomizeDiff instead.

Fixes: #1535

PR #2269, opened 2026-08-24
Ignore list order for nsxt_policy_project tier0_gateway_paths attribu…

ksamoray:
…te (#2050)

NSX may return the list elements in a different order than posted. While this order is redundant, and we could use a TypeSet here, changing this attribute now will break backward compatibility. Therefore we use CustomizeDiff instead.

(cherry picked from commit e7170a6df6cb6bd73c7b4f1d94a42da046222676)


salv-orlando (approved):
LGTM

PR #2271, opened 2026-08-25
cache optimisation 

poojav25:
How the cache is optimized

- **The core problem this branch solves**: for VPC-scoped resources (VpcAttachment, VpcSubnet, Group, GatewayPolicy, StaticRoutes, DHCP bindings, etc.), the cache used to run one NSX Search per VPC. A Terraform config touching 100 VPCs meant 100 separate searches — each populating its own tiny cache bucket — which made "cache mode" slower than no caching at all.

- **The fix (getQueryString + projectScopedSearchContext in cache.go)**: for any VPC/Multitenancy-scoped resource, the cache bucket key drops the VPC ID and keeps only the project ID. One search now covers every VPC in a project, populating one shared bucket that all 100 VPCs' resources read from. This is the actual performance win — search count drops from O(VPCs) to O(1) per project per resource type.

- **The correctness cost of that optimization, and how it's paid down**: widening the bucket to project scope means two different VPCs' resources can now land in the same map. NSX short ids (nsx_id) are only guaranteed unique within a VPC, not across a project — so two VPCs could each have a resource named subnet1, and naively keying the shared bucket by short id would let one VPC's data silently mask the other's. The fix: 14 resource types that can be VPC-scoped are now indexed and looked up by path instead of short id (path is project/system-unique), via shouldIndexByPath + a new CacheKeyForResourceID helper that every CRUD call site and data source was migrated to use.

**Everything else in the diff is either infrastructure for that path-keying migration, or bugs found while doing it**:

-cache_test.go — new tests proving the query-scoping and path-indexing behavior

- The 10+ resource_nsxt_*.go one-liner diffs — switching each type's cache calls from raw id to CacheKeyForResourceID

- resource_nsxt_policy_group.go, resource_nsxt_policy_gateway_policy.go, resource_nsxt_policy_static_route.go, resource_nsxt_policy_dhcp_v4_static_binding.go, resource_nsxt_policy_gateway_policy_rule.go — fixes for call sites that were missed in that migration (shared Create/Read/Update/Delete functions hardcoding the wrong bucket, or a rule resource referencing its parent by the wrong key)

- Mutex/logging changes in cache.go — unrelated hardening (panic-safety, moving log calls outside the lock) picked up along the way

Net effect: cache mode should now do roughly 1 search per resource type per project, instead of 1 per VPC, without the path-keying migration reintroducing a correctness regression for the sake of that speed.

### DIFF
--- a/nsxt/cache.go
+++ b/nsxt/cache.go
@@ -19,6 +19,12 @@ import (
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 )
 
+// resourceTypeCache holds every populated search bucket for one NSX resource type. The
+// outer data map is keyed by the query string (see getCacheQueryKey) that produced a given
+// bucket — the same resource type can have more than one bucket if it's queried under
+// different tag filters or provider configs. The inner map is that bucket's contents,
+// keyed by whatever CacheKeyForResourceID/data-source id was used to look objects up
+// (see converListToMapByType for how an object ends up under multiple keys).
 type resourceTypeCache struct {
 	mu       sync.RWMutex
 	data     map[string]map[string]*data.StructValue // key: query, key: resource ID
@@ -26,6 +32,10 @@ type resourceTypeCache struct {
 	cacheMis atomic.Int64
 }
 
+// typeScopedCache is the process-wide root: one resourceTypeCache per NSX resource type
+// (see gcache). Its own mutex only protects the byTyp map itself (creating/looking up a
+// type's cache); it is never held while a search runs or a bucket is read, so looking up
+// one resource type never blocks concurrent access to another.
 type typeScopedCache struct {
 	mu    sync.RWMutex
 	byTyp map[string]*resourceTypeCache
@@ -35,11 +45,19 @@ type typeScopedCache struct {
 // queries (resource_type:<value>). Values must match NSX Search's resource_type
 // field exactly, including casing, and must not be changed here.
 const (
-	resourceTypeConnectivityPolicy                             = "ConnectivityPolicy"
-	resourceTypeDhcpV4StaticBindingConfig                      = "DhcpV4StaticBindingConfig"
-	resourceTypeDhcpV6StaticBindingConfig                      = "DhcpV6StaticBindingConfig"
-	resourceTypeGatewayPolicy                                  = "gatewaypolicy"
-	resourceTypeGroup                                          = "Group"
+	resourceTypeConnectivityPolicy        = "ConnectivityPolicy"
+	resourceTypeDhcpV4StaticBindingConfig = "DhcpV4StaticBindingConfig"
+	resourceTypeDhcpV6StaticBindingConfig = "DhcpV6StaticBindingConfig"
+	// resourceTypeGatewayPolicy is shared by the Tier-0/Tier-1 gateway policy resource and the
+	// VPC gateway policy resource: both cache under this one bucket. Same sharing applies to
+	// resourceTypeStaticRoutes and resourceTypeDhcpV4StaticBindingConfig below. Keep these
+	// path-indexed (see shouldIndexByPath) since the VPC-scoped side of each pair can collide
+	// on short id once its cache bucket is project-wide-shared.
+	resourceTypeGatewayPolicy = "gatewaypolicy"
+	resourceTypeGroup         = "Group"
+	// resourceTypeVPCGroup is a distinct NSX Search resource_type ("group", lowercase) from
+	// resourceTypeGroup ("Group") above — VPC-scoped groups and Tier-0/Tier-1-scoped groups are
+	// never in the same bucket, unlike the shared types noted above.
 	resourceTypeVPCGroup                                       = "group"
 	resourceTypeIpAddressAllocation                            = "IpAddressAllocation"
 	resourceTypeIpAddressPool                                  = "IpAddressPool"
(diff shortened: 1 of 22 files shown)


nsxt/cache.go
```diff
@@ -203,9 +245,19 @@ func CacheKeyForResourceID(resourceType string, d *schema.ResourceData) string {
 	return d.Id()
```
ksamoray, line 245: `CacheKeyForResourceID` falls back to the short id (`d.Id()`) whenever `path` isn't yet set on the `ResourceData`. That's exactly the state during `Create` and the immediate post-create `Read` for every one of the newly path-indexed VPC-scoped types added in `shouldIndexByPath` below.

Concretely: if two `nsxt_vpc_subnet` resources in different VPCs of the same project share the same `nsx_id` and are created in the same `terraform apply`, both `Create` calls compute the same `postWriteByKey` key (`"VpcSubnet::subnet1"`) since `path` isn't populated yet. Whichever `Read` runs first consumes that entry via `LoadAndDelete`; the other falls through to `readCache`, which repopulates the now-project-wide-shared bucket via `converListToMapByType`'s first-write-wins keying — landing resource A's data under a key resource B's `Read` will hit. That silently populates B's state from A's actual NSX object.

Should this fall back to something more path-like even before the resource has a `path` set (e.g. deriving it from the parent context), or does `MarkPostWriteAndInvalidateCacheForResourceType`/`CacheAwareResourceRead` need to be called only after `path` is set?

nsxt/cache.go
```diff
@@ -728,6 +817,14 @@ func mergeSecurityPolicyCacheSearchResults(parents, children []*data.StructValue
...
 
+// CacheAwareResourceRead is the standard read path for a policy resource's Read function: try
+// the shared search-backed cache first, and fall back to backendRead (a direct GET) on a cache
+// miss, a conversion failure, cache being disabled, or the resource having just been written by
+// this run (postWriteBypass — avoids reading a search snapshot that predates the write). resourceID
+// must be the same key used by the corresponding MarkPostWriteAndInvalidateCacheForResourceType
+// call for this resource (use CacheKeyForResourceID to guarantee that for path-indexed types).
+// Returns (object, cacheUsed, cacheAttempted, error): callers generally only need the object and
+// error; cacheUsed/cacheAttempted exist for tests and diagnostics.
 func CacheAwareResourceRead[T any](d *schema.ResourceData, m interface{}, connector client.Connector, resourceID string, resourceType string, bindingType bindings.BindingType, backendRead func() (*T, error), patchFunc func(obj *T) error) (*T, bool, bool, error) {
```
ksamoray, line 831: The short-id/path-indexed cache-bypass guard added to `cacheAwareDataSourceReadByID` (the `shouldIndexByPath(resourceType) && !strings.Contains(objID, "/")` check a few hunks up) isn't mirrored here in `CacheAwareResourceRead`. The doc comment added above `shouldIndexByPath` even calls out the data-source gap explicitly but doesn't mention this path.

That leaves resource `Read` exposed to the same cross-VPC short-id collision after `terraform import`: importers like `nsxtPolicyPathResourceImporterHelper`/`getVpcPathResourceImporter` call `d.SetId(shortID)` without setting `path`, so the post-import `Read`'s `CacheKeyForResourceID` falls back to the short id and this function proceeds through the normal (non-bypassed) cache path against the project-wide-shared bucket. If another VPC in the same project has a same-typed resource with the same short id, the import could silently pull in the wrong VPC's object data.

Was this intentionally scoped out (e.g. because import is expected to always warm a fresh `postWriteByKey` entry), or should the same bypass guard apply here too?

nsxt/cache.go
```diff
@@ -297,45 +357,56 @@ func getQueryString(resourceType string, context utl.SessionContext) string {
...
+// the search/query scope is widened; the resource lookup within the populated bucket still
+// keys by resourceID, which remains unique within the project.
+func projectScopedSearchContext(context utl.SessionContext) utl.SessionContext {
+	if context.ClientType == utl.VPC || context.ClientType == utl.Multitenancy {
+		context.VPCID = ""
+	}
+	return context
+}
+
 func (c *resourceTypeCache) getQueryResult(query string, resourceID string) (*data.StructValue, error) {
```
ksamoray, line 383: Removing the O(n) fallback scan for a raw `"id"` field match here means lookups now rely entirely on `converListToMapByType` having indexed every object under exactly the key this function will later be asked for. The comment above (`converListToMapByType indexes every object under its raw "id" field ... at populate time`) documents that invariant, but it's a fairly implicit contract across the ~35 resource types sharing this path.

If a resource's typed `model.Id` ever diverges from its raw `StructValue` `"id"` field for some type (SDK normalization quirk, etc.), the object would become silently unreachable by whichever key it wasn't indexed under — previously the O(n) scan would still have found it. Is there test coverage for a typed/raw id mismatch, or is this considered impossible by construction for all cached types?

nsxt/cache.go
poojav25: Good catch — confirmed this is real. Traced it through: two resources of the same path-indexed type in different VPCs of the same project, sharing a colliding short id, created in the same apply both compute the same postWriteByKey key (short id, since path isn't set on d yet at that point). Whichever Read runs first consumes that one-shot marker and does a correctly-scoped live GET. The other Read finds the marker already gone and falls through to gcache.readCache — still keyed by the short id — against the shared bucket, which resolves to whichever VPC's object converListToMapByType indexed first. That's a silent cross-VPC data leak, not a safe miss.
Fixed by adding the same short-id bypass guard cacheAwareDataSourceReadByID already has to CacheAwareResourceRead (and TryCacheRead, which had the identical gap): when the type is path-indexed and resourceID doesn't look like a path, skip the cache attempt entirely and fall through to backendRead, which is always correctly scoped by the resource's own session context. This fixes both directions you asked about — no need to defer the invalidation call or change when it's set, the read side now just refuses to trust a short-id lookup on a shared bucket.

nsxt/cache.go
poojav25: This was the same root cause as the postWriteByKey issue above — not intentionally scoped out, just missed. You're right that importers are a second, cleaner way to hit it: d.SetId(shortID) with no path set means the very first Read after import goes through the normal (non-bypassed) cache path against the shared bucket.
Fixed in the same change as above: CacheAwareResourceRead and TryCacheRead both now carry the identical shouldIndexByPath(resourceType) && !strings.Contains(resourceID, "/") guard that cacheAwareDataSourceReadByID has, gating the entire cache-attempt block (both postWriteByKey and gcache.readCache). Added tests for both functions that reproduce the vulnerable path pre-fix (a nil-connector panic proves control was reaching real search code) and confirm the bypass fires post-fix.

nsxt/cache.go
poojav25: Checked this against the actual vendored SDK binding source rather than assuming — PolicyConfigResourceBindingType() in vsphere-automation-sdk-go/services/nsxt/model/ModelPackageTypes.go defines fields["id"] = ... / fieldNameMap["id"] = "Id". The typed resource.Id field and the raw StructValue.Fields()["id"] field are literally the same wire field, mapped 1:1 — they're not independently derived or normalized differently, so they can't diverge for this binding. That would require a bug in the SDK's own field-mapping table, not something reachable from provider code.
Given that, I don't think there's a meaningful test to add here — forcing them apart would mean manually fabricating a StructValue via SetField that the real SDK could never produce on its own, which would test an impossible state rather than a real risk. Happy to revisit if you know of a specific resource type among the ~35 covered here where the id binding isn't 1:1, but I didn't find one.

ksamoray (approved):

PR #2273, opened 2026-08-25
Guard against nil entries in inter-VRF routing block iterators (#2081)

ksamoray:
Null block elements emitted by Terraform (e.g. empty dynamic blocks) caused a panic in getPolicyInterVRFRoutingFromSchema during Update because the unchecked type assertion brl.(map[string]interface{}) crashed when the slice contained a nil interface value. Add nil-continue guards before every such assertion in the bgp_route_leaking, static_route_advertisement, and advertisement_rule loops.

(cherry picked from commit f3b860b28c30df2686fa6e2088a9034f86ee6712)


salv-orlando (approved):
LGTM

PR #2078, opened 2026-05-11
Fix index out of range panic for short segment paths

yuanyouyao:
isT1Segment() and getT1IdFromSegPath() accessed pathSplit[len-4] while only guarding with `len >= 3`. When segment_path contains exactly three tokens (e.g. "a/b/c"), len-4 evaluates to -1 and Go panics with "index out of range [-1]" instead of returning a graceful error.

Raise the minimum length guard to >= 4 in both functions so the index is always in bounds. A valid T1 segment path has the form /infra/tier-1s/{id}/segments/{id} (six tokens), so the existing happy-path behaviour is unchanged.

PR #2274, opened 2026-08-25
Fix index out of range panic for short segment paths (#2078)

ksamoray:
isT1Segment() and getT1IdFromSegPath() accessed pathSplit[len-4] while only guarding with `len >= 3`. When segment_path contains exactly three tokens (e.g. "a/b/c"), len-4 evaluates to -1 and Go panics with "index out of range [-1]" instead of returning a graceful error.

Raise the minimum length guard to >= 4 in both functions so the index is always in bounds. A valid T1 segment path has the form /infra/tier-1s/{id}/segments/{id} (six tokens), so the existing happy-path behaviour is unchanged.

(cherry picked from commit 10eb67e940a91d4b222d95b5927ed2d14e1c9280)


salv-orlando (approved):
LGTM

PR #2277, opened 2026-08-25
Change instance_count from TypeList to TypeSet in policy constraint r…

ksamoray:
…esource (#2018)

Address a bug with constraint resources where returned list order from NSX confuses TF. Change type to Set as order in this case is not important.

(cherry picked from commit c77ab6797657793e5a02dd5c79f36fe8f576f3c4)


salv-orlando (approved):
LGTM

PR #2278, opened 2026-08-25
Fix acknowledgment of non-ackable upgrade precheck warnings (#2126)

ksamoray:
Skip acknowledgment of upgrade precheck warnings when needs_ack is false.

The Upgrade Coordinator returns CannotAcknowledgeException (code 30976) when trying to acknowledge prechecks that do not require acknowledgment. Precheck warnings like edgeNodeAlarmCheckCheck (which are informational warnings with needs_ack set to false) can be collected and requested for acknowledgment, but there is no need to ack them.

Update acknowledgePrecheckWarnings to filter out warnings where needs_ack is false or has already been acknowledged, preventing unnecessary and invalid API calls. Additionally, handle and gracefully ignore any 30976 error code returned by the NSX API in case the precheck's state changed.

(cherry picked from commit 32ec818c07f847fc0c1e19fdbd1a15a2619149a4)

### DIFF
--- a/nsxt/resource_nsxt_upgrade_precheck_acknowledge.go
+++ b/nsxt/resource_nsxt_upgrade_precheck_acknowledge.go
@@ -119,7 +119,11 @@ func acknowledgePrecheckWarnings(m interface{}, precheckIDs []string) error {
 	}
 	var warns []string
 	for _, warn := range precheckWarnings {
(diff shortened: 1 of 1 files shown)


salv-orlando (approved):
LGTM

PR #2279, opened 2026-08-25
fix(upgrade): validate upgrade_run_id in postcheck datasource and fix…

ksamoray:
… docs (#2066)

- nsxt_upgrade_run now generates a verifiable ID (UUID + FNV hash tag "nsxt_upgrade_run") so it can be validated by consumers
- nsxt_upgrade_postcheck datasource validates upgrade_run_id against the verifiable ID scheme, mirroring how nsxt_upgrade_prepare_ready validates upgrade_prepare_ready_id
- datasource now calls d.SetId() so Terraform tracks state correctly
- docs: fix wrong subcategory/page_title/description, document timeout/ interval/delay args, clarify that failed_group is empty on success

(cherry picked from commit a1d65ee21f88deafe32d5f3cc06f3849014bcf6d)

### DIFF
--- a/docs/data-sources/upgrade_postcheck.md
+++ b/docs/data-sources/upgrade_postcheck.md
@@ -1,12 +1,20 @@
 ---
-subcategory: "Beta"
-page_title: "NSXT: edge_upgrade_group"
-description: An Edge Upgrade Group data source.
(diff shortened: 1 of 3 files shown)


salv-orlando (approved):
LGTM

PR #2280, opened 2026-08-25
Probe LDAP provider only when enabled (#2012)

ksamoray:
The LDAP identity source resource is probing the LDAP provider for availability. Yet, when the servers are disabled, there is no point in doing so as there's nothing to be probed.

(cherry picked from commit e1a41a0a6a46180c89ae57bca50107eb9c1f35b6)


salv-orlando (approved):
LGTM

PR #2286, opened 2026-08-25
Fix title in policy gateway interface doc

ksamoray:
Correct the H1 heading in the policy gateway interface realization data source documentation to match the actual Terraform resource name ('nsxt_policy_gateway_interface_realization') instead of using an unnecessary '_info' suffix.

The filename and top-level heading previously contained '_info', which caused inconsistency with the data source name in HCL.


salv-orlando (approved):
LGTM

PR #2292, opened 2026-08-25
Document realized_id is LM-only for nsxt_policy_transport_zone DS

ksamoray:
NSX only populates realization_id for Local Manager Policy resources (it correlates a migrated MP unique_id); Global Manager intent objects have no MP-migration lineage, so the field is always empty there. This is expected NSX API behavior, not a provider defect, so clarify the doc instead of changing code.

AI-Tool-Used: Claude Code
AI-Tool-Use-Level: medium
AI-Code-Category: non-production


salv-orlando (approved):
LGTM

PR #2293, opened 2026-08-25
Remove redundant attributes from TZ datasource doc

ksamoray:
Remove 'is_default' and 'transport_type' from the exported attributes section of the 'nsxt_policy_transport_zone' data source documentation.

These fields are already fully documented in the 'Argument Reference' section as optional inputs. Since they function as both input arguments and computed attributes, listing them in both sections is redundant and unnecessary.


salv-orlando (approved):
LGTM

PR #2296, opened 2026-08-25
Update nsxt_edge_transport_node docs for bug 3689817 fix

ksamoray:
transport_zone, host_switch_profile, uplink_profile and vtep_ha_profile now require policy paths (validatePolicyPath enforced); bare UUIDs are rejected. Update examples to use .path, fix attribute descriptions, and mark audit_username as Computed.


salv-orlando (approved):
LGTM

PR #2069, opened 2026-05-07
Rename ha_flow_mirroring_enabled in LB UDP docs

yuanyouyao:
Update the policy_lb_fast_udp_application_profile resource documentation to reflect the correct attribute name. The attribute was previously documented as `ha_flow_mirroring_enabled`, but the actual schema argument is `flow_mirroring_enabled`. Both the HCL example and the argument reference are updated accordingly.

PR #2298, opened 2026-08-25
Rename ha_flow_mirroring_enabled in LB UDP docs (#2069)

ksamoray:
Update the policy_lb_fast_udp_application_profile resource documentation to reflect the correct attribute name. The attribute was previously documented as `ha_flow_mirroring_enabled`, but the actual schema argument is `flow_mirroring_enabled`. Both the HCL example and the argument reference are updated accordingly.


salv-orlando (approved):
LGTM

PR #2302, opened 2026-08-26
Mark identity_source_id as Computed on policy role binding

ksamoray:
## Summary
- NSX can populate this field even when it's left unset in config, causing spurious drift. Mark it `Computed` so Terraform accepts the server-assigned value.

Partial port of master's #1959 ("Revise LDAP tests to pass with OpenLDAP server"): only this production schema fix is applicable here — the rest of that PR revises LDAP acceptance tests to target OpenLDAP infrastructure specifics unrelated to this branch.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)
- [x] `golangci-lint run ./nsxt/` — 0 new issues


salv-orlando (approved):
LGTM

PR #2303, opened 2026-08-27
Fix orphaned NSX allocation on IP realization failure

ksamoray:
## Summary
- `resourceNsxtPolicyIPAddressAllocationCreate` waits for realization when `allocation_ip` is unset, and on failure tried to clean up the already-created NSX object via `resourceNsxtPolicyIPAddressAllocationDelete(d, m)`. That function reads `d.Id()`, but the ID is only set on success — so on failure the cleanup silently no-op'd on an empty ID, leaking the IP allocation on NSX Manager with no corresponding Terraform state.
- Fix: temporarily set `d.SetId(id)` so the real delete path (including its cache invalidation) can be reused for cleanup, then clear the ID again before returning the original error so `Create` still correctly reports nothing was created.

## Test plan
- [x] `go build ./...` and `go vet ./...`
- [x] New unit tests in `utgomock_resource_nsxt_policy_ip_address_allocation_test.go` cover both failure paths (realization query error, and realized response missing the `allocation_ip` attribute), asserting cleanup actually deletes the NSX object and `d.Id()` ends up empty
- [x] `go test -tags unittest ./nsxt/...` passes for this resource

### DIFF
--- a/nsxt/resource_nsxt_policy_ip_address_allocation_test.go
+++ b/nsxt/resource_nsxt_policy_ip_address_allocation_test.go
@@ -6,6 +6,7 @@ package nsxt
 
 import (
 	"fmt"
+	"regexp"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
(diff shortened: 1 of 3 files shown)


salv-orlando (approved):
LGTM

nsxt/resource_nsxt_policy_ip_address_allocation_test.go
```diff
@@ -347,3 +479,69 @@ data "nsxt_policy_realization_info" "subnet_realization" {
...
+}`, poolName, subnetName, consumerName)
+
+	if !includeExhausting {
+		return config
+	}
+
+	return config + fmt.Sprintf(`
+
+resource "nsxt_policy_ip_address_allocation" "exhausted" {
+  nsx_id       = "%s"
```
salv-orlando: For a more thorough coverage we might want to remove nsx_id and let the id be auto-generated

ksamoray: Done in 15a37deb — dropped the fixed nsx_id so it's auto-generated. Since the ID isn't known ahead of time anymore, the not-leaked check now lists everything under the pool and confirms only the consumer's ID is present, rather than looking up a specific known ID.

PR #2310, opened 2026-09-02
Use latest NSX SDK

ksamoray:
Bump vsphere-automation-sdk-go/services/nsxt to the latest version.
The new SDK adds an include_conflicts query parameter to most
List/Get/Create endpoints, a force parameter to Share resource
Delete/Patch/Update, and a handful of model changes, none of which
are source-compatible with the previous SDK.

Update all generated API wrappers (api/**), direct SDK call sites in
nsxt/**, and generated mocks (mocks/**) to match the new client
signatures, always passing nil/omitting the new parameters to
preserve existing behavior. A number of wrappers assert c.Client to
an anonymous structural interface rather than a named SDK type; the
compiler does not catch a stale signature there, so those call sites
required the same signature updates plus targeted unit test runs to
surface and fix the runtime-only breakage.

DnsAutoRecordConfig.ptr_record_zone_path became
ptr_record_zone_paths (a list) in the new SDK. Since
nsxt_policy_dns_record_auto_config has not shipped in a release yet,
adopt the improved shape directly: ptr_record_zone_paths is now a
list of DNS zone paths, so a single auto record config can route PTR
records for IP blocks that span multiple CIDRs to their respective
reverse zones. Docs updated accordingly.

RouteControllerBgpNeighborConfig.route_filtering now uses a
dedicated RouteControllerBgpRouteFiltering type instead of the
shared BgpRouteFiltering type; this is an already-released resource,
so this is purely an internal type-rename with no schema impact.


yuanyouyao (approved):
LGTM

PR #2311, opened 2026-09-02
Add unit tests for near-zero-coverage data sources

ksamoray:
Most data source Read functions had no dedicated unit test, relying only on TestAcc* acceptance tests that require a live NSX Manager. Add gomock and search-mock-based unit tests (following the existing utgomock_* convention) for 87 data sources, raising data source coverage from 41.6% to 79.3% and overall package coverage from 64.3% to 68.9%.
    
Also regenerate the ClusterControlPlanesClient mock, which was stale against the SDK version already pinned on this branch.


nsxt/resource_nsxt_policy_dns_record_auto_config.go
```diff
@@ -52,11 +52,11 @@ func resourceNsxtPolicyDnsRecordAutoConfig() *schema.Resource {
 				ValidateFunc: validatePolicyPath(),
 				Description:  "Policy path to a DnsZone (local or shared) in which auto-created A records are placed.",
 			},
-			"ptr_record_zone_path": {
-				Type:         schema.TypeString,
-				Optional:     true,
-				ValidateFunc: validatePolicyPath(),
-				Description:  "Optional policy path to a DnsZone for auto-created PTR (reverse DNS) records. When absent, no PTR record is auto-created.",
+			"ptr_record_zone_paths": {
```
salv-orlando, line 55: Did you serendipitously find a bug, or is this due to the SDK update?

ksamoray: The SDK update revealed that the API was revised (we don't fail yet as we're stuck on the last build before the IDPS issue).

salv-orlando (approved):
LGTM

PR #2312, opened 2026-09-02
Fix nsxt_transport_node data source get-by-id EdgeNode check

ksamoray:
dataSourceNsxtTransportNodeRead converted obj.NodeDeploymentInfo before obj was ever assigned from the freshly-fetched objGet, so the EdgeNode type check always operated on a zero-value node and the "get by id" path could never succeed.


yuanyouyao (approved):
LGTM (Approved). The fix is accurate, surgical, and resolves a 100% reproducible failure in the id lookup path for the nsxt_transport_node data source.

PR #2313, opened 2026-09-02
Expand unit test coverage for low-coverage resources and data sources

ksamoray:
Raises overall unit test coverage on ./nsxt/... from 68.8% to 73.8%,
building on the prior "near-zero-coverage data sources" pass:

- 8 resources whose existing mock tests only exercised error/guard
  paths now cover Create/Read/Update/Delete success paths too:
  cluster_security_config (20% -> ~97%), ldap_identity_source
  (20% -> ~97%), l7_access_profile (24% -> ~89%), idps_cluster_config
  (25% -> ~97%), segment_port_profile_bindings (33% -> ~86%),
  host_transport_node (36% -> ~88%), ip_pool_block_subnet
  (37% -> ~90%), ipsec_vpn_session (41% -> ~74%). Adds minimal
  package-level client-constructor seams (mirroring the existing
  cli*Client convention used throughout the provider) to
  cluster_security_config and idps_cluster_config so their SDK
  clients can be mocked; reuses existing seams elsewhere.

- 9 near-identical LB monitor/application profile resources gain
  success-path tests via their already-shared mock helpers,
  raising each from ~46-50% to 84-100%.

- 13 data sources that previously had no mock test at all now have
  one, nearly all reaching 100%: 8 trivial generic-search wrappers,
  plus provider_info, host_upgrade_group, edge_cluster,
  gateway_interface, and proxy_config (the last reusing an existing
  seam from its sibling resource file instead of constructing its
  SDK client inline).


yuanyouyao (approved):
Approve / LGTM. This PR significantly improves codebase health, testability, and CI reliability without introducing any runtime regressions or breaking changes.

PR #2316, opened 2026-09-06
Expand unit test coverage for the cache layer and realization polling

ksamoray:
cache_test.go's tests predated the unittest tag convention and were named
outside the TestMock*Nsxt*/TestUnitNsxt_* pattern that `make test-unit` (and
CI) filters on via -run, so they never actually executed as part of the
tracked unit test suite. Tag the file `//go:build unittest` and rename its
tests to TestUnitNsxt_* so they're finally counted.

Add nsxt/cache_unit_test.go covering the cache.go helpers that were still at
0%: query-string/query-key construction, the per-type/per-query cache
buckets (hit, miss-and-populate, and post-write-bypass paths) including the
composite gateway-policy/security-policy + rule merge, tag stripping and
provider-managed-tag patching, and TryCacheRead/CacheAwareResourceRead/
cacheAwareDataSourceReadByID end to end using the existing cliQueryClient
stub pattern from policy_search_unit_test.go.

Also add coverage for nsxtPolicyWaitForRealizationStateConf (policy_utils.go),
reusing the existing realization-info mock helper to exercise its Refresh
closure (found, not-yet-realized, and error cases).

Package coverage under `go test ./nsxt -tags=unittest -run='^(TestMock.*Nsxt.*|TestUnitNsxt_.*)' -coverpkg=./nsxt`
moves from 74.0% to 74.8%, with cache.go itself going from mostly-0% to
75-100% coverage across nearly every function.


yuanyouyao (approved):
Approve / LGTM. This PR completes the unit test coverage suite for the caching layer with high precision and zero regressions.

PR #2317, opened 2026-09-07
Document reserved tracking tag in caching guide

yuanyouyao:
Add an explicit warning and troubleshooting entry to the in-memory caching guide regarding the reserved 'nsx-tf/tf-run-id' tag scope.

In config_scope mode, the provider automatically injects and filters this tracking tag on NSX objects for Search API scoping, while intentionally stripping it from Terraform state upon read to prevent internal metadata from polluting user state files.

Manually specifying 'nsx-tf/tf-run-id' in HCL tag blocks creates an unresolved mismatch against the stripped state, causing persistent plan drift on subsequent runs. Explicitly document this limitation.


ksamoray (approved):
LGTM

PR #2319, opened 2026-09-07
Fix two terraform-cache-brownfield regressions in the resource cache

ksamoray:
terraform-cache-brownfield started failing after passing consistently,
across two unrelated bugs surfaced by config_scope cache mode:

1. IpAddressPoolStaticSubnet/IpAddressPoolBlockSubnet permanently
   bypassed the cache on every read. CacheAwareResourceRead has a
   shortIDBypass safety net: for path-indexed resource types, if the
   resourceID passed in doesn't contain "/", the cache is skipped
   entirely in favor of a direct GET, since a short id may not be
   unique across the project-wide search scope these types share.
   Both resources' Read functions unconditionally used the short
   d.Id() as that key instead of preferring d.Get("path") (populated
   since the prior Create) the way every other path-indexed resource
   does, so shortIDBypass fired on every single read, permanently.

   Fix: prefer d.Get("path"), falling back to d.Id() only when path
   isn't set yet (Create-then-Read within the same apply, or import) —
   matching the existing pattern already used by e.g. Service,
   SegmentPort, and PolicyNatRule.

2. GatewayPolicy, SecurityPolicy, and VpcGroup silently wiped their
   provider-managed cache tracking tag (scope="nsx-tf/tf-run-id") on
   every Update, including a bare description change with no config
   drift on tags. Their Create/Update seeds Terraform state's "tag"
   attribute with the merged (user + provider-managed) tag list
   immediately before building the outgoing NSX PATCH payload from
   that same state — the only way the shared build-and-patch helpers
   pick up the provider-managed tag for the payload. That seed used
   initPolicyTagsSet, which filters the provider-managed tag out — a
   filter added so Read never leaks the tag into state — silently
   defeating the seed before the payload was ever built. Every Update
   sent an empty tag list, and the Read that immediately follows
   caught the tag missing and re-patched it — a spurious extra write
   on every steady-state apply.

   Fix: add initPolicyTagsSetForOutgoingPatch, the same conversion
   without the filter, for this one seed-before-build-and-patch use
   across all 8 call sites (gateway policy, security policy, VPC
   gateway policy, VPC group; Create and Update). The tag actually
   persisted to Terraform state is unaffected — each CRUD function's
   own trailing Read still overwrites "tag" via setPolicyTagsInSchema,
   which filters correctly for that purpose.

Both fixes include a mock-based regression test that fails against
the pre-fix code and passes with the fix.

### DIFF
--- a/nsxt/policy_utils.go
+++ b/nsxt/policy_utils.go
@@ -143,6 +143,28 @@ func initPolicyTagsSet(tags []model.Tag) []map[string]interface{} {
 	return tagList
 }
 
+// initPolicyTagsSetForOutgoingPatch is initPolicyTagsSet without the provider-managed-tag
+// filter. GatewayPolicy/SecurityPolicy/VpcGroup's Create and Update seed d's "tag" attribute
+// with getPolicyTagsWithProviderManagedDefaults(d, m) immediately before building the NSX
+// PATCH payload from schema (via getValidatedTagsFromSchema, which just reads d.Get("tag")) —
+// the only way those resources' shared build-and-patch helpers pick up the provider-managed
+// tag for the outgoing payload. Using the filtered initPolicyTagsSet here silently strips the
+// tag right back out before that read ever happens, so every Update (e.g. a plain description
+// change) wipes the tag from the object, only for the next Read's cache-aware tag-check to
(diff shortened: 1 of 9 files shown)


yuanyouyao (approved):
LGTM

PR #2320, opened 2026-09-07
Expand unit test coverage to 81% across idps, gateway, and LB resources

ksamoray:
The idps data sources (cluster_config, custom_signature, settings,
signature_diff, signature_version, system_signatures) and two idps resources
called SDK client constructors directly instead of through the package's
injectable cliXClient var pattern, which is what makes gomock-based unit
testing possible elsewhere in this codebase. Wire them through injectable
vars (reusing two that already existed) and add mocked test coverage for
the previously-untestable read/list paths.

Also fix 8 test functions (5 pre-existing, 3 introduced in this PR) named
TestMock<Foo> without "Nsxt" in the name, so they never matched
`-run='^(TestMock.*Nsxt.*|TestUnitNsxt_.*)'` and silently never ran in
`make test-unit` or CI despite passing when invoked directly.

Add CRUD, error-path, and pure schema-conversion test coverage across the
package's largest remaining gaps: lb_virtual_server, edge_transport_node
(policy and manager variants), tier0_gateway(_interface), tier1_gateway(_interface),
domain, project, service, transit_gateway, lb_pool, predefined_gateway_policy,
predefined_security_policy, nat_rule, ipsec_vpn_service/session,
ip_pool_static_subnet, role_binding, virtual_network_appliance,
dhcp_v4_static_binding, bgp_neighbor, ospf_area, gateway_policy_rule,
l2_vpn_session, static_route, edge_cluster, and the shared
segment_common/segment_port_common helpers.

Note on the coverage number: this project's `go.mod` pins `go 1.26.6`, which
is what CI's `actions/setup-go` (via `go-version-file: go.mod`) actually
builds and tests with. A newer local Go toolchain (e.g. 1.27.x) counts
statements more granularly and will report a noticeably higher, non-comparable
percentage for the identical code — always measure against go.mod's pinned
version when validating this number locally.

### DIFF
--- a/nsxt/data_source_nsxt_policy_idps_cluster_config.go
+++ b/nsxt/data_source_nsxt_policy_idps_cluster_config.go
@@ -9,7 +9,6 @@ import (
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/settings/firewall/security/intrusion_services"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 )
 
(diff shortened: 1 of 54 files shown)


yuanyouyao (approved):
Approve / LGTM. PR #2320 significantly strengthens test coverage and confidence across the provider's most critical resources while strictly adhering to the repository's mocking conventions.

PR #2328, opened 2026-09-10
Fix orphaned NSX allocation on IP realization failure (branch_3122)

ksamoray:
Cherry-pick of #2303 onto `branch_3122`.

## Summary
- `resourceNsxtPolicyIPAddressAllocationCreate` waits for realization when `allocation_ip` is unset, and on failure tried to clean up the already-created NSX object via `resourceNsxtPolicyIPAddressAllocationDelete(d, m)`. That function reads `d.Id()`, but the ID is only set on success — so on failure the cleanup silently no-op'd on an empty ID, leaking the IP allocation on NSX Manager with no corresponding Terraform state.
- Fix: temporarily set `d.SetId(id)` so the real delete path (including its cache invalidation) can be reused for cleanup, then clear the ID again before returning the original error so `Create` still correctly reports nothing was created.
- The `_poolExhausted` acceptance test uses an auto-generated `nsx_id` on the second allocation, matching the common case and the reported customer scenario.

Note: this branch predates the gomock-based unit test infrastructure (`go.uber.org/mock`, `mocks/infra/...`), so the unit test coverage added in #2303 isn't included here — only the production fix and the acceptance test.

## Test plan
- [x] `go build ./...`
- [x] Acceptance test passes against a live NSX appliance, confirming no orphaned allocation is left behind on create failure

### DIFF
--- a/nsxt/resource_nsxt_policy_ip_address_allocation.go
+++ b/nsxt/resource_nsxt_policy_ip_address_allocation.go
@@ -138,9 +138,7 @@ func resourceNsxtPolicyIPAddressAllocationCreate(d *schema.ResourceData, m inter
 		stateConf := nsxtPolicyWaitForRealizationStateConf(connector, d, *obj.Path, timeout)
 		entity, err := stateConf.WaitForState()
 		if err != nil {
-			// Clean up NSX allocation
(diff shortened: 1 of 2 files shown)


salv-orlando (approved):
LGTM

PR #2332, opened 2026-09-14
Mark quotas as computed in nsxt_vpc schema

yuanyouyao:
Add Computed: true to the quotas attribute definition in the nsxt_vpc resource schema.

When nsxt_vpc is created without specifying quotas in configuration, metadata.StructToSchema assigns an empty slice [] when reading the resource state from the NSX API (as Limits is nil). Because quotas was previously defined with Optional: true only, Terraform Core treated the attribute as unset (null) upon creation, resulting in state drift (+ quotas = []) during subsequent refresh or refresh-only plans.

Marking quotas as computed allows Terraform to reconcile the empty slice default without detecting unexpected drift.


yuanyouyao:
FVT test has passed,
+--------------+------------------------------+--------+------------+---------+
| Test Case ID |    Description/Test Name     | Result | Comment(s) |  Owner  |
+--------------+------------------------------+--------+------------+---------+
|      -       | test_cache_context_id_change | Passed |            | Unknown |
+--------------+------------------------------+--------+------------+---------+

salv-orlando (approved):
LGTM

PR #2333, opened 2026-09-14
Fix parallel acceptance test collisions on shared/singleton NSX objects

ksamoray:
## Summary

Two related CI flakes, both caused by acceptance tests running via `resource.ParallelTest` against a resource that NSX only allows one of per shared scope:

- **DNS service**: NSX allows only one DNS service per project. Every DNS zone/rule/record/record-auto-config/service acceptance test creates its own project-scoped `nsxt_policy_dns_service`, and running them concurrently raced and failed with "Only one DNS service per project is allowed" (code 641147).
- **VPC service endpoint**: NSX auto-manages a single `_services` subnet per VPC to host all `nsxt_vpc_service_endpoint` resources, creating it on first use and marking it for deletion once the last endpoint is removed. Tests that create a service endpoint ran against the same shared CI VPC, so one test's destroy could mark `_services` for deletion while another test's create raced it, failing with "is marked for deletion... wait for the purge cycle" (code 500045).

Both are fixed the same way: switch the affected tests from `resource.ParallelTest` to `resource.Test`, matching the existing convention already used in this repo for other NSX singleton constraints (LDAP identity source, proxy config).

## Test plan

- [x] `go build ./...` and `go vet ./nsxt/...` pass
- [ ] Live re-run of the affected tests against the shared CI VPC/project to confirm no more collisions (requires live NSX/CI environment)


yuanyouyao (approved):
please check the failing check. Others LGTM.

PR #2335, opened 2026-09-14
Add nsxt_policy_dns_record_auto_config data source

ksamoray:
Adds a search-API-based data source for looking up existing DNS record auto configs by id or display_name within a project. Includes acceptance test coverage combined with the existing resource test, and unit test coverage using the mocked search client.


nsxt/data_source_nsxt_policy_dns_record_auto_config.go
```diff
@@ -0,0 +1,32 @@
...
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MPL-2.0
+
+package nsxt
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceNsxtPolicyDnsRecordAutoConfig() *schema.Resource {
```
salv-orlando, line 11: If I am not mistaken no other resource references dns_record_auto_config, ie: the path we will fetch with the data source cannot be used in any other resource.

If this is correct then either: this data source is meaningless - besides maybe troubleshooting, or we also make available other attributes like `ip_block_path, a_record_zone_path, ptr_record_zone_path` which might be used in other resources.

ksamoray: I don't think that this will be useful, I'll close this.

PR #2339, opened 2026-09-14
Add nsxt_policy_project_centralized_network_attachment data source

ksamoray:
Adds a search-API-based data source for looking up existing project centralized network attachments by id or display_name within a project. Includes acceptance test coverage combined with the existing resource test, and unit test coverage using the mocked search client.


salv-orlando (commented):
Looks good, but then for consistency should we also add a data source for the TGW attachment?

ksamoray:
> Looks good, but then for consistency should we also add a data source for the TGW attachment?

Makes sense

salv-orlando (approved):
LGTM

PR #2341, opened 2026-09-14
Add nsxt_policy_transit_gateway_bgp_neighbor data source

ksamoray:
Adds a search-API-based data source for looking up existing transit gateway BGP neighbors by id or display_name, scoped to a required parent Transit Gateway path (the underlying BgpNeighborConfig type is shared with Tier-0/Tier-1 gateways, so scoping is required).

The resource had no acceptance test file, so this also adds one (TestAccResourceNsxtPolicyTransitGatewayBgpNeighbor_basic/importBasic) covering the resource itself plus the new data source in the same config. Also adds unit test coverage for the data source using the mocked search client.


ksamoray:
Closing, as this might not have a good use case

PR #2342, opened 2026-09-14
Add nsxt_policy_transit_gateway_community_list data source

ksamoray:
Adds a search-API-based data source for looking up existing transit gateway community lists by id or display_name, scoped to a required parent Transit Gateway path (the underlying CommunityList type is shared with Tier-0 gateway community lists, so scoping is required). Includes acceptance test coverage combined with the existing resource test, and unit test coverage using the mocked search client.


ksamoray:
No good use case for this datasource. Closing

PR #2344, opened 2026-09-14
Add nsxt_policy_transit_gateway_route_map data source

ksamoray:
Adds a search-API-based data source for looking up existing transit gateway route maps by id or display_name, scoped to a required parent Transit Gateway path. Includes acceptance test coverage combined with the existing resource test, and unit test coverage using the mocked search client.


salv-orlando (approved):
LGTM

PR #2345, opened 2026-09-14
Add nsxt_vpc_dhcp_v6_static_binding data source

ksamoray:
Adds a search-API-based data source for looking up existing DHCP IPv6 static bindings by id or display_name, scoped to a required parent VPC Subnet path. Includes acceptance test coverage combined with the existing resource test, and unit test coverage using the mocked search client.


ksamoray:
As we do not have a data source for v4 static bindings, I guess that this might not be very useful. Closing.

PR #2350, opened 2026-09-16
Fix flaky TestAccResourceNsxtPolicyTier0Gateway_withMultiVRF

ksamoray:
## Summary
- `TestAccResourceNsxtPolicyTier0Gateway_withMultiVRF` occasionally fails with `Error while creating objects of type:LocaleServices (code 500157)` / `object marked for deletion... wait for the purge cycle (code 500045)`.
- The parent Tier0 gateway in this test is created without an `nsx_id` (auto-assigned), but the shared teardown step (also used by `_withVRF`) applies a config with a hardcoded `nsx_id = "vrf-parent"`, forcing a delete+recreate of the parent using that literal ID.
- The sibling `TestAccResourceNsxtPolicyTier0Gateway_withVRF` test uses the same hardcoded `"vrf-parent"` ID for its own parent gateway throughout its entire lifecycle. When the two tests run close together, the recreate in `_withMultiVRF`'s teardown step can race `_withVRF`'s create/delete of the same NSX path, hitting the backend's 5-minute purge cycle for the reused path.
- Fix: drop the hardcoded `nsx_id` from both templates so the parent gateway keeps its NSX-assigned ID across all steps, removing the forced replace and the path collision. `nsx_id` is `Optional+Computed`, so omitting it causes no diff for already-created resources.

## Test plan
- [ ] Run `TestAccResourceNsxtPolicyTier0Gateway_withVRF` and `TestAccResourceNsxtPolicyTier0Gateway_withMultiVRF` acceptance tests, including repeated/parallel runs, to confirm the flake no longer reproduces.


yuanyouyao (commented):
Cursor analysis,

1. Findings & Architectural Observations
While the PR solves the immediate path collision, reviewing the broader interaction reveals architectural quirks and potential side effects:

Finding 1: Unintended In-Place Mutation in _withMultiVRF Step 3
Why was parent replaced in _withMultiVRF in the first place? Because _withMultiVRF reused testAccNsxtPolicyTier0WithVRFTearDown(), which was written specifically for _withVRF.

Now that nsx_id is removed and no longer forces a replace, Step 3 of _withMultiVRF performs an in-place update on the parent gateway:

multi_vrf_inter_sr toggled to false: In Steps 1 & 2, parent had multi_vrf_inter_sr = true. In testAccNsxtPolicyTier0WithVRFTearDown(), multi_vrf_inter_sr is omitted. The schema defines Default: false with description:
"Flag to control multi VRF inter SR. This is one time toggle flag and can't be disabled once enabled" Terraform plans multi_vrf_inter_sr: true -> false and sends MultiVrfInterSrRouting = false in resourceNsxtPolicyTier0GatewayUpdate. Depending on the NSX release and backend validation, trying to disable this flag on an existing Tier-0 gateway can be rejected by the NSX API.

Spurious creation of parent-loopback: testAccNsxtPolicyTier0WithVRFTearDown() defines nsxt_policy_tier0_gateway_interface.parent-loopback. In _withMultiVRF, there was never any loopback interface in Steps 1 & 2. During Step 3 (intended as VRF teardown), it unexpectedly creates a new loopback interface (4.4.4.12/24) on the parent gateway, only to destroy it at the end of the test.
Display Name Mutation: display_name mutates from "parent-<name>" back to the static "parent".

Finding 2: Static display_name = "parent"
Both testAccNsxtPolicyTier0WithVRFTemplate and testAccNsxtPolicyTier0WithVRFTearDown still hardcode display_name = "parent". While NSX Policy allows duplicate display names (only the resource ID/path must be unique), hardcoding static names in acceptance tests makes debugging logs and NSX UI confusing when multiple tests run against the same testbed. Using fmt.Sprintf("parent-%s", name) is cleaner.

Finding 3: Subnet Mask Inconsistency in _withVRF Loopback
In testAccNsxtPolicyTier0WithVRFTemplate:

resource "nsxt_policy_tier0_gateway_interface" "parent-loopback" {
  subnets = ["4.4.4.12/32"]
}
In testAccNsxtPolicyTier0WithVRFTearDown:

resource "nsxt_policy_tier0_gateway_interface" "parent-loopback" {
  subnets = ["4.4.4.12/24"]
}
In _withVRF, Step 3 mutates the loopback subnet mask from /32 to /24.

2. Suggested Better / Complete Fix
Rather than having _withMultiVRF share testAccNsxtPolicyTier0WithVRFTearDown(), give _withMultiVRF its own matching teardown template:

func testAccNsxtPolicyTier0WithMultiVRFTearDown(name string) string {
	return testAccNsxtPolicyGatewayInterfaceDeps("11, 12", false) + fmt.Sprintf(`
resource "nsxt_policy_tier0_gateway" "parent" {
  display_name       = "parent-%s"
  ha_mode            = "ACTIVE_ACTIVE"
  edge_cluster_path  = data.nsxt_policy_edge_cluster.EC.path
  multi_vrf_inter_sr = true
}`, name)
}
And in TestAccResourceNsxtPolicyTier0Gateway_withMultiVRF:

			{
				Config: testAccNsxtPolicyTier0WithVRFTearDown(),
				Config: testAccNsxtPolicyTier0WithMultiVRFTearDown(name),
			},

This ensures:
parent retains its exact configuration (multi_vrf_inter_sr = true, same display_name).
No bogus parent-loopback interface is created during teardown.
The VRF (nsxt_policy_tier0_gateway.test) and BGP config are cleanly deleted as intended.

yuanyouyao (approved):

PR #2353, opened 2026-09-17
policy_predefined_gateway/security_policy: Deprecation and fixes

salv-orlando:
This commit includes several changes to
- resource nsxt_policy_predefined_security_policy
- resource nsxt_policy_prefefined_gateway_policy

1) default_rule configuration must be possible only for default
   security policies. User-defined security policy have no rule
   for which is_default:true. The loop for updating default rules
   will skip updates, causing a permadiff. In addition, if
   applied to a non-default policy, these resources might clash
   with rules configured for it, removing any rule which might
   have been added in the same Terraform configuration. This
   applies to both predefined gateway and security policy resources.
2) The scope in the default_rule attribute for the predefined
   gateway policy resource is used to match the rule against
   a rule, already existing on the NSX backend, with the same
   scope. If the scope specified in the default_rule attribute
   does not exist in NSX, apply was succeeding without making
   any change in the backend. This resulted in a permadiff, and
   has been fixed by failing the apply operation.
3) The predefined gateway policies allows multiple default_rule
   blocks. It is important that the user specifies a default_rule
   block for each rule in the NSX gw policy with is_default:true,
   otherwise this will also cause a permadiff. This has been
   addressed by failing the apply operation if not all default
   rules in NSX have a corresponding default_rule attribute in
   the Terraform configuration.
4) The predefined security policy resource is actually not applicable
   to Global Manager, since there are no category-specific default
   rules, both in the default and in site-specific domains. This
   change ensures this resource can only be configured for LM.
5) The predefined security policy resource serves a use case that
   can also be addressed by importing a Default security policy using
   policy_security_policy or adding rules to it using
   policy_security_policy_rule. As mentioned in point (1) above, the
   fact that policy_predefined_security_policy also allows for
   configuring rules might generate a conflict with other resources.
   Since the use case can be implemented using different resource,
   the resource policy_predefine_security_policy is deprecated in
   this commit, and the risks of using it in combination with other
   policy resources are highlighted in the documentation.


salv-orlando:
Hello @martinrohrbach - this PR deprecates a resource.
I'd like to check with you whether, from a user perspective, this deprecation will cause any issue into your automation.

yuanyouyao (approved):
Please fix the docs-lint.

Others LGTM.

PR #2355, opened 2026-09-22
Raise ./nsxt unit test coverage from 84.5% to 86.9%

ksamoray:
Adds unit tests across 18 files, targeting the highest-ROI gaps found by analyzing per-function coverage output: pure schema<->struct conversion helpers that had no direct tests (QoS rate shapers, DNS zones, L7 access profile child entries, context profile attributes, BGP/route-controller BGP neighbor configs, Tier0 gateway interface OSPF config, transit gateway BGP config, IDS profile criteria and signatures), a full sweep of ~20 resource Importer functions whose ID/path-parsing logic was entirely untested, CRUD success-path tests for the LB persistence profile resources and OSPF config resource (previously only their error paths were covered), and tests for several shared low-level helpers (cache.go, policy_common.go, provider.go).

Does not reach the 90% target: the remaining gap is concentrated in resources that resist the same patterns - e.g. resource_nsxt_policy_ domain.go's Create/Update goes through a client var typed to an unexported concrete SDK type that can't be mocked without touching production code, and resource_nsxt_policy_transit_gateway.go's Update has several d.HasChange(...)-gated branches that would need new diff-based ResourceData test infrastructure with no precedent in this suite.


yuanyouyao (approved):
Fills longstanding gaps in converter helpers, resource importers, and persistence profiles that were historically difficult to test.

LGTM

